PR #8, opened 2020-05-15
Rebase "Compose apps" on 2020.5+fio

mike-sul:


src/helpers.cc
```diff
@@ -10,16 +12,71 @@
...
+      if ((*i).isObject() && (*i).isMember("uri")) {
+        LOG_INFO << "\t\t" << i.key().asString() << " -> " << (*i)["uri"].asString();
+      } else {
+        LOG_ERROR << "\t\tInvalid custom data for docker_compose_apps: " << i.key().asString();
+      }
+    }
+  }
+}
+
+static __attribute__((constructor)) void init_pacman() {
```
mike-sul, line 55: It's not necessary to do `static __attribute__((constructor))` in case of an executable,  just `PackageManagerFactory::registerPackageManager` can be called. The original idea was to be able to "inject" a new package manager implemented and delivered as a shared library.

doanac: This just gives me a fool-proof way to make sure it gets called at initialization. Otherwise the LiteClient constructor needs to call registerPackageManager, but only if `-DBUILD_DOCKERAPP` .

doanac:
@mike-sul - so you think this is ready to merge?

mike-sul:
I cannot verify it from end-to-end perspective as it requires creating a  'docker_compose_apps' target instead of 'docker_apps' on the backend. Also, we might wanna update lmp-device-register to configure sota.toml with the 'docker compose' variables instead of 'docker app'.

doanac:
If you want to give it a spin, you can enable it in your factory with this:

 https://source.foundries.io/factories/andy-corp/ci-scripts.git/commit/?id=211e276a4273e4282b2481507e5f2c0c6955a3cf

that will continue to publish the old docker apps and also this now compose app.

And once we get things tested in our factories, yes, we'll update the sota.toml generation to use this instead.

mike-sul:
> that will continue to publish the old docker apps and also this now compose app.

another missing piece for me is `docker-compose download` command.

doanac:
> another missing piece for me is docker-compose download command.

Good catch. This needs to be merged into meta-lmp:

 https://github.com/doanac/meta-lmp/commit/c70d45fcd247e72f2bab2081f8a2f693d48b78e4

mike-sul:
aktually this can be merged as it does not interfere with the docker app standalone package manager.

doanac:
you want to hold off while we re-work the download function to be in aklite, or should do you want to do that in a new PR?

mike-sul:
> you want to hold off while we re-work the download function to be in aklite, or should do you want to do that in a new PR?

This PR has the callback implementation which is agnostic to the package manager type, so if we merge it people can give it try. So, yes, I prefer to do the download function in a new PR.

mike-sul:
@doanac so merge or not to merge :)?

doanac:
I forgot this stuff depended on the callbacks stuff. However, there's only one commit in the "callback" patches that does callbacks. The rest is code cleanup.

mike-sul:
@doanac thanks!

PR #9, opened 2020-05-25
FF-160/download docker app

mike-sul:


mike-sul:
@doanac I enabled clang tidying again by introducing our own testing docker container based on ubuntu.bionic.
I am going to add some negative tests for the introduced functionality.

src/composeappmanager.cc
```diff
@@ -64,15 +112,49 @@ ComposeAppConfig::ComposeAppConfig(const PackageConfig &pconfig) {
...
+  if (raw.count("registry_auth_creds_endpoint") == 1) {
+    registry_conf.auth_creds_endpoint = raw.at("registry_auth_creds_endpoint");
+  } else {
+    const std::string treehub_endpoint = "treehub";
+    const std::string registry_creds_endpoint = "hub-creds/";
+    registry_conf.auth_creds_endpoint = pconfig.ostree_server;
+    auto endpoint_pos = registry_conf.auth_creds_endpoint.find(treehub_endpoint);
+    if (endpoint_pos != std::string::npos) {
+      registry_conf.auth_creds_endpoint.replace(endpoint_pos, registry_creds_endpoint.length(),
+                                                registry_creds_endpoint);
```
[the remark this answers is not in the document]
doanac: I am wondering if we should have this config defined. It pins us to hub.foundries.io. However - docker makes this discoverable. eg:
~~~
$ curl -v https://hub.foundries.io/v2/msul-dev01/shellhttpd/manifests/latest
...
< HTTP/1.1 401 Unauthorized
...
< Www-Authenticate: Bearer realm="https://hub.foundries.io/token-auth/",service="registry",scope="repository:msul-dev01/shellhttpd:pull"
~~~

src/composeappmanager.cc
```diff
@@ -145,25 +227,274 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target &target
...
+  MultiPartSHA256Hasher hasher;
+
+  DownloadCtx download_ctx{output_file, hasher, expected_size};
+  auto get_blob_resp = registry_repo_client->download(compose_app_blob_url, DownloadHandler, nullptr, &download_ctx, 0);
+
+  if (!get_blob_resp.isOk()) {
+    throw std::runtime_error("Failed to download App blob: " + get_blob_resp.getStatusStr());
+  }
+
+  output_file.close();
```
doanac, line 450: this is needed because the file normally won't  get closed until the destructor is called, right?

src/composeappmanager.cc
```diff
@@ -145,25 +227,274 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target &target
...
+  }
+
+  std::string auth_secret_str{username + ':' + secret};
+  auto encoded_auth_secret = Utils::toBase64(auth_secret_str);
+
+  LOG_DEBUG << "Got Docker Registry credentials, username: " << username;
+  return "authorization: basic " + encoded_auth_secret;
+}
+
+std::string RegistryClient::getBearerAuthHeader(const std::string& repo) const {
```
doanac, line 497: this might be too painful for now and we should just keep it like this with a TODO. However, docker provides a way to find credentials. In theory, on LmP, you look at `/usr/lib/docker/config.json` and then `$HOME/.docker/config.json` for `auths` and `credHelpers`. If "auths" you just use the value provided. If "credsHelpers", you call that program and use its stdout.

Maybe we assert hub.foundries.io for now though and because its hub.fio, we know how to handle this w/o using the credsHelper.

src/composeappmanager.cc
```diff
@@ -1,23 +1,67 @@
...
   // Utils::shell isn't interactive. The compose commands can take a few
   // seconds to run, so we use boost::process:system to stream it to stdout/sterr
-  bool cmd_streaming(const std::string &cmd) {
+  bool cmd_streaming(const std::string& cmd) {
     LOG_DEBUG << "Running: " << cmd;
     return boost::process::system(cmd, boost::process::start_dir = root_) == 0;
   }
 
-  bool fetch(const std::string &app_uri) {
+  bool download(const std::string& app_uri);
```
doanac, line 59: I'm not against breaking these out. However, it confused my eyes when I first saw this. Maybe you should create a change before this one that breaks out all the methods of ComposeApps to be like this. Then when you do this change it won't look so different from everything else.

dev-shell
```diff
@@ -9,7 +9,7 @@ cd $here
 
 if ! docker image inspect $container 2>/dev/null 1>/dev/null ; then
 	echo "Container not found, doing one time docker build"
-	docker build -t $container -f ./aktualizr/docker/Dockerfile.debian.testing .
+	docker build -t $container -f ./docker/Dockerfile.ubuntu.bionic .
```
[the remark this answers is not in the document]
mike-sul: let's see what the aktualizr team will do about it if anything, I would stay with whatever they will come up with because if we just switch to something higher we might end-up with some incompatibilities as you stated. I consider it as a temporal solution.

docker/Dockerfile.ubuntu.bionic
```diff
@@ -0,0 +1,82 @@
...
+  xsltproc \
+  zip
+
+WORKDIR /ostree
+RUN git init && git remote add origin https://github.com/ostreedev/ostree
+RUN git fetch origin v2018.9 && git checkout FETCH_HEAD
+RUN NOCONFIGURE=1 ./autogen.sh
+RUN ./configure CFLAGS='-Wno-error=missing-prototypes' --with-libarchive --disable-gtk-doc --disable-gtk-doc-html --disable-gtk-doc-pdf --disable-man --with-builtin-grub2-mkconfig --with-curl --without-soup --prefix=/usr
+RUN make VERBOSE=1 -j4
+RUN make install
```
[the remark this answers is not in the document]
mike-sul: It has to do with the hack around testing the ostree based package manager (LD_PRELOAD and those two overridden methods ), the hack doesn't work with any ostree version.

src/composeappmanager.cc
```diff
@@ -145,25 +227,274 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target &target
...
     }
   }
 }
+
+std::pair<std::string, std::string> ComposeApp::parseAppUri(const std::string& uri) {
+  auto split_pos = uri.find('@');
+  if (split_pos == std::string::npos) {
+    throw std::invalid_argument("Invalid App URI: '@' not found in " + uri);
+  }
+
```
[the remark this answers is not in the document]
mike-sul: Can't it be different than `hub.foundries.io`, e.g. test mocks, tests, on prem deployment. I suppose it can be verified/checked if it's inline with what is defined in the config (registry_base_url).

src/composeappmanager.cc
```diff
@@ -1,23 +1,64 @@
...
+  static constexpr const char* const ArchiveExt{".tgz"};
+
+  struct Manifest : Json::Value {
+    static constexpr const char* const Format{"application/vnd.oci.image.manifest.v1+json"};
+    static constexpr const char* const Version{"v1"};
+
+    Manifest(const Json::Value& value = Json::Value()) : Json::Value(value) {
+      auto manifest_version{(*this)["annotations"]["compose-app"].asString()};
+      if (manifest_version.empty()) {
+        throw std::runtime_error("Got invalid App manifest: " + value.asString());
```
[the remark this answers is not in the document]
mike-sul: no, there is try/catch around functionality that instantiates Manifest.

src/composeappmanager.cc
```diff
@@ -145,25 +227,274 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target &target
...
+  LOG_DEBUG << name_ << ": Factory: " << factory_name;
+
+  auto repo_name{uri.substr(factory_name_pos + 1, split_pos - factory_name_pos - 1)};
+  LOG_DEBUG << name_ << ": App Repo name: " << repo_name;
+
+  return {repo_name, compose_app_digest};
+}
+
+std::string ComposeApp::downloadArchive(const std::string& repo, const Manifest& manifest) {
+  std::string archive_file_name_prefix;
```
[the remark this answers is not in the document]
mike-sul: yeap

src/composeappmanager.cc
mike-sul: yes.

src/composeappmanager.cc
mike-sul: The credHelpers relies on aktaulizr-get. What's the point to go from aktualizr to the docker stuff which in turn comes back to aktualizr if 'we' are aktualizr and go straight to the auth service.
It's not `hub.foundries.io`, it's `https://ota-lite.foundries.io:8443/hub-creds/`.

src/composeappmanager.cc
mike-sul: ok.

src/composeappmanager.cc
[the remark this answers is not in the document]
doanac: this all fits in with my other comment above about just letting docker's api tell us what to authenticate with.

src/composeappmanager.cc
doanac: what i was trying to say here is that: we should really make this generic and not tied to hub.foundries.io. maybe eventually though since its a bit more work. in the case of hub.foundries.io, we could skip looking up the cred helper since we know how to do that.

what's nice about the docker-compose implementation is that it could download the archive from *any* docker registry with *any* credentials docker supports. 

but since this is tied to hub.foundries.io, I think we can skip the configuration data and just hard-code to hub.foundries.io and raise an error if its not a hub.foundries.io app. In the future we could add logic to support other docker registries.

mike-sul: I see what you mean. Still, we can change our registry's hostname overtime or deploy additional one with the same way of getting creds to it (via https://ota-lite.foundries.io:8443/hub-creds).

src/composeappmanager.cc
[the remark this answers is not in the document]
mike-sul: fixed

src/composeappmanager.cc
mike-sul: I just moved it (`download`) to the private methods

src/composeappmanager.cc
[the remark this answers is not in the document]
mike-sul: Maybe I just check if the App URI hostname matches the registry_base_url's hostname here, and check if the registry_base_url is equal to `https://hub.foundries.io` at the configuration time. 
Another option is to get rid off `base_url`, auth_token_endpoint, repo_base_url configuration variables at all and extracts registry hostname from App URI, although '/token-auth/' and /v2/ has to be hardcoded then. BTW, /token-auth/ is hardcoded in the docker compose download implementation,  too.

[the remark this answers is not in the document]
doanac: I think getting rid of them is fine for now. "v2" can be hard-coded. I guess I did "auth-token" incorrectly in my docker-compose patch. The "Www-Authenticate:" header should tell the client where to go. For docker, I noticed my logic would be broke as the do:
~~~
Www-Authenticate: Bearer realm="https://auth.docker.io/token",service="registry.docker.io"...
~~~

mike-sul:
@doanac I removed the registry's configuration params. I didn't implement the auth request discovery as it requires changing in the aktualizr's HTTP client or developing our own, not sure if it's justifiable at the current moment provided that we don't support other than our Registry.

src/composeappmanager.h
```diff
@@ -5,35 +5,123 @@
 
 #define PACKAGE_MANAGER_COMPOSEAPP "ostree+compose_apps"
 
+// TODO: restructure composeappmanager.* so it contains just the manager implemenation
+// ComposeApp and Docker::* stuff should be moved into separate source files
+
```
doanac, line 10: I agree. I think a docker.cc file for the implementation would be nice. This is only going to grow in complexity.

mike-sul:
Checked this branch on RPi and QEMU, it looks like it works as expected.

src/composeappmanager.cc
```diff
@@ -64,15 +115,40 @@ ComposeAppConfig::ComposeAppConfig(const PackageConfig &pconfig) {
     boost::algorithm::to_lower(val);
     docker_prune = val != "0" && val != "false";
   }
+
+  // Docker Registry Configuration (TODO: to be removed)
```
doanac, line 119: Can we just remove this stuff and hard-code it in `getBasicAuthHeader()` where we already have the big TODO's? I'm hoping this will never leak into someone's sota.toml. Or do you need someway to intercept things for unit tests?

mike-sul: What about the on-premises deployment? In this case, the base URL will be different than 'https://ota-lite.foundries.io:8443'.

doanac:
i think this looks pretty good now. Here's how i think we should proceed to merge:

1) lets do a new PR to foundriesio/aktualizr to pull in the Docker fix. ie to do a `[fiofromlist]` patch with https://github.com/advancedtelematic/aktualizr/commit/85776e83dfb744d15612f09b21e7ccdd9b5e67df

2) lets do a new PR to aktualizr-lite that:
  a) pulls in the new foundriesio/aktualizr
  b) fixes whitespace

3) update this PR or do a new PR with this stuff minus all the other noise.
i think breaking out the new docker stuff into docker.[h/cc] files would be great.

let me know if you need help with 1.

src/composeappmanager.cc
doanac: ugh - your right. l guess we should keep the logic that discovers it - lets drop the config value method though.

mike-sul: What is `the config value method`?
I guess you mean that we just deduce the auth base url from `ostree_server` url and append `hub-creds`  to it?

mike-sul:
> 3\. update this PR or do a new PR with this stuff minus all the other noise.

I opt for a new PR, this one has already too many changes.

src/composeappmanager.cc
doanac: correct. we deduce from ostree_server

mike-sul:
I suppose we need to make corresponding changes in lmp-device-register to configure aklite/sota.toml with compose_apps package manager.

doanac:
> I suppose we need to make corresponding changes in lmp-device-register to configure aklite/sota.toml with compose_apps package manager.

eventually. i think we are a few weeks away from that. we need to first get everyone's personal factories using this (and we can update the `fioctl devices config updates` command to move each device) first. once everyone is using it and happy we can make this the new default.

we'll also have to update the partner-setup repo to define things using docker-compose rather than .dockerapp

mike-sul:
> eventually. I think we are a few weeks away from that.

Why not create a branch and PR right now, so we can test the end-to-end flow as it would happen in real-life starting from bitbaking.

doanac:
that's fine.

src/composeappmanager.cc
mike-sul: Removing the ability to configure the auth URL and just deducting it from the ostree/treehub URL is fine.
But, removing this param auth_creds_endpoint and moving the deduction logic in `getBasicAuthHeader` is not fine. First of all, we need to pass at least one param to RegistryClient anyway (either the treehub URL or already deducted auth URL), secondly if we do deduction in `getBasicAuthHeader` then we will do it for each request while this URL is static and should be evaluated once at startup.

doanac: ack.

mike-sul: done

mike-sul:
I rebased this branch on https://github.com/foundriesio/aktualizr/pull/38 and removed the commit that introduces customer docker container for testing.

doanac:
looks good.

PR #11, opened 2020-05-29
Getting rid of the ostree mock

mike-sul:
Effectively the mock was required for getting a checksum of the currently installed ostree revision/commit since the ostree package manager fails to get it if a host is not booted into the non the ostree-based rootfs. But, we can get a checksum of the ostree repo deployment by simply listing the available deployments in the repo and getting a checksum of the first one (the staged deployment always goes first in that list, since we are not booted into in the ostree rootfs then the staged deployment is the one that would be used if we boot into the ostree roots, i.e. the hard-links of rootfs files/dirs points to the same files as hard links of the given deployment).

This is just a draft implementation aimed to depict an idea. There are a few options for a final implementation:

1. Put the required implementation/logic  into OstreeManager::getCurrentHash. Unfortunately, the copy past implementation in the lite will be required in this case too;

2. Leave it in ComposeAppManager::getCurrentHash().

3. Introduce another pacman, e.g. TestOstreePackmanager that inherits ComposeAppManager and overrides getCurrentHash();

4. Introduce some kind of policy or just a simple boolean flag that indicates if we are at test mode and it's optional to be booted into the ostree roots. That flag can be defined via a configuration variable (sota.toml) or compile-time definition;

Also, we might consider adding variables and logic that would help us optionally not to pull the ostree update or just achieve it by specifying the same checksum/hash in Target as we have in the ostree repo or just simply return a desired value/check in the overridden getCurrentHash of some test package manager.


mike-sul:
I think this given approach is flexible enough to fulfill different use-cases,
1. make sure that we get hash/checksum of a booted sysroot
2. allows running aklite and doing update locally for non-booted sysroot
3. integration tests, that work against a real non-booted sysroot 
4. unit tests that don't need actual sysroot, just returning a specific hash/checksum.

src/helpers.cc
```diff
@@ -266,11 +273,17 @@ LiteClient::LiteClient(Config& config_in)
   primary_ecu = ecu_serials[0];
 
   std::vector<std::string> headers;
-  GObjectUniquePtr<OstreeSysroot> sysroot_smart = OstreeManager::LoadSysroot(config.pacman.sysroot);
-  OstreeDeployment* deployment = ostree_sysroot_get_booted_deployment(sysroot_smart.get());
+  bool booted_sysroot = true;
+  if (config.pacman.extra.count("booted") == 1) {
+    booted_sysroot = boost::lexical_cast<bool>(config.pacman.extra.at("booted"));
```
doanac, line 278: neat. i do some boolean stuff on config elsewhere that could use this boost helper.

tests/helpers_test.cc
```diff
@@ -279,38 +293,38 @@ TEST(helpers, containers_initialize) {
...
+//  ASSERT_FALSE(client.dockerAppsChanged());
+
+//  // Change the content
+//  Utils::writeFile(cfg_dir / "foo.txt", std::string("foo text content changed"));
+//  ASSERT_TRUE(client.dockerAppsChanged());
+
+//  // Disable and ensure we detect the change
+//  apps_cfg["docker_app_params"] = "";
+//  ASSERT_TRUE(createClient(cfg_dir, apps_cfg).dockerAppsChanged());
+//  ASSERT_FALSE(boost::filesystem::exists(cfg_dir / ".params-hash"));
```
doanac, line 327: is this logic breaking your tests or something?

mike-sul: It's the docker app manager specific tests, that pacman type is not used anyway, so I just comment them, I can either remove them or create another test and initialize client with the docker app manager so these tests will succeed?

doanac: oh - i see. a new test would probably be cleaner anyway. thanks

mike-sul: added a new test.

PR #13, opened 2020-06-02
Don't (re-)start apps containers if reboot

mike-sul:
This change makes compose apps's containers not to (re-)start
just after their update and (re-)creation if we are about to reboot a
system in order to apply an ostree-based rootfs update.
The updated apps will be started just after a system boot if aklite runs in a daemon mode.
If it is at a non-daemon mode then the updated apps will be started on a first run
of aklite or they can be started manually.

### DIFF
--- a/src/composeapp.cc
+++ b/src/composeapp.cc
@@ -22,7 +22,17 @@ bool ComposeApp::fetch(const std::string& app_uri) {
   return false;
 };
 
-bool ComposeApp::start() { return cmd_streaming(compose_ + "up --remove-orphans -d"); }
+bool ComposeApp::up(bool no_start) {
+  auto mode = no_start ? "--no-start" : "-d";
+
+  if (no_start) {
+    boost::filesystem::ofstream flag_file(root_ / NeedStartFile);
+  }
+
+  return cmd_streaming(compose_ + "up --remove-orphans " + mode);
+}
+
+bool ComposeApp::start() { return cmd_streaming(compose_ + "start"); }
 
 void ComposeApp::remove() {
   if (cmd_streaming(compose_ + "down")) {
--- a/src/composeapp.h
+++ b/src/composeapp.h
@@ -12,12 +12,14 @@ namespace Docker {
 class ComposeApp {
  public:
   static constexpr const char* const ArchiveExt{".tgz"};
+  static constexpr const char* const NeedStartFile{".need_start"};
 
  public:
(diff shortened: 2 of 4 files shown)


doanac:
I think this makes sense. Let's get @mike-scott to take a look. There are some edge cases he'd come up with at one point and can't remember them all now.

mike-sul:
> I think this makes sense. Let's get @mike-scott to take a look. There are some edge cases he'd come up with at one point and can't remember them all now.

I've considered a few implementation options and the given one seems like the simplest (KISS) and requires the least changes. It's far from been perfect but in the light of going towards the ostree-backed storage for docker containers, it might be good enough for now.

mike-sul:
@doanac @mike-scott added tests as well as functionality that makes aklite skip Target and Apps installation if it's already installed. Not sure if we need to check if Target's app is removed on a file system. (immutable Target requires immutable volume/mount for its components, ostree and Apps :))

tests/composeapp_test.cc
```diff
@@ -388,25 +418,117 @@ TEST(ComposeApp, handleRemovedApps) {
...
+  // skipping an App update install because already installed
+  {
+    installed_target_json["custom"]["docker_compose_apps"]["app1"]["uri"] = "hub.io@factory/app1:sha256:12312312312";
+    Uptane::Target installed_target("pull", installed_target_json);
+
+    TestClient client("app1", &installed_target);
+
+    Json::Value target_to_install_json{installed_target_json};
+    target_to_install_json["custom"]["docker_compose_apps"]["app1"]["uri"] = "hub.io@factory/app1:sha256:12312312312";
+
```
doanac, line 476: do we need to call pacman->fetch here so we can somehow assert we aren't trying to fetch the app?

mike-sul: Yeah, I think, you're right, for some reason, I thought that install() is called on the client not pacman and triggers fetch which obviously is not so, thanks for pointing into it.

mike-sul: @doanac I updated the test as well as some other tests. In general, I see lots of things to improve in the test implementation but not sure if it makes sense to invest time & effort in it at this stage (startup) :).

doanac: I think your approach so far is spot on. We keep chipping away at this as we add/improve things. This is already a tremendous amount of progress.

doanac:
LGTM - +1 - merge when you are ready

tests/composeapp_test.cc
mike-sul: thanks

PR #14, opened 2020-06-09
Turn off color in docker log output

mike-sul:

### DIFF
--- a/src/composeapp.cc
+++ b/src/composeapp.cc
@@ -6,6 +6,8 @@
 
 namespace Docker {
 
+const std::string ComposeApp::Cmd::Up{"up --remove-orphans --no-color "};
+
 ComposeApp::ComposeApp(std::string name, const boost::filesystem::path& root_dir, const std::string& compose_bin,
                        const Docker::RegistryClient& registry_client)
     : name_{std::move(name)}, root_{root_dir / name_}, compose_{compose_bin}, registry_client_{registry_client} {}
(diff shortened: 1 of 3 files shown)


mike-sul:
It helps to prevent outputting [BLOB] in the logs, but it suppresses outputting docker image pulling status, though.

src/composeapp.cc
```diff
@@ -6,6 +6,9 @@
 
 namespace Docker {
 
+const std::string ComposeApp::Cmd::Up{"up --remove-orphans --no-color --quiet-pull "};
+const std::string ComposeApp::Cmd::Pull{"pull -q"};
+
```
doanac, line 11: The function that outputs things in color appears to be this:

 https://github.com/docker/compose/blob/master/compose/progress_stream.py#L19

I wonder if we can launch docker-compose in a way such that stdout/stderr aren't tty's?

mike-sul: > I wonder if we can launch docker-compose in a way such that stdout/stderr aren't tty's?

at first glance, it looks like it will disabled a progress output which I have already disabled by using `-q/--quiet-pull` option. What causes an issue is that 'done' is output in green. In any case, it doesn't help us to have a proper output of the progress log messages it will just turn them off.

doanac: good point. just a little worried we won't have any progress on the pull. i have a container that takes about 90s to pull down and extract. maybe we shouldn't worry to much now since we have a plan to get away from "docker-compose pull" anyway.

@ricardosalveti - what do you think?

ricardosalveti: Yeah, guess it might be better to keep the blob output for now (which can be visualized with journald all). Thanks for investigating this issue anyway.

PR #15, opened 2020-06-12
Fix a race condition in the event reporting

mike-sul:
1) It turned out that there are two instances of ReportQueue running in the
context of aktualizr-lite. One instance is owned by LiteClient the
second one by SotaUptaneClient. Each instance runs a thread that fetches
events from the DB and sends them to the backend, this procedure is not
synchronized what causes the race condition with side effects like
sending the same event twice.
The key question here is why do we need both clients LiteClient and SotaUptaneClient?
Taking into account that SotaUptaneClient is tuned for the Uptane
use-case it doesn't make sense to use it as it requires hacking around in order
to eliminate the Uptane specific logic that is not required in the TUF
case as well as the logic around Secondary support. Instead we can just
reuse the libaktualizr's mechanisms in LiteClient in the way required to fulfil our
specific use-cases. Doing so helps us to fix the immediate issue with
two ReportQueue instances as well as achieve the strategic goal with
minimizing dependency on and vulnerability to changes in libaktualizr.

2) Fix sending the installation applied event
    
    The event EcuInstallationApplied should be sent to the backend if an
    ostree update is successfully fetched and installed (checkout but system
    is not booted into it yet). The issue was that a system reboot was
    invoked in the context of running LiteClient without doing
    its tearing down prior to it. As result, the event was either persisted in
    the DB and sent after reboot or sent actually before reboot if
    ReportQueue's thread managed to fetch and sent it before reboot
    or simply lost due to the race condition caused by two active ReportQueue
    instances and on-going reboot.
    This fix just simply lets LiteClient to deinitialize/teardown its context
    properly including flushing ReportQueue before triggering a system
    reboot.

3) Fix comparision of a new and installed Targets
    
    A list of apps specified in the sota.toml configuration
    as well as what is actually located on a file system should be
    taken into account during comparision of the new and currently
    installed Targets. Otherwise, aklite will keep trying to fetch and
    install Target that is actually installed or never re-install an app
    that has been removed from a file system.
    Use cases:
     - a list of apps in the config is subset of Target's list of apps
     - an app has been removed from a file system

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,9 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # OSTree is mandatory for aktualizr-lite
 option(BUILD_OSTREE "Set to ON to compile with ostree support" ON)
-option(BUILD_DOCKERAPP "Set to ON to compile with package manager support of docker-app" OFF)
+option(BUILD_DOCKERAPP "Set to ON to compile with package manager support of docker-app" ON)
+# always build with DOCKERAPP for aktualizr-lite, optional only for aktualizr
+set(BUILD_DOCKERAPP ON)
 option(BUILD_WITH_CLANG_TIDY "Set to ON to do clang-tidy during source compilation" OFF)
 
 # If we build the sota tools we don't need aklite (???) and vice versa
--- a/src/helpers.cc
+++ b/src/helpers.cc
@@ -256,7 +256,6 @@ LiteClient::LiteClient(Config& config_in)
   primary_ecu = ecu_serials[0];
 
   std::vector<std::string> headers;
-  bool booted_sysroot = true;
   if (config.pacman.extra.count("booted") == 1) {
     booted_sysroot = boost::lexical_cast<bool>(config.pacman.extra.at("booted"));
   }
@@ -554,6 +553,7 @@ data::ResultCode::Numeric LiteClient::install(const Uptane::Target& target) {
   if (iresult.result_code.num_code == data::ResultCode::Numeric::kNeedCompletion) {
     LOG_INFO << "Update complete. Please reboot the device to activate";
     storage->savePrimaryInstalledVersion(target, InstalledVersionUpdateMode::kPending);
+    is_reboot_required_ = booted_sysroot;
   } else if (iresult.result_code.num_code == data::ResultCode::Numeric::kOk) {
     LOG_INFO << "Update complete. No reboot needed";
     storage->savePrimaryInstalledVersion(target, InstalledVersionUpdateMode::kCurrent);
--- a/src/helpers.cc
+++ b/src/helpers.cc
@@ -53,23 +53,6 @@ void log_info_target(const std::string& prefix, const Config& config, const Upta
   }
 }
 
-static void registerComposeAppManager(const std::shared_ptr<OSTree::Sysroot>& sysroot) {
-  static bool is_registered{false};
-
-  if (is_registered) {
-    return;
-  }
-
-  PackageManagerFactory::registerPackageManager(
-      ComposeAppManager::Name,
-      [sysroot](const PackageConfig& pconfig, const BootloaderConfig& bconfig,
-                const std::shared_ptr<INvStorage>& storage, const std::shared_ptr<HttpInterface>& http) {
-        return new ComposeAppManager(pconfig, bconfig, storage, http, sysroot);
-      });
-
-  is_registered = true;
-}
-
 static void add_apps_header(std::vector<std::string>& headers, PackageConfig& config) {
   if (config.type == PACKAGE_MANAGER_OSTREEDOCKERAPP) {
     DockerAppManagerConfig dappcfg(config);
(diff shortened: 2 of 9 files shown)


src/helpers.cc
```diff
@@ -312,10 +296,17 @@ LiteClient::LiteClient(Config& config_in)
   KeyManager keys(storage, config.keymanagerConfig());
   keys.copyCertsToCurl(*http_client);
 
-  registerComposeAppManager(ostree_sysroot);
-
-  primary =
-      std::make_shared<SotaUptaneClient>(config, storage, http_client, nullptr, primary_ecu.first, primary_ecu.second);
+  // TODO: consider improving this factory method
+  if (config.pacman.type == ComposeAppManager::Name) {
```
doanac, line 300: I think you'll need an `ifdef BUILD_DOCKERAPP` around this, or some other cleaner way.

src/helpers.cc
```diff
@@ -53,23 +53,6 @@ void log_info_target(const std::string& prefix, const Config& config, const Upta
   }
 }
 
-static void registerComposeAppManager(const std::shared_ptr<OSTree::Sysroot>& sysroot) {
```
doanac, line 56: So I guess we never needed to register this, and the package manager already knew about it?

doanac: n/m - i see  this isn't needed because we are basically doing our own factory logic now. makes sense.

src/helpers.cc
```diff
@@ -312,10 +296,17 @@ LiteClient::LiteClient(Config& config_in)
...
+  if (config.pacman.type == ComposeAppManager::Name) {
+    package_manager_ =
+        std::make_shared<ComposeAppManager>(config.pacman, config.bootloader, storage, http_client, ostree_sysroot);
+  } else if (config.pacman.type == PACKAGE_MANAGER_OSTREEDOCKERAPP) {
+    package_manager_ = std::make_shared<DockerAppManager>(config.pacman, config.bootloader, storage, http_client);
+  } else if (config.pacman.type == PACKAGE_MANAGER_OSTREE) {
+    package_manager_ = std::make_shared<OstreeManager>(config.pacman, config.bootloader, storage, http_client);
+  } else {
+    throw std::runtime_error("Unsupported package manager type: " + config.pacman.type);
+  }
```
doanac, line 309: This is turning into a pretty big function. I wonder if we should break this piece out into its own "mkpkg_mgr" helper? Might make adding unit tests for this specific thing easier if we ever got to it.

src/helpers.cc
mike-sul: what about compiling everything and just let choose a package manager type by using the configuration variable. Effectively,  lmp-device-register could drive it.

src/helpers.cc
mike-sul: If we remove PACKAGE_MANAGER_OSTREEDOCKERAPP in the future it will be just two 'if's.
Also, I am thinking about having just one type of package manager in the future just PACKAGE_MANAGER_OSTREE and introduce another component((s) responsible specifically for the app installation and a new composite component that would compose or aggregate the aforementioned objects (prefer composition over inheritance :)).

src/helpers.cc
mike-sul: yes

mike-sul:
> This should allow us to drop some of our "[fio extras]" patches from our aktualizr tree which is also great.

I suppose you mean 3ee742ad and 43d47c48?

src/helpers.cc
doanac: +100

src/helpers.cc
doanac: sounds good.

src/main.cc
```diff
@@ -222,6 +222,8 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
           // since reboot is required to apply/finalize the currently installed update (aka pending update)
           break;
         }
+      } else {
+        LOG_INFO << "A device is up-to-date";
```
doanac, line 226: Can probably say either "Device is up-to-date" or "The device is up-to-date"

mike-sul: ok

mike-sul:
for some reason reboot doesn't happen, looking into it.

mike-sul:
@doanac I think it's good to go
@EmbeddedAndroid It should fix the issues you reported.

CMakeLists.txt
```diff
@@ -15,7 +15,9 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # OSTree is mandatory for aktualizr-lite
 option(BUILD_OSTREE "Set to ON to compile with ostree support" ON)
-option(BUILD_DOCKERAPP "Set to ON to compile with package manager support of docker-app" OFF)
+option(BUILD_DOCKERAPP "Set to ON to compile with package manager support of docker-app" ON)
+# always build with DOCKERAPP for aktualizr-lite, optional only for aktualizr
+set(BUILD_DOCKERAPP ON)
```
doanac, line 20: Is there any need to keep this around?

Also - we have to remember that the update to our meta-updater will require removing this as well.

mike-sul: I think we need BUILD_DOCKERAPP until the docker app manager implementation is in libaktualizr.

mike-sul:
@doanac I suppose, we can merge it.

doanac:
yes. +1

PR #17, opened 2020-06-18
Don't reboot a device if non-daemon mode

mike-sul:

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -351,7 +351,9 @@ int main(int argc, char* argv[]) {
       LOG_DEBUG << "Running " << (*cmd_to_run).first;
       LiteClient client(config);
       ret_val = (*cmd_to_run).second(client, commandline_map);
-      is_reboot_required = client.isRebootRequired();
+      if (cmd == "daemon") {
+        is_reboot_required = client.isRebootRequired();
+      }
     }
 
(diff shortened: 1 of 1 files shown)


mike-sul:
@mike-scott @doanac This is a cowboy/hot fix to not reboot a device if aklite runs in a non-daemon mode.
I feel like I would like to refactor code around that fix/place but it will take some time, so I'll do it along the way while adding new functionality.

src/main.cc
```diff
@@ -351,7 +351,9 @@ int main(int argc, char* argv[]) {
       LOG_DEBUG << "Running " << (*cmd_to_run).first;
       LiteClient client(config);
       ret_val = (*cmd_to_run).second(client, commandline_map);
-      is_reboot_required = client.isRebootRequired();
+      if (cmd == "daemon") {
+        is_reboot_required = client.isRebootRequired();
+      }
```
doanac, line 356: maybe add an "else { you need to reboot now }"?

mike-sul: A similar message is yielded before this place https://github.com/foundriesio/aktualizr-lite/blob/ead98c90887c71aa15b4993036b52f671149ce12/src/helpers.cc#L530

doanac: ah - okay. looks good then. lets ship it

PR #18, opened 2020-06-22
Fix and improve Target matching

mike-sul:

### DIFF
--- a/src/helpers.cc
+++ b/src/helpers.cc
@@ -577,7 +577,7 @@ bool LiteClient::isTargetCurrent(const Uptane::Target& target) const {
 // at least the logic that gets a list of apps and apps root folder since they are package manager specific.
 
 bool targets_eq(const Uptane::Target& t1, const Uptane::Target& t2, bool compareDockerApps) {
-  if (!t1.MatchTarget(t2)) {
+  if (!match_target_base(t1, t2)) {
     return false;
   }
 
@@ -645,4 +645,25 @@ bool known_local_target(LiteClient& client, const Uptane::Target& t, std::vector
     }
   }
   return known_target;
+}
+
+bool match_target_base(const Uptane::Target& t1, const Uptane::Target& t2) {
+  if (t1.type() != t2.type()) {
+    return false;
+  }
+
+  if (t1.length() != t2.length()) {
+    return false;
+  }
+
+  if (t1.filename() != t2.filename()) {
+    return false;
+  }
+
+  // match just OSTREE commit
+  if (t1.sha256Hash() != t2.sha256Hash()) {
+    return false;
+  }
+
+  return true;
 }
--- a/src/main.cc
+++ b/src/main.cc
@@ -70,6 +70,15 @@ static std::unique_ptr<Uptane::Target> find_target(LiteClient& client,
   bool find_latest = (version == "latest");
   std::unique_ptr<Uptane::Target> latest = nullptr;
   for (auto& t : client.allTargets()) {
+
+    if (!t.IsValid()) {
+      continue;
+    }
+
+    if (!t.IsOstree()) {
+      continue;
+    }
+
     if (!target_has_tags(t, tags)) {
       continue;
     }
@@ -92,7 +101,6 @@ static std::unique_ptr<Uptane::Target> find_target(LiteClient& client,
 }
 
 static data::ResultCode::Numeric do_update(LiteClient& client, Uptane::Target target) {
-  target.InsertEcu({client.primary_ecu.first, client.primary_ecu.second});
   generate_correlation_id(target);
 
   data::ResultCode::Numeric rc = client.download(target);
@@ -122,10 +130,6 @@ static int update_main(LiteClient& client, const bpo::variables_map& variables_m
   }
   LOG_INFO << "Finding " << version << " to update to...";
   auto target_to_install = find_target(client, hwid, client.tags, version);
-  if (target_to_install == nullptr) {
-    LOG_INFO << "Already up-to-date";
-    return 0;
-  }
 
   if (!force_update) {
     auto currently_installed_target = client.getCurrent();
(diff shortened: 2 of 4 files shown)


src/helpers.cc
```diff
@@ -646,3 +646,24 @@ bool known_local_target(LiteClient& client, const Uptane::Target& t, std::vector
...
+  if (t1.length() != t2.length()) {
+    return false;
+  }
+
+  if (t1.filename() != t2.filename()) {
+    return false;
+  }
+
+  // match just OSTREE commit
+  if (t1.sha256Hash() != t2.sha256Hash()) {
```
doanac, line 664: This is exactly how i'd always planned to implement this. I have wondered if making it generic like the upstream version does to account for sha512 might make sense. However, ostree can't do sha512, right? So I *think* the simplicity of this approach is better.

src/main.cc
```diff
@@ -70,6 +70,15 @@ static std::unique_ptr<Uptane::Target> find_target(LiteClient& client,
...
   std::unique_ptr<Uptane::Target> latest = nullptr;
   for (auto& t : client.allTargets()) {
+
+    if (!t.IsValid()) {
+      continue;
+    }
+
+    if (!t.IsOstree()) {
+      continue;
+    }
```
doanac, line 80: this is subtle, but I think you are trying to make things operate in a more secure manner? this makes this logic always return a valid target if there is one, right?

src/main.cc
```diff
@@ -92,7 +101,6 @@ static std::unique_ptr<Uptane::Target> find_target(LiteClient& client,
 }
 
 static data::ResultCode::Numeric do_update(LiteClient& client, Uptane::Target target) {
-  target.InsertEcu({client.primary_ecu.first, client.primary_ecu.second});
```
doanac, line 95: Have you tested that update events will still get reported to the server with this removed? Now that we have our own "TufClient" its probably fine, but in the old SotuptaneClient approach, this was required.

doanac:
This looks fine. I think we can merge this. However, even though I think the 2nd commit looks good. I vote for updating the aktualizr recipe in meta-lmp with the the hash of the 1st commit for our v71 update. that way we are *only* pulling in one thing that we know fixes the one thing that was messing us up on Friday. Sound good?

src/helpers.cc
mike-sul: yes, just KISS

src/main.cc
mike-sul: yes, I am gonna improve here and yield some warning/error messages.

src/main.cc
mike-sul: I'll double check

mike-sul:
> This looks fine. I think we can merge this. However, even though I think the 2nd commit looks good. I vote for updating the aktualizr recipe in meta-lmp with the the hash of the 1st commit for our v71 update. that way we are _only_ pulling in one thing that we know fixes the one thing that was messing us up on Friday. Sound good?

The second commit is required too because the first commit causes another issue - it fills ecus_ field  for the current target that is loaded from DB while Target from metadata doesn't have it so their comparison will always return false.

mike-sul:
Also, I cannot properly verify it for the docker app. This fixes the issue that Ricardo ran into, I can check it as it happens before the issue I have :), but I cannot verify it from end-to-end perspective.

mike-sul:
> I think this logic is only determining if an app has been added. The original logic also checks if an app has been removed

In general, I think we should get rid of docker|compose_app param in the config and let it be driven solely by Target so we have KISS :).

When are saying "added/removed" you mean in the config or on a file system?

doanac:
we have compatibility to maintain, but I see my confusion now. `LiteClient::dockerAppsChanged()` is called at startup. I was thinking you were doing something with that. However, this is a different thing.  Now I'm trying to understand what's this is doing that targets_eq isn't.

mike-sul:
> we have compatibility to maintain, but I see my confusion now. `LiteClient::dockerAppsChanged()` is called at startup. I was thinking you were doing something with that. However, this is a different thing. Now I'm trying to understand what's this is doing that targets_eq isn't.

So, it doesn't make sense to check if the app list in config has changed in the update cycle (isTargetCurrent, targets_eq) because aklite has to be restarted in order to apply config changes and if it's restarted then dockerAppsChanged() is called what would trigger an update if any change in the config?

doanac:
> So, it doesn't make sense to check if the app list in config has changed in the update cycle (isTargetCurrent, targets_eq) because aklite has to be restarted in order to apply config changes and if it's restarted then dockerAppsChanged() is called what would trigger an update if any change in the config?

Yes. I think that's where we are getting mixed up here. I'm not sure we need the logic in `isTargetCurrent`. Its taking me some time to unpack and sort through mentally though.

mike-sul:
> > So, it doesn't make sense to check if the app list in config has changed in the update cycle (isTargetCurrent, targets_eq) because aklite has to be restarted in order to apply config changes and if it's restarted then dockerAppsChanged() is called what would trigger an update if any change in the config?
> 
> Yes. I think that's where we are getting mixed up here. I'm not sure we need the logic in `isTargetCurrent`. Its taking me some time to unpack and sort through mentally though.

The only reason I put it there is for the case if the app(s) is removed somehow from a filesystem so isTargetCurrent() would detect it without a need in aklite restart, not sure if it makes sense to have it though.

doanac:
> The only reason I put it there is for the case if the app(s) is removed somehow from a filesystem so isTargetCurrent() would detect it without a need in aklite restart, not sure if it makes sense to have it though. 

Its probably worth keeping around for compose-apps. Its one more way we keep users safe-r. However, i don't think its worth adding for docker-apps.

mike-sul:
> Its probably worth keeping around for compose-apps. Its one more way we keep users safe-r. However, i don't think its worth adding for docker-apps.

ok, agree. I finished testing and see that I need to add dockerAppsChanged() for the non-daemon mode, the rest looks ok.

doanac:
one thing we have missing - i guess we can do it in this PR - is to remove the "force" option for `aktualir-lite update`. We should just force in that mode and assume its a "expert" type decision

mike-sul:
@doanac It should work as we expect now.

PR #20, opened 2020-06-24
Update format of currently installed Target if needed

mike-sul:

### DIFF
--- a/src/helpers.cc
+++ b/src/helpers.cc
@@ -199,6 +199,19 @@ LiteClient::LiteClient(Config& config_in)
   std::string pkey;
   storage = INvStorage::newStorage(config.storage);
   storage->importData(config.import);
+  {
+    // hack: "fix" currently installed version that was imported from installed_versions
+    boost::optional<Uptane::Target> current_version;
+
+    storage->loadPrimaryInstalledVersions(&current_version, nullptr);
+    if (!!current_version && current_version->type() != "OSTREE") {
+      auto custom_data = current_version->custom_data();
+      custom_data["targetFormat"] = "OSTREE";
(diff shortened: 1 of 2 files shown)


ricardosalveti:
> * @ricardosalveti - i generally would say "we've fixed the root issue, lets not carry this hack". However, given the importance of this piece of code to always work, i think its worth discussing.

Should we abort in such situation or should we instead force as this pr is doing?

I'm find having this hack in place, just needs rebase.

mike-sul:
I think a user can encounter with the issue that this fix/PR fixes only if he/she flashes a build made between these two commits https://github.com/foundriesio/aktualizr-lite/commit/382c16380c7f1ad0f53d038dca3c0c9ebf541ca9 and https://github.com/foundriesio/ci-scripts/commit/e8a4565f72d99cc9d6996f295b1c4da7101bba75 which means between Jun 16 and  Jun 24.
I suppose the probability of it decreases by the time, so maybe just wait for a few weeks and then just simply drop this PR, less code is better :)

ricardosalveti:
> I think a user can encounter with the issue that this fix/PR fixes only if he/she flashes a build made between these two commits [382c163](https://github.com/foundriesio/aktualizr-lite/commit/382c16380c7f1ad0f53d038dca3c0c9ebf541ca9) and [foundriesio/ci-scripts@e8a4565](https://github.com/foundriesio/ci-scripts/commit/e8a4565f72d99cc9d6996f295b1c4da7101bba75) which means between Jun 16 and Jun 24.
> I suppose the probability of it decreases by the time, so maybe just wait for a few weeks and then just simply drop this PR, less code is better :)

Sounds good, let's close this one then as we didn't do a release with that commit.

PR #23, opened 2020-07-31
helpers: Fix concurrency problem found with HSMs

doanac:
This code has always been wrong and I think we were seeing curl complain
about this in the past. However, it wasn't problematic until we started
testing with SoftHSM more. The HSM needs critical sections around access
to it from a process. The KeyManager by way of the P11EngineGuard does
this. However, it does on the object instance. By sharing the instance
we get the proper locking required.


ricardosalveti:
Closing in favor of https://github.com/foundriesio/aktualizr-lite/pull/24

PR #25, opened 2020-09-03
Check if App is fully installed and running

mike-sul:
Check if Compose App along with its container images is fully installed as well as if its container images are created and running. If not then force App update.

### DIFF
--- a/src/composeapp.cc
+++ b/src/composeapp.cc
@@ -4,6 +4,8 @@
 
 #include <boost/process.hpp>
 
+#include "utilities/utils.h"
+
 namespace Docker {
 
 ComposeApp::ComposeApp(std::string name, const boost::filesystem::path& root_dir, const std::string& compose_bin,
@@ -40,6 +42,54 @@ void ComposeApp::remove() {
   } else {
     LOG_ERROR << "docker-compose was unable to bring down: " << root_;
   }
+}
+
+// The given implementation is kind of best effort to detect whether App's container images are
+// installed & running with a minimum effort. This is rather a trade-off and is far from being an ideal solution.
+// The given implementation can detect missing container images of App as well as its removed or not created containers.
+// The implementation doesn't detect whether App's containers are stopped or not (provided that container instances
+// exists), for example, docker-compose stop, docker stop, or some kind of container process killing.
+bool ComposeApp::isRunning() const {
+  bool cmd_res{false};
+  std::string cmd_output;
+
+  std::tie(cmd_res, cmd_output) = cmd("config");
+  if (!cmd_res) {
+    LOG_WARNING << "Missing and incomplete App config: " << name_;
+    return false;
+  }
+
+  // calculate a number of container images that a given App consists of
+  std::string::size_type find_pos{0};
+  std::string::size_type line_pos{0};
+  int expected_container_number{0};
+
+  while ((find_pos = cmd_output.find("image:", line_pos)) != std::string::npos) {
+    if (cmd_output.substr(line_pos, (find_pos - line_pos)).find_first_of('#') == std::string::npos) {
+      ++expected_container_number;
+    }
+    line_pos = cmd_output.find('\n', find_pos);
+    ++find_pos;
+  }
+
+  // Get a number of running container images
+  std::tie(cmd_res, cmd_output) = cmd("ps -q");
+  if (!cmd_res) {
+    LOG_WARNING << "Failed to get a list of App's containers: " << name_;
+    return false;
+  }
+
+  int running_container_number =
+      std::count_if(cmd_output.begin(), cmd_output.end(), [](const char& symbol) { return symbol == '\n'; });
+
+  if (running_container_number < expected_container_number) {
+    LOG_DEBUG << "Number of running containers is less than a number of images specified in the compose file"
+              << "; App: " << name_ << "; expected container number: " << expected_container_number
+              << "; number of running containers: " << running_container_number;
+    return false;
+  }
+
+  return true;
 }
 
 // Private implementation
@@ -84,6 +134,14 @@ struct Manifest : Json::Value {
 bool ComposeApp::cmd_streaming(const std::string& cmd) {
   LOG_DEBUG << "Running: " << cmd;
   return boost::process::system(cmd, boost::process::start_dir = root_) == 0;
+}
+
+std::pair<bool, std::string> ComposeApp::cmd(const std::string& cmd) const {
+  auto compose_file = root_ / ComposeFile;
+  std::string out_str;
+  int exit_code = Utils::shell(compose_ + " -f " + compose_file.string() + " " + cmd, &out_str);
+
+  return {(exit_code == EXIT_SUCCESS), out_str};
 }
 
 bool ComposeApp::download(const std::string& app_uri) {
--- a/src/composeapp.h
+++ b/src/composeapp.h
@@ -13,6 +13,7 @@ class ComposeApp {
  public:
   static constexpr const char* const ArchiveExt{".tgz"};
   static constexpr const char* const NeedStartFile{".need_start"};
+  static constexpr const char* const ComposeFile{"docker-compose.yml"};
 
  public:
   ComposeApp(std::string name, const boost::filesystem::path& root_dir, const std::string& compose_bin,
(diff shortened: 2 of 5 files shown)


src/composeapp.cc
```diff
@@ -42,6 +44,68 @@ void ComposeApp::remove() {
...
+    if (!cmd_res) {
+      LOG_WARNING << "Missing and incomplete App config: " << name_;
+      break;
+    }
+
+    // calculate a number of container images that a given App consists of
+    std::string::size_type find_pos{0};
+    int expected_image_number{0};
+
+    while ((find_pos = cmd_output.find("image:", find_pos)) != std::string::npos) {
```
doanac, line 64: I suppose we don't have a YAML parser available. Should be fine. But - you should check to make sure this line doesn't start with a `#`. I could see people doing something like:
~~~
services:
  andy:
    # image: foo-devel
    image: foo
~~~

src/composeapp.cc
```diff
@@ -42,6 +44,68 @@ void ComposeApp::remove() {
...
+        std::count_if(cmd_output.begin(), cmd_output.end(), [](const char& symbol) { return symbol == '\n'; });
+
+    if (installed_image_number < expected_image_number) {
+      LOG_DEBUG << "Number of installed images is less than a number of images specified in the compose file"
+                << "; App: " << name_ << "; expected image number: " << expected_image_number
+                << "; number of installed images: " << installed_image_number;
+      break;
+    }
+
+    // Get a number of running container images
```
doanac, line 86: a bit of a premature optimization. but the docker-compose binary is a little slow to "start". Checking if the image exists and is running is great for debugging, but we really only need this "is it running" check right?

doanac (commented):
some preliminary comments. I'm going to find some time to test this locally today/tomorrow

src/composeapp.cc
```diff
@@ -86,6 +150,14 @@ bool ComposeApp::cmd_streaming(const std::string& cmd) {
   return boost::process::system(cmd, boost::process::start_dir = root_) == 0;
 }
 
+std::pair<bool, std::string> ComposeApp::cmd(const std::string& cmd) {
+  auto compose_file = root_ / ComposeFile;
+  std::string out_str;
+  int exit_code = Utils::shell(compose_ + " -f " + compose_file.string() + " " + cmd, &out_str);
```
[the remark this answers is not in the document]
mike-sul: This should be OK since by default a project dir is equal to the Compose file directory, but I will double-check it.
`--project-directory PATH    Specify an alternate working directory
                              (default: the path of the Compose file)
`

src/composeapp.cc
mike-sul: ok

src/composeapp.cc
mike-sul: an old App can be running while an update of a new App was interrupted, although in this case Target in the DB should not be updated so the next update cycle should start App update process anyway... I will check it out.

src/composeapp.cc
mike-sul: There are C/C++ yaml parsers, of course, I just didn't want to introduce another dependency.

doanac: no - thanks for not!

src/composeapp.cc
mike-sul: removed 'images' check.

src/composeapp.cc
mike-sul: > But - you should check to make sure this line doesn't start with a `#`. I could see people doing something like:

done

src/composeapp.cc
```diff
@@ -42,6 +44,55 @@ void ComposeApp::remove() {
...
+      ++find_pos;
+    }
+
+    // Get a number of running container images
+    std::tie(cmd_res, cmd_output) = cmd("ps -q");
+    if (!cmd_res) {
+      LOG_WARNING << "Failed to get a list of App's containers: " << name_;
+      break;
+    }
+
```
doanac, line 79: I like this approach. Its what I was thinking also. I do think we should add a comment for future reference like:
// NOTE - This logic isn't foolproof. The services could *change* but the total remains the same. This logic does not detect that.

src/composeapp.cc
```diff
@@ -42,6 +44,55 @@ void ComposeApp::remove() {
...
+    if (running_container_number < expected_container_number) {
+      LOG_DEBUG << "Number of running containers is less than a number of images specified in the compose file"
+                << "; App: " << name_ << "; expected container number: " << expected_container_number
+                << "; number of running containers: " << running_container_number;
+      break;
+    }
+
+    result = true;
+  } while (false);
```
doanac, line 91: shouldn't this be `while (true)`?

mike-sul: no, this statement do/while(false) is used for avoiding nested 'ifs' or multiple returns. I think I will remove this statement at all and just `return false` instead of `break`, this trick is justifiable IMHO if there are really lots of returns and the code is more complex.

doanac:
a couple of observations after playing with this some:

1) `docker-compose ps` takes about 2.5 seconds to run. Not a deal breaker, but long term something we'll need to improve.

2) `docker-compose ps` doesn't notice if you've done `docker-compose stop`. This wasn't obvious at first and then made sense to me. It notices if `docker-compose down` was done.

3a) I'm not sure if the logging format is intentional:
~~~
  debug: Skipping Image repo Targets download; stored version is still current.
  debug:  checking if fiotest is installed and running...
   info: Device is up-to-date
~~~

There's a leading space in the "check if if ..." line.

3b) We wind up printing that debug 2 times. Once detecting a change and once applying the update. It makes sense to me technically. Its just one of those things people that try to debug by logfile might stumble on. I don't think its something we can/should change though unless you have some ideas?

The code works for me. That's the important thing.

mike-sul:
>docker-compose ps takes about 2.5 seconds to run. Not a deal breaker, but long term something we'll need to improve. 

Yes, I know, that's why I call it "far from an ideal". A proper solution will consist of two parts. The first part makes sure all required container images are installed and it "ostree for containers" should tackle it. The second part should monitor the running container status. It will take time to implement both of them while we have the immediate issue and the given PR fixes it. 

I'll check out the running time of `docker-compose ps` and `docker-compose images` if there is a significant difference than we might stay just with `docker-compose images`. Also, I can try and make a direct API call to dockerd and see if it helps to speed up the process. Effectively, to fix the issue that Mike and you(Juul) reported it's just enough to this kind of checking on the first boot. Maybe running this check on aktualizr-lite's first Upate Cycle/Loop would be the right trade-off, it fixes our immediate issue and doesn't use much CPU resources on regular bases?

mike-sul:
> 2\. `docker-compose ps` doesn't notice if you've done `docker-compose stop`. This wasn't obvious at first and then made sense to me. It notices if `docker-compose down` was done.

Yeah, I know, the same is true for  `docker stop`, my argument is the same as the above "not ideal" and the proper solution will require lots of time. I'll check out if calling dockerd's API directly can help here and is feasible.

mike-sul:
> I'm not sure if the logging format is intentional:

+, will fix

mike-sul:
> 3b) We wind up printing that debug 2 times. Once detecting a change and once applying the update. It makes sense to me technically. Its just one of those things people that try to debug by logfile might stumble on. I don't think its something we can/should change though unless you have some ideas?

Yeah, I noticed this behavior a long time before, but  I didn't want to "fix" it until we get rid of docker apps (less code) since it doesn't create any tech issue and a default logging prevents from users see it.
My current idea is to store a result of "app change detection" in the context and check the result during the update application to avoid double verification.

src/composeapp.cc
mike-sul: I am not sure if I understand fully the comment. The services can change only via OTA if a new Target is created what triggers an update regardless of the given verification logic. Or do you mean that docker-compose.yml can be modified manually directly on a device?

I think it makes sense to add a comment saying that the given logic cannot detect all cases of App containers stopping (docker-compose stop).

doanac: your comment idea is fine.

mike-sul:
@doanac we could probably use `docker-compose ps  --filter "status=running"` but there is an issue https://github.com/docker/compose/issues/5996, so container filtering doesn't work :), only filtering of services, but service can be made of more than one container...

doanac:
> we could probably use `docker-compose ps --filter "status=running"`

I hit the same thing. I don't think its a huge deal. This is progress and I think is good for now.

doanac:
i think the code is good. My only question would be if its worth just doing it at startup or doing it for each cycle. Let me know what you think.

mike-sul:
> i think the code is good. My only question would be if its worth just doing it at startup or doing it for each cycle. Let me know what you think.

I think, doing it just at aklite startup would be the right trade-off atm.

src/composeapp.cc
mike-sul: ok

PR #26, opened 2020-09-09
Report aktualizr config to server

doanac:

### DIFF
--- a/src/helpers.cc
+++ b/src/helpers.cc
@@ -485,6 +485,8 @@ void LiteClient::reportAktualizrConfiguration() {
         http_client->put(config.tls.server + "/system_info/config", "application/toml", conf_str);
     if (response.isOk()) {
       storage->storeDeviceDataHash("configuration", new_hash.HashString());
+    } else {
+      LOG_DEBUG << "Unable to report libaktualizr configuration: " << response.getStatusStr();
     }
   }
 }
@@ -494,12 +496,13 @@ void LiteClient::reportNetworkInfo() {
     LOG_DEBUG << "Reporting network information";
     Json::Value network_info = Utils::getNetworkInfo();
     if (network_info != last_network_info_reported_) {
-      HttpResponse response = http_client->put(config.tls.server + "/system_info/network", network_info);
+      const HttpResponse response = http_client->put(config.tls.server + "/system_info/network", network_info);
       if (response.isOk()) {
         last_network_info_reported_ = network_info;
+      } else {
+        LOG_DEBUG << "Unable to report network information: " << response.getStatusStr();
       }
     }
-
   } else {
     LOG_DEBUG << "Not reporting network information because telemetry is disabled";
   }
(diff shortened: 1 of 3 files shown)


doanac:
This relates to the ota-lite MR.

doanac:
I need to test this in the morning.

mike-sul:
@doanac Unformatted code

src/main.cc
```diff
@@ -176,6 +176,8 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
     }
 
     if (firstLoop) {
+      client.reportAktualizrConfiguration();
```
mike-sul, line 179: I am wondering if it can be invoked before `while (true)`/Update Loop?

src/helpers.cc
```diff
@@ -467,6 +467,29 @@ std::pair<bool, Uptane::Target> LiteClient::downloadImage(const Uptane::Target&
...
+  std::stringstream conf_ss;
+  config.writeToStream(conf_ss);
+  const std::string conf_str = conf_ss.str();
+  const Hash new_hash = Hash::generate(Hash::Type::kSha256, conf_str);
+  std::string stored_hash;
+  if (!(storage->loadDeviceDataHash("configuration", &stored_hash) &&
+        new_hash == Hash(Hash::Type::kSha256, stored_hash))) {
+    LOG_DEBUG << "Reporting libaktualizr configuration";
+    const HttpResponse response = http_client->post(config.tls.server + "/system_info/config", "application/toml", conf_str);
+    if (response.isOk()) {
```
mike-sul, line 485: maybe print some warning if !response.isOk?

src/helpers.cc
```diff
@@ -467,6 +467,29 @@ std::pair<bool, Uptane::Target> LiteClient::downloadImage(const Uptane::Target&
...
+    return;
+  }
+
+  std::stringstream conf_ss;
+  config.writeToStream(conf_ss);
+  const std::string conf_str = conf_ss.str();
+  const Hash new_hash = Hash::generate(Hash::Type::kSha256, conf_str);
+  std::string stored_hash;
+  if (!(storage->loadDeviceDataHash("configuration", &stored_hash) &&
+        new_hash == Hash(Hash::Type::kSha256, stored_hash))) {
```
mike-sul, line 482: `if (storage->loadDeviceDataHash("configuration", &stored_hash) && new_hash == Hash(Hash::Type::kSha256, stored_hash)) { LOG <<; return;}` 
IMO, the code will be slightly cleaner if we return if the config has not changed.

doanac: Sure. I just took the code as-is from sotauptaneclient.cc.

src/helpers.cc
doanac: we aren't doing this for any of the "report" methods. I'll add a 2nd commit to this PR that does that.

src/helpers.cc
doanac: So i've updated the current first commit of this PR to just do like our other report* methods do. I can then do follow up commits to clean this code.

mike-sul (approved):

src/helpers.cc
doanac: I've tried to clean things up with the latest push. Let me know what you think

mike-sul (approved):

PR #30, opened 2020-09-24
Fix sorting of the "list" subcommand

doanac:
This has always given rather random output. This sorts by the version
number using a flat_map.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1,6 +1,7 @@
 #include <iostream>
 #include <unordered_map>
 
+#include <boost/container/flat_map.hpp>
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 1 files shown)


mike-sul (approved):
LGTM

I was always wondering why does liteclient need to call `Uptane::LazyTargetsList(image_repo_, storage, uptane_fetcher_)` instead of just calling image_repo_->getTargets().

doanac:
> I was always wondering why does liteclient need to call Uptane::LazyTargetsList(image_repo_, storage, uptane_fetcher_) instead of just calling image_repo_->getTargets().

I think before you built our own client, there was no image_repo available.

PR #34, opened 2020-10-12
docker-apps: Remove docker app impl from the lite client

mike-sul:
The second step towards removing the docker-app implementation - removing it just from the Lite client.
The next steps are:
- switching to 2020.9 aktualizr
- removing the docker-app pack manager from aktualizr (cherry-picking Andy's commit)


CMakeLists.txt
```diff
@@ -15,9 +15,6 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # OSTree is mandatory for aktualizr-lite
 option(BUILD_OSTREE "Set to ON to compile with ostree support" ON)
-option(BUILD_DOCKERAPP "Set to ON to compile with package manager support of docker-app" ON)
-# always build with DOCKERAPP for aktualizr-lite, optional only for aktualizr
-set(BUILD_DOCKERAPP ON)
```
doanac, line 20: We still need some option to disable compose-apps. We don't have docker/golang on riscv yet.

src/composeappmanager.cc
```diff
@@ -136,15 +136,38 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
   return apps_to_update;
 }
 
+ComposeAppManager::AppsContainer ComposeAppManager::getAppsToRemove(const Uptane::Target& t) const {
```
doanac, line 139: I'm not sure I see how this change fits in with this commit? This seems orthogonal to removal of docker-app package manager?

CMakeLists.txt
mike-sul: BUILD_DOCKERAPP has to be removed anyway as it doesn't manage the inclusion or exclusion of the compose app manager into a compilation.

>We don't have docker/golang on riscv yet

But it doesn't mean that the compose-apps package manager itself can be built on those platforms.

doanac: true. i guess we just need to make sure we never set that package manager on those platforms. @ricardosalveti - How are we handling riscv related to this?

src/composeappmanager.cc
mike-sul: It's related to removing the `LiteClient::dockerAppsChanged`.  If the docker-apps specific code is removed from that method then only invoking of `appListChanged`, the business logic of which overlaps with `ComposeAppManager::getAppsToUpdate` except determining of apps to be removed hence I added ComposeAppManager::getAppsToRemove.

mike-sul: But, I think, it makes sense to make this change in two steps, at first just remove docker apps as it is and then restructure the code. BTW, it also fixes a bug with app removal if it's removed in new Target but not from the config and aklite is not restarted and running in daemon mode.

PR #35, opened 2020-10-13
Get rid of docker apps

mike-sul:


mike-sul:
@doanac meta-lmp does not use `BUILD_DOCKERAPP` build variable anymore https://github.com/foundriesio/meta-lmp/commit/005a5a701a75a63a2faa99cc69f3f3b876ae2b4d. So the plan is to build the compose apps pack manager unconditionally, whether to use the ostree pack manager or compose apps one is a matter of aklite config. lmp-device-regster configures aklite according to this https://github.com/foundriesio/meta-lmp/blob/master/meta-lmp-base/recipes-sota/lmp-device-register/lmp-device-register_git.bb#L16, so overriding `PACKAGECONFIG ?= "aklitetags composeapp"` is the mechanism to turn on/off the compose app usage on a specific platform/lmp-build.

PR #36, opened 2020-10-13
Switch to aktualizr 2020.9

mike-sul:


doanac:
i'm going to pull this into my factory and do some testing either today or tomorrow.

unit-test
```diff
@@ -14,7 +14,7 @@ build=$(mktemp -d -p $here)
 cd $build
 
 # do clang-tidy during compilation
-../cmake-init.sh -DBUILD_WITH_CLANG_TIDY=ON
+../cmake-init.sh
```
doanac, line 17: should we really remove the ability to run clang tidy, or just disable it by default for CI until we can fix?

mike-sul: I just removed the clang tidy verification that happens during the source code building, I added it initially and it's kind of duplication of what the aktualizr clang tidy functionality does. So there is still clang tidy the one that aktualizr provides.

doanac:
looks sane. i'll give this a test tomorrow

mike-sul:
> looks sane. i'll give this a test tomorrow

Thanks, I'll do the same today too.

PR #37, opened 2020-10-14
fixup: Remove apps from the list safely

mike-sul:


mike-sul:
@doanac This is the fix for the issue. The test fails because I switched the container image used for testing to the newer version with clang-tidy v 10.

mike-sul:
@doanac I built the old container and pushed it to the docker hub, so the test can be executed.

doanac (approved):

PR #38, opened 2020-10-16
aklite: Print device UUID and name

mike-sul:
- Output device UUID and name on `aktualizr-lite status` call.

I will add a device name printing once this PR https://github.com/foundriesio/device-gateway/pull/3/ is merged.


doanac (approved):
That was easier than i expected

doanac:
I guess we should set a uuid in the db so the unit test doesn't fail. just noticed that.

mike-sul:
@doanac Added getting and printing a device name.

doanac (approved):

PR #31, opened 2020-09-29
Add ability manually install old Targets

doanac:
This is a handy feature for debug/test. Sometimes you want to "go back
in time". However, the installed-versions DB prevents this. This new
flag allows you to:
~~~
 # You are on Target 5.
 $ aktualizr-lite update --update-name 4 --clear-installed-versions
 # Update will work, you are on Target 4.
 $ aktualizr-lite daemon  # move back to target 5
~~~
Signed-off-by: Andy Doan <andy@foundries.io>

PR #39, opened 2020-11-02
Add ability manually install old Targets

mike-sul:
This is a handy feature for debug/test. Sometimes you want to "go back
in time". However, the installed-versions DB prevents this. This new
flag allows you to:

 # You are on Target 5.
 $ aktualizr-lite update --update-name 4 --clear-installed-versions
 # Update will work, you are on Target 4.
 $ aktualizr-lite daemon  # will update to Target 5

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -159,6 +159,12 @@ static int update_main(LiteClient& client, const bpo::variables_map& variables_m
     version = variables_map["update-name"].as<std::string>();
   }
 
+  // This is only available if -DALLOW_MANUAL_ROLLBACK is set in the CLI args below.
+  if (variables_map.count("clear-installed-versions") > 0) {
+    LOG_WARNING << "Clearing installed version history!!!";
+    client.storage->clearInstalledVersions();
+  }
+
   LOG_INFO << "Finding " << version << " to update to...";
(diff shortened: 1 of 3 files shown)


mike-sul:
original PR https://github.com/foundriesio/aktualizr-lite/pull/31

src/main.cc
```diff
@@ -159,6 +159,12 @@ static int update_main(LiteClient& client, const bpo::variables_map& variables_m
     version = variables_map["update-name"].as<std::string>();
   }
 
+  // This is only available if -DALLOWED_OPTIONS is set in the CLI args below.
```
ricardosalveti, line 162: Comment here should be if -DALLOW_MANUAL_ROLLBACK is set, no?

Wonder if we should also protect this block with an ifdef.

doanac: Its totally possible. Its essentially dead code, but I did like that its *always* compiled so we'll know of build breaks

mike-sul: > Comment here should be if -DALLOW_MANUAL_ROLLBACK is set, no?

corrected.

ricardosalveti: Sure, wfm.

PR #40, opened 2020-11-04
Install all Target Apps if compose_apps is not specified

mike-sul:
- If `compose_apps` is not specified in sota.toml then install all
Target's Apps;
- If `compose_apps` is specified in sota.toml then an intersection of
Target's Apps and the compose_apps' list is installed.
- The previous statement implies that an empty `compose_apps` instructs
to NOT install any Target's Apps at all.


mike-sul:
@doanac What do you think about it?

doanac (approved):

PR #41, opened 2020-11-24
Fix/reporting target to backend

mike-sul:
- Set the correct value of `x-ats-dockerapps` and `x-ats-tags` headers that are sent to the backend;
  - If compose_apps is not specified in the config then add all Target's Apps to the header;

- Report correct current Target at aklite startup
  -  Correctly determine currently installed Target at aklite startup if a rollback has happened or a system was not rebooted after a system update, just aklite restarted.


mike-sul:
@ricardosalveti I tested it by emulating a rollback manually running aklite process on a local machine. It would be great to test a real device.

ricardosalveti:
Looks good, let me do some testing now.

ricardosalveti (approved):
LGTM, tested on hardware, working as expected.

PR #43, opened 2020-12-09
Initial prototype of OSTree-based Compose Apps

mike-sul:


mike-sul:
@doanac This is part 1.3 of the implementation plan https://docs.google.com/document/d/1lQs5Sas0WwVFzcaH4v-UYelgvaYn6_PBkEGs-Ndop5I/edit#heading=h.e4bsxuh2k3zs

src/ostree/repo.h
```diff
@@ -0,0 +1,39 @@
...
+
+
+namespace OSTree {
+
+class Repo {
+ public:
+  using Ptr = std::shared_ptr<Repo>;
+
+ public:
+  static Ptr create(std::string path) { return std::make_shared<Repo>(std::move(path)); }
```
doanac, line 16: i'm not too up on c++ idioms. This seemed odd to me though. Seems like a user can use the make_shared/make_unique calls directly? The reason i noticed, was that I'd think this might be a std::unique instead of std::shared pointer.

src/composeappmanager.cc
```diff
@@ -182,10 +192,18 @@ bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetche
...
 
   bool passed = true;
-  for (const auto& pair : cur_apps_to_fetch_and_update_) {
-    LOG_INFO << "Fetching " << pair.first << " -> " << pair.second;
-    if (!app_ctor_(pair.first).fetch(pair.second)) {
-      passed = false;
+  const auto& branch_to_pull = target.custom_data()["compose-apps-branch"].asString();
+  const auto& commit_hash_to_pull = target.custom_data()["compose-apps-hash"].asString();
+  if (!branch_to_pull.empty()) {
+    LOG_INFO << "Pulling from App Tree; branch: " << branch_to_pull << " commit: " << commit_hash_to_pull;
```
doanac, line 198: as this develops, will you need to also assert that the device was configured to use ostree apps?

src/composeappmanager.cc
```diff
@@ -182,10 +192,18 @@ bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetche
...
+    app_tree_.pull(config.ostree_server, keys, commit_hash_to_pull);
+    app_tree_.pull(config.ostree_server, keys, branch_to_pull);
```
doanac, line 200: wouldn't pulling the commit hash be enough? what is pulling the branch doing?

doanac (commented):
I was expecting a *lot* more code to handle this. This isn't so scary so far :)

src/ostree/repo.cc
```diff
@@ -0,0 +1,138 @@
...
+                                   "unix::device,unix::inode,unix::mode,unix::uid,unix::gid,unix::rdev")
+
+
+
+template<class Func, class ...Args>
+void libostree_call(Func func, Args... args) {
+  g_autoptr(GError) error = nullptr;
+  //std::cout << FuncName << std::endl;
+  if (!func(args..., &error)) {
+    throw std::runtime_error("FuncName");
```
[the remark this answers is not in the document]
mike-sul: This code is actually not present in the second/last commit, it was part of the original version, but then I decided to remove it.

src/composeappmanager.cc
mike-sul: to update the branch/ref in the local ostree repo, otherwise, a user won't see the current/latest commit via `ostree --repo=<> refs`.

src/composeappmanager.cc
mike-sul: I think about just doing fallback to the current way to pull apps. 
>the device was configured to use ostree apps?

Do we wanna another configuration variable? Or if an app's ostree repo exists/preloaded in a system image and Target has a commit to the app's ostree repo in Treehub then do the ostree-cased update otherwise fallback?

src/ostree/repo.h
mike-sul: Whether unique and shared pointer is needed really depends on an use-case, but in this particular case, we don't need this factory method at all, going to remove it.

src/composeappmanager.cc
doanac: I was hoping we could avoid the config variable and determine by the existence of the local repo

mike-sul: ok, what entity should create the local repo, the preloading functionality, or the preloading and lmp-device-register if the repo is not preloaded?

mike-sul:
@doanac Updated according to your comments. I actually added a configuration variable to force the ostree repo creation because it's really inconvenient to test without it and I think it will be useful in the future.

src/composeapp.cc
```diff
@@ -36,6 +36,7 @@ bool ComposeApp::up(bool no_start) {
 }
 
 bool ComposeApp::start() { return cmd_streaming(compose_ + "start"); }
+bool ComposeApp::stop() { return cmd_streaming(compose_ + "stop"); }
```
doanac, line 39: this seems to be unused?

doanac (commented):
i have one dumb question, but this appears pretty harmless to our core code path, so lets get this merged soon.

src/composeapp.cc
mike-sul: yes, removed and updated PR via forced push.

mike-sul:
Rebased on the latest master and updated PR.

doanac:
is it time to finally merge this magic?

mike-sul:
@doanac I think I am ready to merge it :).

doanac:
lets do it.

PR #46, opened 2021-01-21
[RFC] Drop use of docker binary for inspecting containers

doanac:
An idea I'm toying with. Let me know what you think. We could probably do some further simplifications based on this.

NOTE: it requires a patch to aktualizr:
~~~
$ git diff
diff --git a/src/libaktualizr/http/httpclient.h b/src/libaktualizr/http/httpclient.h
index f2b329ce..2b24d943 100644
--- a/src/libaktualizr/http/httpclient.h
+++ b/src/libaktualizr/http/httpclient.h
@@ -45,11 +45,13 @@ class HttpClient : public HttpInterface {
                 const std::string &pkey, CryptoSource pkey_source) override;
   bool updateHeader(const std::string &name, const std::string &value);
 
+ protected:
+  CURL *curl;
+
  private:
   FRIEND_TEST(GetTest, download_speed_limit);
 
   static CurlGlobalInitWrapper manageCurlGlobalInit_;
-  CURL *curl;
   curl_slist *headers;
   HttpResponse perform(CURL *curl_handler, int retry_times, int64_t size_limit);
   static curl_slist *curl_slist_dup(curl_slist *sl);
~~~


doanac:
@ldts - took the idea but made it much better.

PR #47, opened 2021-01-25
Wait for the 'interval' if metadata update fails

mike-sul:
If the check for new TUF metadata fails then wait for the TUF cycle
interval time which is specified in sota.toml. In other words, don't
start the next update cycle straight away if TUF metadata update fails,
wait for a configurable amount of time before doing it.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -208,8 +208,9 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
     LOG_INFO << "Active Target: " << current.filename() << ", sha256: " << current.sha256Hash();
     LOG_INFO << "Checking for a new Target...";
     if (!client.checkForUpdates()) {
-      LOG_WARNING << "Unable to update latest metadata";
-      std::this_thread::sleep_for(std::chrono::seconds(10));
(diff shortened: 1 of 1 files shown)


ricardosalveti (approved):
LGTM

ricardosalveti:
I can confirm that this works as expected, and curl indeed tries a few times before giving up:

```
   info: Active Target: raspberrypi3-64-lmp-221, sha256: 6f9b8912b7325fd2d0a746e95ad82f860991ad70c7b9e30544fd3847c3bb0d59
   info: Checking for a new Target...
  error: curl error 6 (http code 0): Couldn't resolve host name
  error: curl error 6 (http code 0): Couldn't resolve host name
  error: curl error 6 (http code 0): Couldn't resolve host name
  error: curl error 6 (http code 0): Couldn't resolve host name
  error: curl error 6 (http code 0): Couldn't resolve host name
  error: curl error 6 (http code 0): Couldn't resolve host name
  error: Failed to update Image repo metadata: Failed to fetch role timestamp in image repository.
warning: Unable to update latest metadata, going to sleep for 30 seconds before starting a new update cycle
```

PR #49, opened 2021-01-28
ostree-apps: catch pull and checkout exceptions

mike-sul:
LiteClient expects error code/boolean as a return code and use it to
determine Target download and install status. We need to adjust the
ostree apps implementation to the given protocol.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -204,7 +204,14 @@ bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetche
   const auto& apps_uri = target.custom_data()["compose-apps-uri"].asString();
   if (app_tree_ && !apps_uri.empty()) {
     LOG_INFO << "Fetching Apps Tree -> " << apps_uri;
-    app_tree_->pull(config.ostree_server, keys, apps_uri);
+
+    try {
+      app_tree_->pull(config.ostree_server, keys, apps_uri);
+    } catch (const std::exception& exc) {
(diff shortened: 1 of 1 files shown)


src/composeappmanager.cc
```diff
@@ -240,7 +247,13 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
   const auto& apps_uri = target.custom_data()["compose-apps-uri"].asString();
   if (app_tree_ && !apps_uri.empty()) {
     LOG_INFO << "Checking out updated Apps: " << apps_uri;
-    const_cast<ComposeAppManager*>(this)->app_tree_->checkout(apps_uri);
+    try {
+      const_cast<ComposeAppManager*>(this)->app_tree_->checkout(apps_uri);
+    } catch (const std::exception& exc) {
+      LOG_ERROR << "Failed to checkout Apps from the ostree repo; uri: " << apps_uri << ", err: " << exc.what();
+      return data::InstallationResult(data::ResultCode::Numeric::kInstallFailed,
+                                      "Could not ckeckout Apps from the ostree repo");
```
doanac, line 255: just a typo `ckeckout`. looks good otherwise. 

Its too bad `fetchTarget` only returns a boolean. It would be nice if it did a ResultCode so that we could bubble up more informative information.

mike-sul: fixed the typo.

>Its too bad fetchTarget only returns a boolean. It would be nice if it did a ResultCode so that we could bubble up more informative information.

agree, let's create a ticket for it, I recall we did similar for Uptane Client.

doanac (approved):

PR #52, opened 2021-02-02
aklite: move the client code from the helper files to dedicated source files

mike-sul:


tests/helpers_test.cc
```diff
@@ -94,97 +94,6 @@ TEST(helpers, target_has_tags) {
   ASSERT_FALSE(target_has_tags(t, config_tags));
 }
```
doanac, line 96: i guess these tests aren't useful now?

mike-sul: yes.

doanac (approved):

PR #53, opened 2021-02-02
Refact/drop full status check flag

mike-sul:
Get rid of the flag indicating whether to perform a 'full' status check or not. It turned out to be too complicated and useless as well as the implementation was awkward. This commit drops the given flag and makes aklite perform the full status check unconditionally. The next step is to come up with a lighter/optimized way to perform a status check of running docker containers.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -45,10 +45,6 @@ ComposeAppManager::Config::Config(const PackageConfig& pconfig) {
   if (raw.count("force_update") > 0) {
     force_update = boost::lexical_cast<bool>(raw.at("force_update"));
   }
-
-  if (raw.count("full_status_check") > 0) {
-    full_status_check = boost::lexical_cast<bool>(raw.at("full_status_check"));
-  }
 }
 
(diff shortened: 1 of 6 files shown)


mike-sul:
rebased on the latest master.

mike-sul:
@doanac I think it's good to go, shouldn't introduce any regression.

doanac:
I think this is moving in the right direction. I need to read it a couple more times to really make sense of it. One thing that still bothers me is `cur_apps_to_fetch_and_update_`. I wish there were a way to remove that. We call `getCurrent` in a few places and treat it as a light weight call, but I wonder if `getCurrent` should return the real Target(including the *running* apps). And the `find_target` function were to return the desired target. Then our main loop and update logic is almost can't fail simple.

Probably just follow-up work to this though.

mike-sul:
> Probably just follow-up work to this though.

Yes, that's my plan, I would like to make changes gradually. The given PR is dedicated solely to removing the full status check flag/option. This PR is aimed for the further "simplification" https://github.com/foundriesio/aktualizr-lite/pull/54.

doanac (approved):

PR #54, opened 2021-02-03
WIP: Encapsulate and simplify apps checking

mike-sul:


src/main.cc
```diff
@@ -271,17 +246,13 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
         }
 
       } else {
-        if (!client.appsInSync()) {
-          do_app_sync(client);
-        }
+        do_app_sync(client);
```
doanac, line 249: nothing new, I guess the package-manager will log any problems and its safe to ignore the return-code here?

mike-sul: I've done a few different approaches and decided to go with the one that introduces the least amount of changes, yet simplify logic and encapsulate the app checking within a single scope. I still would like to do many other changes and cleaning but I am afraid to make drastic changes just would like to do it gradually.

>I guess the package-manager will log any problems and it's safe to ignore the return-code here?

There is no any use/value in that return code in the given context, we just inform the backend about the app download/installation status in the context of app sync and continue TUF cycling. This is WIP, I still would like to do a few clean up as well as pass a reason for app syncing/update if no new target.

mike-sul:
@doanac please, look into it, this is what I came with so far. I still would like to try a bit different approach when packman.getCurrent() returns the intersection of what is stored in DB, actually running and is in the shortlist, then the lite client compares it with the desired Target, generates a new Target that includes exactly what needs to be (re-)installed, and sends it the packman (download(0, install()). I think this is what you implied in your comments. I actually, tried a bit the given approach initially but it didn't work well but with the current version I think it should work.

PR #56, opened 2021-02-12
aklite: allow updating to previously successfully installed Target

mike-sul:
Currently aklite doesn't allow to install Target that it tried to install previously. Even if Target installation was successful it treats it as a "bad" Target that caused rollback and doesn't allow installing it.
The given change checks Target to be installed against Targets that aklite tried to install but never installed them except pending one, which can happen only in the case of rollback.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -224,8 +224,8 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
     interval = variables_map["interval"].as<uint64_t>();
   }
 
-  std::vector<Uptane::Target> installed_versions;
-  client.storage->loadPrimaryInstallationLog(&installed_versions, false);
+  std::vector<Uptane::Target> known_but_not_installed_versions;
+  get_known_but_not_installed_versions(client, known_but_not_installed_versions);
 
   client.reportAktualizrConfiguration();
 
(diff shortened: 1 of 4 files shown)


src/main.cc
```diff
@@ -224,8 +224,21 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
...
+
+  std::vector<Uptane::Target> known_but_not_installed_versions;
+  for (const auto& t : known_versions) {
+    if (installed_versions.end() ==
+        std::find_if(installed_versions.begin(), installed_versions.end(),
+                     [&t](const Uptane::Target& t1) { return t.filename() == t1.filename(); })) {
+      // known but never successfully installed version
+      known_but_not_installed_versions.push_back(t);
+    }
+  }
```
doanac, line 241: you and ricardo seem to have hashed this out, so i won't say much. I wonder if you can push this into a helper function that we could then set up a unit test with?

mike-sul: > I wonder if you can push this into a helper function that we could then set up a unit test with?

done.

doanac: thanks. i'll leave it to you and ricardo to decide now.

mike-sul: @ricardosalveti have you had a chance to verify it?

ricardosalveti: Nice, thanks for the tests. Looking good, testing this today and will reply back.

ricardosalveti (approved):
LGTM, tested updating to a newer image, rollback and moving back and updating to what was supposed to be the bad image, all good.

PR #57, opened 2021-02-16
Refact/reuse libaktualizr finalization

mike-sul:
Since the libaktualizr maintainer accepted some patches into the upstream (e.g. getCurrentHash) we can make use of OstreeManager::finalizeInstall instead of copy-pasting it in the lite client.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -67,8 +67,6 @@ LiteClient::LiteClient(Config& config_in)
   headers.push_back(header);
   add_apps_header(headers, config.pacman);
 
-  headers.emplace_back("x-ats-target: unknown");
-
   if (!config.telemetry.report_network) {
     // Provide the random primary ECU serial so our backend will have some
     // idea of the number of unique devices using the system
@@ -81,21 +79,10 @@ LiteClient::LiteClient(Config& config_in)
   uptane_fetcher_ = std::make_shared<Uptane::Fetcher>(config, http_client);
   report_queue = std_::make_unique<ReportQueue>(config, http_client, storage);
 
-  // finalizeIfNeeded it looks like copy-paste of SotaUptaneClient::finalizeAfterReboot
-  // can we use just SotaUptaneClient::finalizeAfterReboot or even maybe SotaUptaneClient::initialize ???
-  // in this case we could do our specific finalization, including starting apps, in ComposeAppManager::finalizeInstall
-  Uptane::Target current_target{Uptane::Target::Unknown()};
-  Uptane::Target target_been_applied{Uptane::Target::Unknown()};
-  data::InstallationResult target_installation_result;
-  std::tie(current_target, target_been_applied, target_installation_result) =
-      finalizeIfNeeded(*ostree_sysroot, *storage, config);
-  update_request_headers(http_client, current_target, config.pacman);
-
   key_manager_ = std_::make_unique<KeyManager>(storage, config.keymanagerConfig());
   key_manager_->loadKeys();
   key_manager_->copyCertsToCurl(*http_client);
 
-  // TODO: consider improving this factory method
   if (config.pacman.type == ComposeAppManager::Name) {
     package_manager_ =
         std::make_shared<ComposeAppManager>(config.pacman, config.bootloader, storage, http_client, ostree_sysroot);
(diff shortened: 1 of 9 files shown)


src/liteclient.cc
```diff
@@ -105,10 +94,28 @@ LiteClient::LiteClient(Config& config_in)
...
+    if (!!pending_target) {
+      data::InstallationResult update_finalization_result = package_manager_->finalizeInstall(*pending_target);
+      if (update_finalization_result.isSuccess()) {
+        LOG_INFO << "Marking target install complete for: " << *pending_target;
+        storage->saveInstalledVersion("", *pending_target, InstalledVersionUpdateMode::kCurrent);
+      }
+
+      if (update_finalization_result.result_code != data::ResultCode::Numeric::kAlreadyProcessed ||
+          update_finalization_result.result_code != data::ResultCode::Numeric::kNeedCompletion) {
+        notifyInstallFinished(*pending_target, update_finalization_result);
```
doanac, line 111: this will hit our backend without any `x-ats-target: unknown`?

src/liteclient.cc
```diff
@@ -105,10 +94,28 @@ LiteClient::LiteClient(Config& config_in)
...
+    // finalize a pending update if any
+    boost::optional<Uptane::Target> pending_target;
+    storage->loadInstalledVersions("", nullptr, &pending_target);
+
+    if (!!pending_target) {
+      data::InstallationResult update_finalization_result = package_manager_->finalizeInstall(*pending_target);
+      if (update_finalization_result.isSuccess()) {
+        LOG_INFO << "Marking target install complete for: " << *pending_target;
+        storage->saveInstalledVersion("", *pending_target, InstalledVersionUpdateMode::kCurrent);
+      }
```
doanac, line 107: it looks like there used to be an "else" hanging of `!!pending_target`
~~~
} else {
        if (bootloader.rebootDetected()) {
          std::string err = "Expected to boot on " + target.sha256Hash() + " buf found " + current_hash +
                            ", system might have experienced a rollback";
          LOG_ERROR << err;
          storage.saveInstalledVersion("", target, InstalledVersionUpdateMode::kNone);
          bootloader.rebootFlagClear();
          ir.result_code = data::ResultCode::Numeric::kInstallFailed;
          ir.description = err;
        } else {
          // Update still pending as no reboot was detected
          ir.result_code = data::ResultCode::Numeric::kNeedCompletion;
        }   
        // Installation was not successful
        return std::tie(*current_version, target, ir);
      }   
~~~

I think that's kind of `package_manager_->finalizeInstall(*pending_target);` but slightly different? Not sure if its functionally different.

doanac (commented):
This one scares me a little. Are we absolutely sure the risk/reward ratio of this change is worth it?

src/liteclient.cc
mike-sul: Not sure if I fully understand the comment, but just in case there is the setting of x-ats-target to "unknown" at line #70.
Then this header is updated whenever an installation was successful, either in `daemon_main` or in LiteClient ctor after a reboot and successful finalization.

src/liteclient.cc
mike-sul: conceptually the same, but OstreeManager::finalizeInstall does it correctly and is well tested, the aklite actually used it originally when it used sotauptaneclient.
My version LiteClient::finalizeIfNeeded() has a tiny issue, it actually doesn't verify if a reboot has happened if there is a pending version, just simply checks the current hash which can be applied manually by doing `ostree admin deploy` without reboot. Also, an installation finalization should be done by an entity that actually performs installation, i.e. package manager in our case.

mike-sul:
> Are we absolutely sure the risk/reward ratio of this change is worth it?

I think so, in the mid and long term the change is worth it, the potential risk of regression can be mitigated by adding more tests, which we need to add anyway. 
With any change, there is a risk of regression or the introduction of new issues. The only way to avoid risk for 100%/sure is not to make any changes at all. But it's not possible, so it's all about automated verification/testing of any change. Not cleaning code and reducing tech debts accumulates issues, it just postpones risks moreover it increases potential impact, and it will be harder to fix something if we get more customers/devices.

src/liteclient.cc
doanac: i guess my question should be: does sending an installFinished event with an x-ats-target=unknown header cause any unintended problems on our backend. Is the device.status correctly set, etc?

src/liteclient.cc
doanac: ack. i'd like to have @ricardosalveti take a look at this snippet since he understands the rollback stuff more than i do.

ricardosalveti:
s/make us/make use/g on the commit message.

src/liteclient.cc
ricardosalveti: There is one difference that I could notice is that on the new path we don't call storage.saveInstalledVersion("", target, InstalledVersionUpdateMode::kNone); anymore, and not sure if that can potentially cause an issue on the rollback logic.

src/liteclient.cc
ricardosalveti: Right, I remember that the reason to call update_reguest_headers before was to make sure x-ats-target had at least the current installed version string.

ricardosalveti:
I think this is a good change, as it reduces the special logic we had that wasn't covered by the package management part.

We just need to make sure we test and validate the rollback logic in full (and extend the current unit tests if needed).

src/liteclient.cc
mike-sul: Yeah, I got it, I just thought about something else here...
It looks like `headers.emplace_back("x-ats-target: unknown");` at line #70 is redundant/useless.

src/liteclient.cc
mike-sul: It's kind of redundant because this call doesn't change the existing record for the given target. What important for the rollback functionality is the flag `was_installed`. It should be set to 1/true only for those Targets that were successfully installed and we do it either in LiteClient::install() for updates that do not require a reboot and in LiteClient::LiteClient() if a reboot is required, reboot happened and the finalization was successful. So, if Target finalization was not successful we just simply don't do anything with the Target record in DB because its `was_installed` is set to 0 anyway.

mike-sul:
>We just need to make sure we test and validate the rollback logic in full (and extend the current unit tests if needed).

We do have unit tests for the rollback-related functionality. Unfortunately, we don't have proper tests to test the finalization and rollback because it requires not unit-tests but rather functional or even integration tests, i.e. use LiteClient instance, which in turn requires TUF repo and Treehub mocks development of which can take some time. I actually started working on it, so maybe I pull some of the components into the given PR and add test(s) specifically for finalization&rollback.

mike-sul:
@ricardosalveti @doanac I've updated (forced push) PR according to your comments. 
I think there is no rush in merging the given PR into master so I am going to pull the test components/mocks I've been working on into this PR and build some finaliztion&rollback related test(s).

mike-sul:
@ricardosalveti @doanac I added the tests that I planned to add, so I think it's safe to merge it now.

doanac (commented):
I appreciate the work you did to create a fully mocked test to cover this.

mike-sul:
> I appreciate the work you did to create a fully mocked test to cover this.

So, can we merge it?

ricardosalveti (approved):
Looks fine from my perspective, happy to land and have it available in lmp for further functional testing.

PR #65, opened 2021-03-10
liteclient: let storage know its client (TUF/Uptane)

ldts:
Instead of providing the client information in the configuration class
(which describes its fundamental characteristics but not how these will
be used) we choose to provide it as a parameter to the instantiation
of the class.

Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>


ldts:
error is normal since it is not using the right version of the lib. need to incorporate it

mike-sul (approved):

PR #70, opened 2021-03-15
Test headers aklite sends to Device Gateway

mike-sul:
Make sure that aklite sends the requests headers to Device Gateway that it uses to infer/deduce a device status.

### DIFF
--- a/tests/liteclient_test.cc
+++ b/tests/liteclient_test.cc
@@ -57,11 +57,11 @@ class MockAppEngine : public AppEngine {
 
 class SysRootFS {
  public:
-  static std::string GeneratorPath;
+  static std::string CreateCmd;
  public:
   SysRootFS(std::string path_in, std::string branch_in, std::string hw_id_in, std::string os_in):
       path{std::move(path_in)}, branch{std::move(branch_in)}, hw_id{std::move(hw_id_in)}, os{std::move(os_in)} {
-    run_cmd(GeneratorPath, { path, branch, hw_id, os }, "generate a system rootfs template");
+    run_cmd(CreateCmd, { path, branch, hw_id, os }, "generate a system rootfs template");
   }
 
   const std::string path;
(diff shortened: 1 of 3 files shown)


tests/liteclient_test.cc
```diff
@@ -200,15 +182,45 @@ class TufRepoMock {
...
+                    ostree_mock_{ostree_mock},
+                    tuf_repo_mock_{tuf_repo_mock},
+                    port_{TestUtils::getFreePort()},
+                    url_{"http://localhost:" + port_},
+                    process_{RootPath, "--port", port_, "--ostree", ostree_mock_.path(), "--tuf-repo", tuf_repo_mock_.path()} {
+    TestUtils::waitForServer(url_ + "/");
+    LOG_INFO << "Device Gateway is running on port " << port_;
+  }
+  ~DeviceGatewayMock() {
+    process_.terminate();
```
ldts, line 189: maybe check process_.running() before calling terminate and wait?

mike-sul: the required checks are inside of these methods.

tests/liteclient_test.cc
```diff
@@ -340,14 +329,22 @@ class LiteClientTest : public ::testing::Test {
         // make sure that the new Target has been applied
         ASSERT_EQ(client.getCurrent().sha256Hash(), to_target.sha256Hash());
         ASSERT_EQ(client.getCurrent().filename(), to_target.filename());
+        // emulate the daemon_main work, this is all looks really ugly
+        // but without pulling the update simplifaction PR changes it's really tough to
+        // make it nice, clean and avoiding code duplication
+        // line 264 in main.cc
```
doanac, line 335: "line 264 in main.cc" isn't going to be valid for long. I'd either drop the comment or maybe add a "TODO code duplication ...." with a versioned reference to main.cc.

Otherwise - this is a nice test. We've had at least 2 regressions on these headers in the past.

mike-sul: Updated with "TODO: "

doanac (approved):

PR #73, opened 2021-03-22
apps: send app list in a header unconditionally

mike-sul:
- send a currently installed app list in a header request to the backend
unconditionally, regadles if an app shortlist specified or not;
- add tests to verify the given use-case

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -478,18 +478,16 @@ void LiteClient::update_request_headers(std::shared_ptr<HttpClient>& http_client
   if (config.type == ComposeAppManager::Name) {
     ComposeAppManager::Config cfg(config);
 
-    // If App list was not specified in the config then we need to update the request header with a list of
-    // Apps specified in the currently installed Target
-    if (!cfg.apps) {
-      std::list<std::string> apps;
-      auto target_apps = target.custom_data()["docker_compose_apps"];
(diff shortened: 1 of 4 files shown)


doanac (approved):
i think this is the least invasive way to improve this logic. So +1. 

I think in the future we should send the *running* apps. Then our backend could look at the device's toml, the target its on, and the apps that are running and know if the device is in the correct state.

doanac:
NOTE - i didn't read the tests too closely. you and @ldts seem to have a good handle on that now

munoz0raul:
@mike-sul this fix solve my problem of not display the apps in the UI.

Just tested this pull request on my device, flashed the new image, registered and the app witch was downloaded automatically is listed in the UI and in the fioctl:

```
fioctl devices show raspberrypi3-64
UUID:		f3e132f5-f179-4463-ac2a-41a04228d65e
Owner:		5e13232f73927550af883e7b
Factory:	tutorial
Up to date:	true
Target:		raspberrypi3-64-lmp-5 / sha256(bba54a445f16586f84b04a6a0dd83bbb96e37361ab6682ec309c24796aa03477)
Ostree Hash:	bba54a445f16586f84b04a6a0dd83bbb96e37361ab6682ec309c24796aa03477
Created:	2021-03-23T17:45:04+00:00
Last Seen:	2021-03-23T17:45:10+00:00
Tags:		devel
Docker Apps:	shellhttpd
```

PR #74, opened 2021-03-22
docker http client and yaml parser: integration with composeappsengine, unit tests

ldts:
Use an external tool to convert yaml to json.
Use curl to query docker


ldts:
um, the clang checks werent working on my build; let me fix that before going ahead with the review.

ldts:
now failing the tests because of the missing libfyaml dependency.

src/composeinfo.cc
```diff
@@ -0,0 +1,44 @@
...
+  image = json_.root_["services"][service.asString()]["image"].asString();
+  return !image.empty();
+}
+
+bool ComposeInfo::getServiceHash(const Json::Value& service, std::string& hash) {
+  hash = json_.root_["services"][service.asString()]["labels"]["io.compose-spec.config-hash"].asString();
+  return !hash.empty();
+}
+
+int ComposeInfo::getNbrImages() {
```
doanac, line 30: I suspect you did this because Mike's original code was counting "images". The reason he did this was that counting "services" in YAML by straight string parsing isn't really feasible. We have a requirement that every service must have an image, so he can basically grep for images and get the total number of *services*.

tldr; we need a `getNumServices` method which is close to what you have below but can become simpler by dropping the `!image.empty()` stuff

src/composeinfo.h
```diff
@@ -0,0 +1,24 @@
...
+#include <json/json.h>
+#include <string>
+#include <vector>
+#include "yaml2json.h"
+namespace Docker {
+
+class ComposeInfo {
+ public:
+  ComposeInfo(const std::string& yaml);
+  bool getServices(std::vector<Json::Value>& services);
```
doanac, line 13: Any reason to not convert to actual types? e.g
~~~
struct ComposeService {
  std::string name
  std::string svc_hash;
};
~~~

Then you could drop the 2 helper methods below.

src/yaml2json.cc
```diff
@@ -0,0 +1,21 @@
...
+#include <logging/logging.h>
+
+Yaml2Json::Yaml2Json(const std::string& yaml) {
+  g_autofree gchar* stdout_buff = nullptr;
+  g_autofree gchar* stderr_buff = nullptr;
+  g_autofree GError* error = nullptr;
+  gint status;
+
+  // clang-tidy insane line wrapping requirements follows
+  if (g_spawn_command_line_sync(("fy-tool --mode json " + yaml).c_str(), &stdout_buff, &stderr_buff, &status, &error) !=
```
doanac, line 12: libaktualizr has a helper to make this easier: https://github.com/foundriesio/aktualizr/blob/c31f3956277fb6c9dcdc4a7aa7a61d6f3dc07964/src/libaktualizr/utilities/utils.h#L42

src/composeappengine.cc
```diff
@@ -42,46 +43,42 @@ void ComposeAppEngine::remove(const App& app) {
...
 
-  int running_container_number =
-      std::count_if(cmd_output.begin(), cmd_output.end(), [](const char& symbol) { return symbol == '\n'; });
-
-  if (running_container_number < expected_container_number) {
-    LOG_DEBUG << "Number of running containers is less than a number of images specified in the compose file"
-              << "; App: " << app.name << "; expected container number: " << expected_container_number
-              << "; number of running containers: " << running_container_number;
-    return false;
+    for (std::size_t i = 0; i < services.size(); i++) {
```
doanac, line 55: I think you already mentioned this, but I see two phases of optimization.

1) change only this method with something like
~~~
docker ps -q --filter=status=running --filter=label=com.docker.compose.project=shellhttpd --format='{{.Label "com.docker.compose.service" }} {{.Label "io.compose-spec.config-hash"}}'
httpd df03aff6219ace2d86e9003ac8cd294f3438d271c1614820ebe62432b07920b8
~~~

2) Call docker-ps one level up in the stack *once* with:
~~~
docker ps -q --filter=status=running --format='{{.Label "com.docker.compose.project"}} {{.Label "com.docker.compose.service" }} {{.Label "io.compose-spec.config-hash"}}'
fiotest fiotest 2fc5236772a8cc436c105493c5b8739d969e1aca3d90492382ea74719e6f18b3
shellhttpd httpd df03aff6219ace2d86e9003ac8cd294f3438d271c1614820ebe62432b07920b8
~~~

Point 2 will probably be a noticeable performance improvement.

doanac (commented):
I've been waiting for this for so long.

src/composeappengine.cc
mike-sul: @doanac We've already discussed it with @ldts. The ultimate goal is what you outlined here https://github.com/foundriesio/aktualizr-lite/pull/46. In a nutshell, a new component `namespace Docker::Engine/Daemon` that exposes API to get required details of running containers as well as installed images (if needed) and using libcurl to make requests to the docker daemon. Then, ComposeAppEngine is initialized with an instance of Docker::Daemon or its factory method and makes request(s) to it just once before looping over apps in isRunning() method.

src/composeinfo.cc
mike-sul: +1, there is no need in neither getNbrImages() nor in getNumServices() since there is `getServices(std::vector<Json::Value>& services)` and services.size().

src/composeinfo.h
mike-sul: We agreed to work on it iteratively.

src/composeinfo.cc
ldts: yes I agree, it is pointless (and ugly). will remove

src/yaml2json.cc
ldts: yes; just changed.
also we should probably be giving full paths to the binaries (so what gets executed is not affected by chages in $PATH)

src/composeappengine.cc
ldts: right, if we are going to get rid of docker ps then it seemed simpler to me just parse this way and do a final parser in some other way when we are down to that... but I can optimize it here too

ldts: anyway, fixed it too

doanac: I get the impression, this might not be time for me to review this and i should wait a bit?

I assume this PR isn't going to move us to talking directly with the API though? i.e. - so not running `docker ps` in  loop is still worth doing since this will get into master?

mike-sul: >I assume this PR isn't going to move us to talking directly with the API though? 
No, it's not.

But Jorge does one `docker ps` per app now, so it's not worse than it was before.

mike-sul: @doanac I think it's ready for review.

mike-sul:
@ldts Please, update this PR with changes to https://github.com/foundriesio/aktualizr-lite/blob/master/how-to-run-locally.md so it includes details on the installation of the new dependency (yaml2json).

src/composeappengine.cc
```diff
@@ -42,46 +43,64 @@ void ComposeAppEngine::remove(const App& app) {
...
-  if (running_container_number < expected_container_number) {
-    LOG_DEBUG << "Number of running containers is less than a number of images specified in the compose file"
-              << "; App: " << app.name << "; expected container number: " << expected_container_number
-              << "; number of running containers: " << running_container_number;
-    return false;
-  }
+    std::tie(ret, output) = cmd(docker_ + req);
+    if (!ret || output.empty()) {
+      LOG_INFO << req;
+      exit(0);
```
ldts, line 67: argh, this is a left over...need to fix!

how-to-run-locally.md
```diff
@@ -74,6 +74,23 @@ ostree config --repo="${OSTREE_SYSROOT}/ostree/repo" set core.mode bare-user-onl
...
+Aktualizr-lite uses libfyaml to consume docker.yml files in json format.
+In order to resolve that binary dependency we need to install it in the same host where aktualizr-lite will be executing.
+The LmP targets install v0.6 so to replicate that install:
+
+```
+cd /tmp/
+wget https://github.com/pantoniou/libfyaml/archive/refs/tags/v0.6.zip && unzip v0.6.zip
+cd libfyaml-0.6
+echo "0.6.0" > .tarball-version
+./boostrap.sh
```
mike-sul, line 87: misspelling `boostrap.sh` -> `bootstrap.sh`

src/composeappengine.cc
```diff
@@ -42,46 +43,67 @@ void ComposeAppEngine::remove(const App& app) {
...
 
-  // Get a number of running container images
-  std::tie(cmd_res, cmd_output) =
-      cmd(docker_ + "ps -q --filter=status=running --filter=label=com.docker.compose.project=" + app.name);
-  if (!cmd_res) {
-    LOG_WARNING << "Failed to get a list of App's containers: " << app.name;
-    return false;
-  }
+    // create a json array from the command output
+    std::string json("[" + output + "]");
```
mike-sul, line 74: I wonder if these square brackets are really needed?

src/composeappengine.cc
```diff
@@ -42,46 +43,67 @@ void ComposeAppEngine::remove(const App& app) {
...
+    std::istringstream sin(json);
 
-  int running_container_number =
-      std::count_if(cmd_output.begin(), cmd_output.end(), [](const char& symbol) { return symbol == '\n'; });
+    Json::Value running;
+    sin >> running;
```
mike-sul, line 78: Utils::parseJSON() can be used, instead.

ldts: what is the advantage in that?

src/composeappengine.cc
ldts: yes, otherwise you will generate an exception

ldts:
just an early preview of rest.
fixing clang-tidy ATM.
we can review when done.

src/docker/containerclient.h
```diff
@@ -0,0 +1,18 @@
...
+#define CONTAINER_CLIENT_H
+
+#include <json/json.h>
+#include <string>
+#include <vector>
+#include "dockerclient.h"
+#include "restclient.h"
+using namespace std;
+
+class ContainerClient : public DockerClient {
```
mike-sul, line 11: namespace Docker?

src/docker/containerclient.h
```diff
@@ -0,0 +1,18 @@
...
+ public:
+  ContainerClient() : DockerClient(){};
```
mike-sul, line 13: ContainerClient seems like a redundant class.

src/docker/dockerclient.h
```diff
@@ -0,0 +1,24 @@
...
+#include <string>
+#include "restclient.h"
+using namespace std;
+
+class DockerClient {
+ public:
+  DockerClient() : uri_("http://127.0.0.1") { RestClient::init(); }
+  DockerClient(string host) : uri_(host) { RestClient::init(); }
+  virtual ~DockerClient() { RestClient::disable(); }
+  string param(string param_name, string param_value);
```
mike-sul, line 15: `param` methods don't look like the docker client class methods rather some URI class methods if it's needed at all.

src/composeappengine.cc
```diff
@@ -42,46 +45,44 @@ void ComposeAppEngine::remove(const App& app) {
...
-  while ((find_pos = cmd_output.find("image:", line_pos)) != std::string::npos) {
-    if (cmd_output.substr(line_pos, (find_pos - line_pos)).find_first_of('#') == std::string::npos) {
-      ++expected_container_number;
+  try {
+    ComposeInfo compose((appRoot(app) / ComposeFile).string());
+
+    // get the running information from docker
+    Json::Value running;
+    ContainerClient client;
+    if (!client.getContainer(false, app.name, running)) {
```
mike-sul, line 54: With such an approach, we still do one call to the docker daemon per app...

src/docker/include/callbacks.h
```diff
@@ -0,0 +1,43 @@
+#ifndef CALLBACKS_H
+#define CALLBACKS_H
+
+#include <algorithm>
+#include <functional>
+#include <string>
+#include "version.h"
+
+namespace RestClient {
```
doanac, line 9: I think you greatly simplify this by using libaktualizr's HttpClient class that has all this stuff. e.g.:

 https://github.com/foundriesio/aktualizr-lite/pull/46/files#diff-d2f08e43f860c947307b0b54cbbdfe1bbea2d669ec67ee3635b5bc82569e96cd

doanac: Also - one clever thing I've started copying from @mike-sul is the use of factories for things like this so that you can create nice mocks in the test class.

src/composeappengine.cc
ldts: did I misunderstand? do we not want that?
I mean it is easy to fix it - so grab all services at once and then parse.......let me know if you'd rather have that!

src/docker/include/callbacks.h
ldts: @doanac what is all that? all that we need is this....

`HttpClient client("/var/run/docker.sock");
 std::string cmd = "http://localhost/containers/json?" + app;
 root_ =  client.get(cmd, HttpInterface::kNoLimit).getJson();
`

src/composeappengine.cc
ldts: @doanac removed the docker ps stuff....at some point we should review and merge before I continue adding stuff...

src/docker/include/callbacks.h
ldts: btw if the requirement we need is to have an occasional http://localhost/containers/json?all request (instead of per-app), let me know and I will amend...is really straightforward

ldts:
btw I opted for adding support for unix sockets to aktuzlirs http instead of deriving from that class....seemed more logical to me.

src/composeappengine.cc
mike-sul: In fact, after looking into the implementation https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_reader.cpp#L1983 I think it's better compared to https://github.com/foundriesio/aktualizr/blob/2021.1%2Bfio/src/libaktualizr/utilities/utils.cc#L244. :). The libaktualizr's version never checks for parsing error...

src/dockerclient.h
```diff
@@ -0,0 +1,20 @@
+#ifndef AKTUALIZR_LITE_DOCKERCLIENT_H
+#define AKTUALIZR_LITE_DOCKERCLIENT_H
+
+#include <json/json.h>
+#include <string>
+
+namespace Docker {
+
+class DockerClient {
```
mike-sul, line 9: Not critical, since `DockerClient` is already in the scope of `Docker` namespace then `Docker` prefix becomes redundant in the class name.

src/dockerclient.cc
```diff
@@ -0,0 +1,42 @@
+#include "dockerclient.h"
+#include <logging/logging.h>
+#include "http/httpclient.h"
+#include "utilities/utils.h"
+
+namespace Docker {
+
+DockerClient::DockerClient(const std::string& app, bool curl) {
+  if (curl) {
+    std::string cmd = "/usr/bin/curl --unix-socket /var/run/docker.sock http://localhost/containers/json?" + app;
```
mike-sul, line 10: I guess it can be removed now?

mike-sul:
@ldts It looks nice now. The next phase is testing, i.e. tests -> LiteClient()->ComposeAppManager->ComposeAppEngine->FakeComoseInfo, FakeDockerClient. To be able to do it we need DI (dependency injection), instead of this `ComposeAppEngine(boost::filesystem::path root_dir, std::string compose_bin, std::string docker_bin,
                   Docker::RegistryClient::Ptr registry_client);` https://github.com/foundriesio/aktualizr-lite/blob/dcfc946103efa3c78833e5c801415f544c3b674b/src/composeappengine.h#L20
                
pass Docker::Client::Ptr and Docker::Compose::Ptr or their factory methods to ComposeAppEngine ctor.

src/dockerclient.cc
ldts: why not leave it as an option? it also serves as documentation IMO

src/composeappengine.cc
doanac: yep - i think its time to make this good enough and merge as well.

PR #76, opened 2021-03-29
apps: don't start updated apps if ostree rollbacked

mike-sul:
- Don't start any updated apps if the rootfs/ostree update was not
successful and rollback happened during a boot;
- Add a test to cover the given use-case.

### DIFF
--- a/src/compose/composeappmanager.cc
+++ b/src/compose/composeappmanager.cc
@@ -68,17 +68,6 @@ ComposeAppManager::ComposeAppManager(const PackageConfig& pconfig, const Bootloa
   } catch (const std::exception& exc) {
     LOG_DEBUG << "Failed to initialize Compose App Tree (ostree) at " << cfg_.apps_tree << ". Error: " << exc.what();
   }
-
-  const auto& current_apps = getApps(getCurrent());
-  // TODO: refactor it to fix "app rollback when ostree rollback" issue
-  for (const auto& app_pair : current_apps) {
-    const auto& app_name = app_pair.first;
-    auto need_start_flag = cfg_.apps_root / app_name / Docker::ComposeAppEngine::NeedStartFile;
-    if (boost::filesystem::exists(need_start_flag)) {
-      app_engine_->run({app_pair.first, app_pair.second});
-      boost::filesystem::remove(need_start_flag);
-    }
-  }
 }
 
 // Returns an intersection of apps specified in Target and the configuration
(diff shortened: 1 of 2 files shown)


mike-sul:
@doanac I think after all refactoring we've done this "need_start" stuff and special/dedicated start of the updated apps out of the main update cycle becomes redundant. So, we can just simply remove that code that generates `need_start` flag and the code that iterates over apps and runs them in the finalization method. The update cycle/daemon_main will bring the correct version of Apps in both cases, successful finalization (ostree and apps udpate) and in the case of rollback (sync). I'll double-check it and update PR accordingly.

doanac:
@mike-sul - so the apps still get started, its just not going to start the newer apps that we are rolling back from, right?

mike-sul:
> @mike-sul - so the apps still get started, its just not going to start the newer apps that we are rolling back from, right?

correct.

doanac (approved):

mike-sul:
@doanac actually we don't need to do anything in terms of starting apps in the finalization at all, please, see the latest forced update.

doanac:
I like a PR that's deleting confusing code. However, I'm a little confused about the chain of events when we have a GOOD target update (with ostree hash change). I *think* what happens is:

 * device reboots
 * `daemon_main`
 * `!client.appsInSync()` is hit since the `run` method on the app will not have been called
 * This will trigger a `do_app_sync` call which does a `client.install` to "start" the apps.

If my understanding is correct, there's a problem. The `client.install` will report update events to the server making it look like another install was performed (included *downloading* the compose bundle which should be in place) when we are really just calling docker-compose-up?

mike-sul:
> If my understanding is correct, there's a problem.

That's correct. Why it's a problem? The backend will get all headers set in the correct state/value eventually.

doanac:
> Why it's a problem? The backend will get all headers set in the correct state/value eventually.

The biggest thing would be because its hard to audit a device's update events. You now have 2x the "updates" to look at. It also seems like if a `->run` fails, in the old code we'd mark the original install as FAILED. whereas now, we'll mark that install as PASSED, and then have a 2nd install that gets marked FAILED? In this case the first one wasn't a PASS it was really just incomplete.

mike-sul:
> > Why it's a problem? The backend will get all headers set in the correct state/value eventually.
> 
> The biggest thing would be because its hard to audit a device's update events. You now have 2x the "updates" to look at. It also seems like if a `->run` fails, in the old code we'd mark the original install as FAILED. whereas now, we'll mark that install as PASSED, and then have a 2nd install that gets marked FAILED? In this case, the first one wasn't a PASS it was really just incomplete.

I thought we agreed that we shouldn't rely on the report events for deducing a device state on the backend, instead trust in the information aklite sends to the backend at the beginning of each update cycle, which are, currently, request headers. The report events should be treated as a kind of log message.
In addition to that, I think we shouldn't start any download at all if required apps installed and I think it's so in the "cleaned/simplified" version (we have isInstalled() method in addition to isRunning).

Having said that I will roll back to the previous version since we don't the "simplfied" version in place yet as well as the backend might get crazy.

doanac:
the backend does need to sort things out, but having the client tell us the install is complete before its actually completed doesn't make that easier. If we could somehow keep the correlation-id and re-use that in app-sync logic we could at least make the update start to appear as a single update. But I'm guessing that would be pretty messy?

mike-sul:
> the backend does need to sort things out, but having the client tell us the install is complete before its actually completed doesn't make that easier. If we could somehow keep the correlation-id and re-use that in app-sync logic we could at least make the update start to appear as a single update. But I'm guessing that would be pretty messy?

I disagree, once more, events can be lost, so the backend SHOULD not rely on them and treat them as just hints or logs.
Any decision about a device state should be based on information that is delivered to the backend in a reliable way. We talked about sending of device manifest to the backend...

Anyway, let's discuss it offline, as I said I will roll back it to the previous version.

mike-sul:
@doanac Rollbacked the "rolback" fix via forced push.

PR #78, opened 2021-03-30
Encapsulate Target-related functionality

mike-sul:
This is the first "cherry-pick" from the "simplification" branch https://github.com/foundriesio/aktualizr-lite/tree/refact/update-logic-simplification. It defines a new class/namespace aimed to encapsulate all Target-related functionality and decouple its clients from implementation details as well as avoid hard coding and copy-pasting. It's not fully leveraged in the given PR as its full utilization require further changes/refactoring in the code base, but will be so later.

### DIFF
--- a/src/compose/composeappmanager.cc
+++ b/src/compose/composeappmanager.cc
@@ -1,4 +1,5 @@
 #include "composeappmanager.h"
+#include "target.h"
 
 ComposeAppManager::Config::Config(const PackageConfig& pconfig) {
   const std::map<std::string, std::string> raw = pconfig.extra;
@@ -105,7 +106,7 @@ ComposeAppManager::AppsContainer ComposeAppManager::getApps(const Uptane::Target
 ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane::Target& t) const {
   AppsContainer apps_to_update;
 
-  auto currently_installed_target_apps = OstreeManager::getCurrent().custom_data()["docker_compose_apps"];
+  auto currently_installed_target_apps = Target::appsJson(t);
   auto new_target_apps = getApps(t);  // intersection of apps specified in Target and the configuration
 
   for (const auto& app_pair : new_target_apps) {
@@ -171,7 +172,7 @@ bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetche
   LOG_INFO << "Found " << cur_apps_to_fetch_and_update_.size() << " Apps to update";
 
   bool passed = true;
-  const auto& apps_uri = target.custom_data()["compose-apps-uri"].asString();
+  const auto& apps_uri = Target::ostreeURI(target);
   if (app_tree_ && !apps_uri.empty()) {
     LOG_INFO << "Fetching Apps Tree -> " << apps_uri;
 
@@ -214,7 +215,7 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
 
   handleRemovedApps(target);
 
-  const auto& apps_uri = target.custom_data()["compose-apps-uri"].asString();
+  const auto& apps_uri = Target::ostreeURI(target);
   if (app_tree_ && !apps_uri.empty()) {
     LOG_INFO << "Checking out updated Apps: " << apps_uri;
     try {
@@ -305,7 +306,6 @@ void ComposeAppManager::handleRemovedApps(const Uptane::Target& target) const {
     LOG_DEBUG << "cfg_.apps_root does not exist";
     return;
   }
-  std::vector<std::string> target_apps = target.custom_data()["docker_compose_apps"].getMemberNames();
 
   // an intersection of apps specified in Target and the configuration
   // i.e. the apps that are supposed to be installed and running
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -9,6 +9,7 @@
 
 #include "composeappmanager.h"
 #include "crypto/keymanager.h"
+#include "target.h"
 
 LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, bool finalize)
     : config{std::move(config_in)}, primary_ecu{Uptane::EcuSerial::Unknown(), ""} {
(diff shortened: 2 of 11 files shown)


src/target.cc
```diff
@@ -0,0 +1,84 @@
...
+        (*app_shortlist).end() != std::find((*app_shortlist).begin(), (*app_shortlist).end(), app.name)) {
+      apps.push_back(app.name);
+    }
+  }
+  return boost::algorithm::join(apps, ",");
+}
+
+void Target::setAppsJson(Uptane::Target& target, const Json::Value& apps_json) {
+  Json::Value custom_data = target.custom_data();
+  if (!custom_data.isNull()) {
```
doanac, line 58: seems odd. Should this method fail if its not able to set this property or throw an exception? It seems an exceptional situation to have a target with no custom data defined?

Also - this only seems to be called from `createTarget`. I don't see that referenced anywhere. Is this dead code for now we don't yet need?

src/liteclient.cc
```diff
@@ -489,20 +490,13 @@ void LiteClient::add_apps_header(std::vector<std::string>& headers, PackageConfi
...
-    for (Json::ValueIterator ii = target_apps.begin(); ii != target_apps.end(); ++ii) {
-      if ((*ii).isObject() && (*ii).isMember("uri")) {
-        const auto& target_app_name = ii.key().asString();
-        if (!cfg.apps || (*cfg.apps).end() != std::find((*cfg.apps).begin(), (*cfg.apps).end(), target_app_name)) {
-          apps.push_back(target_app_name);
-        }
-      }
-    }
-    http_client->updateHeader("x-ats-dockerapps", boost::algorithm::join(apps, ","));
+    http_client->updateHeader("x-ats-dockerapps", Target::appsStr(target, ComposeAppManager::Config(config).apps));
```
doanac, line 494: Not sure if the `appsStr` logic would ever be useful anywhere else? If so, it might be better to return as `std::vector<std::string>` and then do the boost::join here.

doanac (commented):
Looks pretty good. Nothing major

src/liteclient.cc
mike-sul: It's used by the tests (liteclient_test) to emulate daemon_main, the final "simplified" version won't needed so I'll update it accordingly then.

src/target.cc
mike-sul: It's needed in the future :). The final "simplified" version makes use of it. I can remove it from here and put it back again, although it's kind of a waste of time...

tests/liteclient_test.cc
```diff
@@ -385,9 +385,11 @@ class LiteClientTest : public ::testing::Test {
   /**
    * method createApp
    */
-  AppEngine::App createApp(const std::string& name, const std::string& factory = "test-factory") {
-    const std::string uri = "localhost:" + getDeviceGateway().getPort() + "/" + factory + "/" + name +
-                            "@sha256:7ca42b1567ca068dfd6a5392432a5a36700a4aa3e321922e91d974f832a2f243";
+  AppEngine::App createApp(const std::string& name,
```
ldts, line 389: I think we should modify the params since once we switch to using the http client registry we will only need the factory and the name...so maybe having the order as name, factory and hash so we can keep a call with two defaults

src/target.h
```diff
@@ -0,0 +1,97 @@
...
+
+#include "uptane/tuf.h"
+
+class Target {
+ public:
+  static constexpr const char* const TagField{"tags"};
+  static constexpr const char* const ComposeAppField{"docker_compose_apps"};
+  static constexpr const char* const ComposeAppOstreeUri{"compose-apps-uri"};
+
+  struct Version {
```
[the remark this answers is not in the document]
mike-sul: That's a good point, but not just remove from helpers, but also update clients so they use this version.

src/liteclient.h
```diff
@@ -26,7 +26,7 @@ class Lock {
 
 class LiteClient {
  public:
-  LiteClient(Config& config_in, const std::shared_ptr<AppEngine>& app_engine = nullptr);
+  LiteClient(Config& config_in, const std::shared_ptr<AppEngine>& app_engine = nullptr, bool finalize = true);
```
ldts, line 29: maybe set it to false by default? seems less dangerous

ldts: just wondering, instead of a flag, why dont we add it to the config class?

src/target.cc
doanac: that's fine. how about the null check thought?

mike-sul: yep, that's rather an exceptional situation, will add throwing an exception here.

src/liteclient.h
mike-sul: It was always "true" before the flag was introduced, hence I decided to leave it as a default.

It will require making changes to libaktualizr while we would like to reduce dependency on it.

ldts: and if instead of a flag we create an Actions class? not sure if in the future we will need to perform some other actions?

ldts: also as a general rule, would it be possible if we use smaller names?
ie finalizeInstallation --> completeInstall or  doInstall? 
I find it much easier to read and remember (but this is just a personal perference)

src/compose/composeappmanager.cc
```diff
@@ -287,8 +276,20 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
 
 data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target& target) {
   auto ir = OstreeManager::finalizeInstall(target);
-  if (ir.result_code != data::ResultCode::Numeric::kAlreadyProcessed ||
-      ir.result_code != data::ResultCode::Numeric::kNeedCompletion) {
+
+  const auto& current_apps = getApps(target);
+  for (const auto& app_pair : current_apps) {
+    const auto& app_name = app_pair.first;
+    auto need_start_flag = cfg_.apps_root / app_name / Docker::ComposeAppEngine::NeedStartFile;
```
ldts, line 284: can we choose a smaller name? start?

src/compose/composeappmanager.cc
```diff
@@ -287,8 +276,20 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
 
 data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target& target) {
   auto ir = OstreeManager::finalizeInstall(target);
-  if (ir.result_code != data::ResultCode::Numeric::kAlreadyProcessed ||
-      ir.result_code != data::ResultCode::Numeric::kNeedCompletion) {
+
+  const auto& current_apps = getApps(target);
```
ldts, line 281: current_apps --> apps?

src/compose/composeappmanager.cc
```diff
@@ -287,8 +276,20 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
...
+  for (const auto& app_pair : current_apps) {
+    const auto& app_name = app_pair.first;
```
ldts, line 283: app_name --> name?

src/compose/composeappmanager.cc
doanac: "current" is pretty important context in many cases.

src/target.h
```diff
@@ -0,0 +1,97 @@
...
+    Uptane::Target createTarget(const Uptane::Target& target) {
+      auto result{target};
+      Target::setAppsJson(result, target_apps_json_);
+      return result;
+    }
+
+    bool empty() const { return target_apps_json_.empty(); }
+
+   private:
+    Json::Value target_apps_json_;
```
ldts, line 93: why do we write the type of the data member? I dont think it is necessary and just adds noise when reading the code....

src/target.h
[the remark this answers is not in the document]
mike-sul: done

src/target.h
mike-sul: Not sure if I understand the question.

src/target.cc
mike-sul: I actually removed the currently unused code since it's not possible to verify it. Will added it once it becomes needed.

src/liteclient.h
mike-sul: this is the name of the package manager interface, so it makes sense to name in the way the underlying code uses, so we have consistent terminology.

mike-sul: >if instead of a flag we create an Actions class?

it looks like overengineering in the given context. I would rather consider refactoring an overall main.cc with something as you are proposing, but I would do it only after we have a decent CI/CD  in place to make sure we don't introduce any regressions.

mike-sul:
@doanac I think it's good to go

doanac (approved):

PR #79, opened 2021-04-05
test: add mTLS support to Fake Device Gateway

mike-sul:
Add mTLS support to Fake Device Gateway so our tests/LiteClient can talk to it via HTTS/mTLS and enables the HSM path.

The following is an example, how to run/turn on it

`
./tests/device-gateway_fake.py --port 8443 -o build-ninja/ -o build-ninja/ -t build-ninja/ -j build-ninja/headers.json -s ./build-ninja/certs/server
`

How to test it

`
curl -I https://localhost:8443/ --cacert ca/ca.crt --key client/client.key --cert client/client.crt
`

The following is an example of how to generate PKI/key/certs https://github.com/mike-sul/prompt-http-server/blob/master/certs/simple-kpi.md.

This can be useful as context info https://github.com/mike-sul/docs/blob/master/TLS/tls.md

### DIFF
--- a/tests/device-gateway_fake.py
+++ b/tests/device-gateway_fake.py
@@ -5,6 +5,7 @@ import sys
 import argparse
 import json
 import logging
+import ssl
 
 from http.server import SimpleHTTPRequestHandler, HTTPServer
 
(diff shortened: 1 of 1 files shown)


doanac:
@mike-sul - so is the idea here that you'll have follow-on work that will generate certs during test(s)?

mike-sul:
> @mike-sul - so is the idea here that you'll have follow-on work that will generate certs during test(s)?

yep, @ldts is working on it. We would like our current tests run for LiteClient configured with enabled HSM, SoftHSM for the first milestone, and maybe later libsks/optee in qemu.

doanac:
I'll give it the old LGTM and and defer to @ldts for this then.

ldts:
@doanac yes, I am looking into importing libp11 and using library calls to handle that (instead of forking to pkcs11 tool). the idea is for the tests to be as close to the target releases as possible

mike-sul:
@ldts Feel free to merge it now or whenever you need it, it doesn't cause any regression nor code conflicts so can be merged at any moment.

ldts:
sure, I'll merge it as soon as I integrate it with the rest of the code that will use it so the PR is a complete set.

PR #80, opened 2021-04-05
tests: compose app engine tests and test fixtures

mike-sul:


src/docker/composeinfo.h
```diff
@@ -1,10 +1,12 @@
-#ifndef AKTUALIZR_LITE_COMPOSEINFO_H
-#define AKTUALIZR_LITE_COMPOSEINFO_H
+#ifndef AKTUALIZR_LITE_COMPOSE_INFO_H
+#define AKTUALIZR_LITE_COMPOSE_INFO_H
```
doanac, line 2: 20 - 30 years from now we'll be able to `#pragma once` this type stuff :)

src/docker/composeappengine.cc
```diff
@@ -51,11 +51,11 @@ bool ComposeAppEngine::isRunning(const App& app) const {
       LOG_WARNING << "App: " << app.name << ", no services in docker file!";
       return false;
     }
-    DockerClient client(app.name);
+
```
doanac, line 54: now that the docker-client doesn't get instantiated each call, its `root_` will be static. This looks to be a nice performance boost. however - don't we need some kind of `docker_client_->refresh()` type method so it picks up changes in the system?

src/docker/dockerclient.cc
```diff
@@ -10,7 +10,24 @@ DockerClient::HttpClientFactory DockerClient::DefaultHttpClientFactory = []() {
   return std::make_shared<HttpClient>("/var/run/docker.sock");
 };
 
-DockerClient::DockerClient(bool curl, const HttpClientFactory& http_client_factory) {
+DockerClient::DockerClient(std::shared_ptr<HttpInterface> http_client) : http_client_{std::move(http_client)} {}
+
+bool DockerClient::serviceRunning(const std::string& app, const std::string& service, const std::string& hash) {
+  updateContainerStatus();
```
doanac, line 16: ah - you've address my previous comment. You should probably squash these two commits .

doanac (commented):
i need to re-read the last commit and all its testing but this seems to move things in a good direction.

src/docker/dockerclient.cc
mike-sul: yep, I think I will update the previous commit with this change and remove it from the commit related to testing so both commits are self-sufficient and don't mix up topics.

mike-sul: In addition, to this change, in the future, I am going to add another method to AppEngine interface (e.g. refreshStat()) so the docker daemon is inquired once per an update cycle not per each App as it is now.

mike-sul:
@doanac I restructured commits a little bit to make them self-sufficient and not to mix different types of changes in a single commit.

mike-sul:
@doanac please, review this PR.

tests/composeappengine_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  // TODO: AppEngine API doesn't provide mean(s) to check if App is installed
+  ASSERT_TRUE(app_engine->install(app));
+  ASSERT_FALSE(app_engine->isRunning(app));
+}
+
+TEST_F(ComposeAppEngineeTest, FetchAndRun) {
+  auto app = registry.addApp(fixtures::ComposeApp::create("app-01"));
+  ASSERT_TRUE(app_engine->fetch(app));
+  ASSERT_TRUE(app_engine->run(app));
+  ASSERT_TRUE(app_engine->isRunning(app));
```
doanac, line 36: is calling `run` without calling `install` a valid thing?

doanac (commented):
thanks for re-working this. it made it a lot easier to understand. I just had one silly question.

Commit b6a38d54d2921da3cfb726adde5acec5edbe53a8 is a bit to digest, but it makes it all worth it when you see bda3d83df06302b52015d97bfb49bf865b2199a2

tests/composeappengine_test.cc
mike-sul: That's what ComposeAppManager does it invokes `docker-compose up -d` if just Apps are updated. If both, ostree and Apps then it invokes `docker-compose up --no-start` (aka install). https://github.com/foundriesio/aktualizr-lite/blob/master/src/compose/composeappmanager.cc#L253

doanac (approved):

PR #81, opened 2021-04-06
fixup! get currently installed apps correctly

mike-sul:
This is a regression issue, the fix just reverts back the mistakenly made change.

(my new tests, I am working on, has discovered this issue). 


doanac (approved):
wow. good catch. this regression came in after v80, right?

mike-sul:
> wow. good catch. this regression came in after v80, right?

correct.

PR #82, opened 2021-04-08
Liteclient HSM test

ldts:


mike-sul:
@ldts please, rebase it on the latest master.

tests/device-gateway_fake.py
```diff
@@ -18,6 +18,9 @@ class Handler(SimpleHTTPRequestHandler):
     AuthPrefix = "/hub-creds/"
     RegistryAuthPrefix = "/token-auth/"
     RegistryPrefix = "/v2/"
+    def do_POST(self):
```
mike-sul, line 21: I think it makes sense to remove this change from the given commit and add it to this one https://github.com/foundriesio/aktualizr-lite/pull/82/commits/789b83d982095c26a7e09ed0e7fcceaf3f4e1116.

tests/liteclientHSM_test.cc
```diff
@@ -0,0 +1,1179 @@
...
+  MOCK_METHOD(bool, run, (const App& app), (override));
+  MOCK_METHOD(void, remove, (const App& app), (override));
+  MOCK_METHOD(bool, isRunning, (const App& app), (const, override));
+};
+
+/**
+ * Class SysRootFS
+ *
+ */
+class SysRootFS {
```
mike-sul, line 64: I moved the common test components into the "fixture" directory https://github.com/foundriesio/aktualizr-lite/blob/master/tests/fixtures/liteclienttest.cc so you can make use of them by including them into the given test and remove their implementation from this test.

tests/liteclientHSM_test.cc
```diff
@@ -0,0 +1,1179 @@
...
+/**
+ * Class DeviceGatewayMock
+ *
+ */
+class DeviceGatewayMock {
+ public:
+  static std::string RunCmd;
+
+ public:
+  DeviceGatewayMock(const OSTreeRepoMock& ostree, const TufRepoMock& tuf, std::string certDir)
```
mike-sul, line 186: It makes sense to make certDir optional if we would like to reuse implementation.

tests/liteclientHSM_test.cc
```diff
@@ -0,0 +1,1179 @@
...
+  const std::string req_headers_file_;
+  boost::process::child process_;
+};
+
+std::string DeviceGatewayMock::RunCmd;
+
+class RootCaPKI {
+ public:
+  RootCaPKI(std::string path, std::string key, std::string crt)
+      : key_(path + std::move(key)), crt_(path + std::move(crt)) {
```
mike-sul, line 225: Looks like there is no much sense in passing `path` into the ctor as a client can pass full paths to `key` and `crt` or as an option, just a folder can be passed into ctor and assumed key and cert names (ca.key and ca.crt)?

tests/liteclientHSM_test.cc
```diff
@@ -0,0 +1,1179 @@
...
+ */
+class LiteClientHSMTest : public ::testing::Test {
+ public:
+  static std::string SysRootSrc;
+
+ protected:
+  static void SetUpTestSuite() {
+    boost::filesystem::path path = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
+    Utils::createDirectories(path, S_IRWXU);
+    hsm_ = new Hsm(path.string());
```
mike-sul, line 479: Why it has to be allocated in a heap?

mike-sul: Never mind, it's not important, I just thought about `static`.

tests/liteclientHSM_test.cc
```diff
@@ -0,0 +1,1179 @@
...
+ * Class LiteClientHSMTest
+ *
+ */
+class LiteClientHSMTest : public ::testing::Test {
+ public:
+  static std::string SysRootSrc;
+
+ protected:
+  static void SetUpTestSuite() {
+    boost::filesystem::path path = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
```
mike-sul, line 477: Maybe instead of `boost::filesystem::unique_path()` it makes sense to define this subfolder as `hsm` or `pki`, just makes debugging a bit easier since you clearly see where is a folder with PKI stuff.

mike-sul:
@ldts as I already mentioned it, our test container https://github.com/foundriesio/aktualizr-lite/blob/master/docker/Dockerfile is not fully correct because it uses a default libostree which is configured to use libsoup not libcurl. We didn't have any issue with it before because we never used mTLS in our tests. Now we use mTLS and libaktualizr sets key/certs to libcurl not to libsoup thus the test fails. Therefore, please, update the Dockerfile so it builds libostree in the way we do it for the local dev env https://github.com/foundriesio/aktualizr-lite/blob/master/how-to-run-locally.md#check-for-ostreelibostree-and-build-it-if-needed

tests/device-gateway_fake.py
ldts: ok moving it to its own commit (I dont think we want to hard reset the tree?)

tests/liteclientHSM_test.cc
ldts: ok

ldts:
I agree on the paths stuff. is all sedimented code...I initially wanted to be able to have difference devices on the same tests but not anymore.

ldts:
added some documentation to the HSM test sources (is easy to get lost otherwise)

docker/Dockerfile
```diff
@@ -81,6 +83,18 @@ RUN git clone https://github.com/pantoniou/libfyaml.git
 RUN cd libfyaml && ./bootstrap.sh && ./configure --prefix /usr && make && make install
 RUN ldconfig
 
+# OSTREE with libcurl support for HSM
+#  This allows ostree calls to libcurl to be able to keys controlled by pkcs11
+WORKDIR /tmp
+RUN mkdir ostree && cd ostree && git init && git remote add origin https://github.com/ldts/ostree && \
```
mike-sul, line 89: Maybe it makes sense to pull `ostree` from `https://github.com/ldts/ostree` and use the version we currently use in LmP, which is `v2020.7`. Your version/fork might be useful for debugging, I suppose. If you prefer to use it then maybe it makes sense to push/fork it to github.com/foundriesio/ostree?

ldts: argh, that shouldnt be there. my fault. will fix.

tests/fixtures/liteclienthsmtest.cc
```diff
@@ -0,0 +1,672 @@
...
+  std::string keyId_;
+  std::string certId_;
+
+ private:
+  std::string keyLabel_;
+  std::string csr_;
+  std::string crt_;
+};
+
+class ClientHSMTest : public ::testing::Test {
```
mike-sul, line 284: I wonder if it's possible to reuse the liteclient fixture https://github.com/foundriesio/aktualizr-lite/blob/7cf3104293638743542cd6e95d18ab8724f63399/tests/fixtures/liteclienttest.cc#L179, in the way it's reused in https://github.com/foundriesio/aktualizr-lite/blob/7cf3104293638743542cd6e95d18ab8724f63399/tests/aklite_test.cc#L19. It might require making the following method virtual ` std::shared_ptr<LiteClient> createLiteClient(const std::shared_ptr<AppEngine>& app_engine,
                                               InitialVersion initial_version = InitialVersion::kOn,
                                               boost::optional<std::vector<std::string>> apps = boost::none,
                                               const std::string& compose_apps_root = "") ` 
                                               
The idea is that `ClientHSMTest` implements just HSM&KPI stuff. Then `LiteClientHSMTest` will do multiple inheretence https://github.com/foundriesio/aktualizr-lite/pull/82/files#diff-520be724fef6b0a0102c23bea7df3d283005bb7d5e81f49765e541c9f2c5d187R55, `class LiteClientHSMTest: public fixtures::ClientTest, public fixtures::HSMTest`

ldts: sure I can do that. if we are not pressed for time I can rework things quite a bit now that I have a better picture of the end goal.  I could represent the rootCA as a persistent entity along with the server...anyway, ok., let me rework it all a bit

ldts: I think each mock should also bit on its on file with some -minimal- comments as to what it does. will do that as well

ldts: @mike-sul have a look

mike-sul: @ldts It looks fine at first glance, will check it out thoroughly tomorrow morning.

tests/fixtures/liteclient/devicehsm.cc
```diff
@@ -0,0 +1,115 @@
...
+*  3) Creates a CSR and requests its CRT from the RootCA
+*  4) Imports the CRT into the HSM.
+*  5) Generates HSM debug information if needed
+*
+*/
+class DeviceHsm {
+ public:
+  DeviceHsm(SoftHsm* hsm, RootCaPKI& rootCa, std::string conf) : hsm_(hsm), rootCa_(rootCa), cnf_(hsm_->path_ + conf) {
+    std::ofstream cnfOut(cnf_);
+    cnfOut << "openssl_conf = oc\n";
```
mike-sul, line 15: Maybe it makes sense to use a string literal for the config template, e.g. https://github.com/foundriesio/aktualizr/blob/e2482e8426e328dac63e8abd7cc55bd68994e44c/src/libaktualizr/utilities/xml2json_test.cc#L50 and then use std::snprintf(). Not critical, just might improve readability as well as might be useful for manual testing, so we can just copy-paste the config template.

tests/fixtures/liteclienttest.cc
```diff
@@ -1,198 +1,67 @@
...
-/**
- * Class ClientTest
- *
- */
 class ClientTest :virtual public ::testing::Test {
  public:
   static std::string SysRootSrc;
 
  protected:
+#if defined(USE_HSM)
```
mike-sul, line 21: I'd try to avoid `#if` for the HSM. For example by extending ClientTest `class ClientHSMTest: public ClientTest` then actuall test would do `class Test: public ClientHSMTest`, or `class ClientHSMTest: virtual public ::testing::Test` and then `class Test: public ClientTest, public ClientHSMTest`.
But, that's just a matter of personal taste, so that's OK for testing :).

mike-sul (approved):

tests/fixtures/liteclienttest.cc
ldts: agree 100%:  I'd never use macros in a production environment (is really a bad smell...just like any kind of conditional statements to be honest, that is why classes exist )...but since you mention, let me take a stab at it (Even though I think it will make the code a bit more confusing)

ldts: @mike-sul pushed implementing an interface.
I think this code makes the implicit needs for enabling HSM perhaps bit more obscure to those not familiar with the code but is more proper.

ldts:
just squashed the commit (dont think they had too much value on their own).
I believe I am done with this. please let me know

mike-sul:
> just squashed the commit (dont think they had too much value on their own).
> I believe I am done with this. please let me know

That's great! Thanks!

PR #83, opened 2021-04-14
doc: add info about enabling of HSM support

mike-sul:


ldts:
looks good to me. 

my lessons learnt was -from the hsm acces POV- that libcurl uses openssh which must be configured for pkcs11 whereas libsoap has no openssh access and therefore can not be used to accessed an HSM

PR #84, opened 2021-04-14
docker ps

ldts:


src/docker/dockerclient.cc
```diff
@@ -13,7 +14,7 @@ DockerClient::HttpClientFactory DockerClient::DefaultHttpClientFactory = []() {
 DockerClient::DockerClient(std::shared_ptr<HttpInterface> http_client) : http_client_{std::move(http_client)} {}
 
 bool DockerClient::serviceRunning(const std::string& app, const std::string& service, const std::string& hash) {
-  updateContainerStatus();
+  refresh();
```
doanac, line 17: i hit submit on a comment right after your force-push update. i wonder if making this class stateful via `updateContainerStatus` is going to cause confusion like this in the future? or maybe its just a couple of places and its easy to reason about?

ldts:
we probably should add a  lock in case we are refreshing from some other thread; will do

src/docker/dockerclient.cc
ldts: AFAIU I think it makes sense to always work on the latest info (just as we would do if we were using ps); but we need to add a lock on the refresh method

doanac:
> we probably should add a lock in case we are refreshing from some other thread; will do

That's sort of the opposite of my point. Maybe this class shouldn't carry state around at all. i.e. its either a stateless object that populates attributes when its constructed. then its more like a `ContainerStatus` struct/object. or you make it an actual "Client" where `refresh` returns the data to the caller and there's no state kept around between calls.

ldts:
Funny that you mention this because that is the way I initially coded it - then it has morphed into something different that needs a lock.
My initial thought was just an object that when instantiated would http to docker and then do the necessary processing. But that is no longer what it is (hence why I extended it). I would be happy to go back to the initial implementation, just let me know

mike-sul:
> Funny that you mention this because that is the way I initially coded it - then it has morphed into something different that needs a lock.
> My initial thought was just an object that when instantiated would http to docker and then do the necessary processing. But that is no longer what it is (hence why I extended it). I would be happy to go back to the initial implementation, just let me know
> 
> [bf451d5](https://github.com/foundriesio/aktualizr-lite/commit/bf451d5fcd318dc79d455a9665017ab80aae8fd7)

The goal is to query the docker daemon once per update cycle (for all apps) instead of querying it for each App. My plan is to make the refresh (or similar) method public and invoke it from the composeapp manager, e.g. https://github.com/foundriesio/aktualizr-lite/blob/93ddf54d129787d031c8d10c6882345b41be4176/src/composeappmanager.cc#L112 in the current implementation. I just didn't want to do it before pulling the "update logic" simplification for the "cleaning" branch. Also, the initial implementation doesn't allow "injection" of the mock http client for DockerClient.

mike-sul:
> we probably should add a lock in case we are refreshing from some other thread; will do

there is no other thread. There are just two threads, one that does update and the second that send report events. The latter doesn't invoke/deal with composeappmanager nor appengine.

mike-sul:
We need to have GetRunningApps() in the AppEngine interface anyway, it will allow adding another type of app engine easily in the future, for example ostree-based apps that are not executed by the docker daemon.

src/docker/dockerclient.cc
```diff
@@ -13,7 +13,8 @@ DockerClient::HttpClientFactory DockerClient::DefaultHttpClientFactory = []() {
 DockerClient::DockerClient(std::shared_ptr<HttpInterface> http_client) : http_client_{std::move(http_client)} {}
 
 bool DockerClient::serviceRunning(const std::string& app, const std::string& service, const std::string& hash) {
-  updateContainerStatus();
+  boost::mutex::scoped_lock lock(lock_);
```
mike-sul, line 16: there is no need for thread sync here since an update cycle happens within a single thread.

src/appengine.h
```diff
@@ -19,6 +19,7 @@ class AppEngine {
   virtual bool run(const App& app) = 0;
   virtual void remove(const App& app) = 0;
   virtual bool isRunning(const App& app) const = 0;
+  virtual std::string containerDetails() const = 0;
```
mike-sul, line 22: maybe `getRunningApps()` instead of `containerDetails`. Also, returning Json can be more flexible, clients can construct a string/output they need in their specific use-case.

mike-sul:
Please, add a test into this test suite https://github.com/foundriesio/aktualizr-lite/blob/master/tests/composeappengine_test.cc for the given functionality.

doanac:
> The goal is to query the docker daemon once per update cycle (for all apps) instead of querying it for each App. My plan is to make the refresh (or similar) method public and invoke it from the composeapp manager, e.g. 

This is where i'm hitting friction with our code.  I want to call it once, but I want it to be safe. e.g. you also need the running apps *after* the install for reporting. We are tying both running apps state and the tool to get it into a stateful object.

I think the DockerClient should *always* talk to the daemon. That's implied in the name of the class to me. But we don't want to call it constantly. So - maybe the GetRunningApps is the piece that calls this for the update cycle *one time*. and then passes an immutable "running apps" value/object to the packagemanager.

src/docker/dockerclient.cc
ldts: the lock is protecting the access to the root_; if there is no other thread in the system the lock can be removed. but assuming these things is _always_ a bad recipe bearing in mind that the architecture might evolve and that properly protecting variables cost nothing, just knowledge....

src/appengine.h
ldts: sure, just named it using the original function it replaced....will do runningApps()

ldts:
whatever gets decided must to be done explicitly....if it is one per update cycle, then we need a cycle_update state somewhere ?(I havent found the state machine yet in the code, could you point me to it?)

src/appengine.h
ldts: sure I'll return Json (the original code was returning a string)

mike-sul:
> So - maybe the GetRunningApps is the piece that calls this for the update cycle _one time_. and then passes an immutable "running apps" value/object to the packagemanager.

yep. That was my plan.

src/appengine.h
ldts: what format does the json value needs?

mike-sul:
@ldts LGTM. Now, it's time to remove this field https://github.com/foundriesio/aktualizr-lite/blob/8eedafc0a85a521d979e0990664de0bb4b4d3815/src/composeappmanager.h#L25.

mike-sul (approved):
@ldts Feel free to merge it once you verify it in a real Factory and a device.
There are two follow-up action items:

1. There is a bug related to this functionality. Running apps are reported only in the case of an ostree update, if there is just an App update then currently running apps are not reported to the backend.  We need a test that detects this issue as well as an actuall fix.

2. Make a request to the docker daemon per App not per App service.

PR #87, opened 2021-04-19
Bootloader

ldts:
as discussed, implement the bootloader functionality in aktualizr-lite

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -209,6 +209,7 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
       LOG_ERROR << "Failed to install OSTree target, skipping Docker Compose Apps";
       return res;
     }
+    const_cast<ComposeAppManager*>(this)->installNotify(target);
   } else {
     LOG_INFO << "Target " << target.sha256Hash() << " is same as current";
     res = data::InstallationResult(data::ResultCode::Numeric::kOk, "OSTree hash already installed, same as current");


src/bootloader/bootloaderlite.cc
```diff
@@ -0,0 +1,85 @@
...
+  reboot_detect_supported_ = true;
+}
+
+void BootloaderLite::setBootOK() const { rollback_->setBootOK(); }
+
+void BootloaderLite::updateNotify() const { rollback_->updateNotify(); }
+
+bool BootloaderLite::supportRebootDetection() const { return reboot_detect_supported_; }
+
+bool BootloaderLite::rebootDetected() const {
```
mike-sul, line 37: It looks like most of the methods are copy-paste from libaktualizr's Bootloader. If so, then they can be removed from here since BootloaderLite inherits Bootloader.

src/bootloader/bootloaderlite.h
```diff
@@ -0,0 +1,41 @@
...
+  void rebootFake();
+
+ private:
+  const BootloaderConfig config_;
+
+  INvStorage& storage_;
+  std::unique_ptr<Rollback> rollback_;
+  boost::filesystem::path reboot_sentinel_;
+  std::string reboot_command_;
+  bool reboot_detect_supported_{false};
```
mike-sul, line 38: why do we need to duplicate the fields from the class we inherit from?

src/bootloader/bootloaderlite.cc
ldts: what this means is that the intention - as it was via the abstract interface- is to not reuse anything from the base class.

Now, apparently this is now not the objective (I am sure I misunderstood); so that being the case, I'd much rather go back to having an abstract class which is the way it should be....so clients can deliver their own implementations.

src/bootloader/bootloaderlite.h
ldts: because my understanding was that we would not reuse anything from the base class.

If that is not the case, then really we should have an abstract interface and clients providing their bootloader specific implementations. why would you want to depend on the libaktualizr more than what is necesseray (ie, we only need the interfaces)

mike-sul: >why would you want to depend on the libaktualizr implementation more than what is necessary (ie, we only need the interfaces)

I don't want, but we already depend on ostreemanager that wants an object that implements Bootloader. If you prefer you can skip inheritance then.

src/bootloader/rollbacks/fiovb.h
```diff
@@ -35,9 +35,21 @@ class FiovbRollback : public Rollback {
   }
 
   void installNotify(const Uptane::Target& target) {
+    std::string version = getVersion(target);
+    if (version.empty()) {
+      return;
```
ldts, line 40: @igoropaniuk can you confirm that this is what you expect?

src/bootloader/rollbacks/fiovb.h
```diff
@@ -35,9 +35,21 @@ class FiovbRollback : public Rollback {
...
+    }
     std::string sink;
-    if (Utils::shell("fiovb_setenv bootupgrade_available 1", &sink) != 0) {
-      LOG_WARNING << "Failed to set bootupgrade_available";
+    if (Utils::shell("fiovb_printenv bootfirmware_version", &sink) != 0) {
+      LOG_WARNING << "Failed to read bootfirmware_version";
+      return;
```
ldts, line 45: @igoropaniuk same here

src/bootloader/rollbacks/rollback.h
```diff
@@ -10,6 +16,36 @@ class Rollback {
   virtual void setBootOK() {}
   virtual void updateNotify() {}
   virtual void installNotify(const Uptane::Target& target) { (void)target; }
+
+ protected:
+  std::string getVersion(const Uptane::Target& target) {
+    std::string file;
+    for (auto& p : boost::filesystem::directory_iterator("/ostree/deploy/lmp/deploy/")) {
```
ldts, line 23: @igoropaniuk also notice that we will iterate over all directories looking for one with the sha256 (apparently they could be named sha.1, sha.2, sha.0.....so this covers all the cases)

mike-sul (approved):

src/bootloader/rollbacks/fiovb.h
```diff
@@ -0,0 +1,57 @@
...
+    if (Utils::shell("fiovb_setenv bootcount 0", &sink) != 0) {
+      LOG_WARNING << "Failed resetting bootcount";
+    }
+    if (Utils::shell("fiovb_setenv upgrade_available 1", &sink) != 0) {
+      LOG_WARNING << "Failed setting upgrade_available";
+    }
+    if (Utils::shell("fiovb_setenv rollback 0", &sink) != 0) {
+      LOG_WARNING << "Failed resetting rollback flag";
+    }
+    if (Utils::shell("fiovb_setenv bootupgrade_available 1", &sink) != 0) {
```
igoropaniuk, line 32: we should not set any flags here, only in `installNotify`

src/bootloader/rollbacks/fiovb.h
igoropaniuk: yes, we should not update if we can not obrtain read version txt file.

src/bootloader/rollbacks/fiovb.h
igoropaniuk: Here it's different. We have to consider the case when `bootfirmware_version` is not set yet (for example, just after flashing via uuu)

@ricardosalveti could you please provide your input here

igoropaniuk (commented):
In general LGTM, just some minor comments

src/bootloader/rollbacks/fiovb.h
ricardosalveti: Yes, if not yet set we can set to the current version.

src/bootloader/rollbacks/rollback.h
ricardosalveti: Can we have this path and the firmware/version.txt path provided via config? Or would that be too much? Just that hardcoding it here might end up biting us later.

src/bootloader/rollbacks/fiovb.h
```diff
@@ -0,0 +1,57 @@
...
+    std::string sink;
+    if (Utils::shell("fiovb_setenv bootcount 0", &sink) != 0) {
+      LOG_WARNING << "Failed resetting bootcount";
+    }
+    if (Utils::shell("fiovb_setenv upgrade_available 0", &sink) != 0) {
+      LOG_WARNING << "Failed resetting upgrade_available";
+    }
+  }
+
+  void updateNofify() {
```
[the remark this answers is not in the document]
ldts: thanks!

src/bootloader/rollbacks/rollback.h
ldts: will change

PR #89, opened 2021-05-05
Report running Apps on each update type

mike-sul:


doanac (approved):

src/liteclient.cc
```diff
@@ -78,13 +77,14 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, bool
...
   http_client = std::make_shared<HttpClient>(&headers);
-  uptane_fetcher_ = std::make_shared<Uptane::Fetcher>(config, http_client);
-  report_queue = std_::make_unique<ReportQueue>(config, http_client, storage);
 
   key_manager_ = std_::make_unique<KeyManager>(storage, config.keymanagerConfig());
   key_manager_->loadKeys();
   key_manager_->copyCertsToCurl(*http_client);
 
+  uptane_fetcher_ = std::make_shared<Uptane::Fetcher>(config, http_client);
+  report_queue = std_::make_unique<ReportQueue>(config, http_client, storage);
```
doanac, line 86: this is that long standing bug where we saw weird ssl stuff at start up, right?

mike-sul: I hope so :)

mike-sul:
@doanac It turned out to be a rabbit hole because `runningApps` needs dockerd so `tests/test_lite.sh` became dependent on it hence suddenly we need to run dockerd in a container for the CI run... :)
Anyway, I think it's good to go.

src/liteclient.cc
```diff
@@ -78,7 +79,7 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, bool
 
   http_client = std::make_shared<HttpClient>(&headers);
 
-  key_manager_ = std_::make_unique<KeyManager>(storage, config.keymanagerConfig());
+  key_manager_ = std_::make_unique<KeyManager>(storage, config.keymanagerConfig(), p11);
```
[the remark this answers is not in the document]
mike-sul: Not necessary, if `p11` is not specified then KeyManager will create a default one, which is OK in the default case.

doanac (approved):

PR #91, opened 2021-05-10
aklite: switch to the updated 2021.1+fio branch

mike-sul:


mike-sul:
@doanac I've forgotten to switch aklite to the tip of the aktualizr branch we currently use.

PR #93, opened 2021-05-21
Assume TUF repo server is Hub creds provider

mike-sul:
Use the TUF repo (DG) URL to compose an URL to obtain creds from for Docker Registry.
Previously, the URL to Treehub/Ostree server was used for this purpose.
Since we are splitting Device Gateway into two servers, Device Gateway (TUF) and OstreeProxy then their URLs will differ and the Auth endpoint is exposed by Device Gateway/TUF server.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -17,7 +17,7 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, bool
   storage = INvStorage::newStorage(config.storage, false, StorageClient::kTUF);
   storage->importData(config.import);
 
-  const std::map<std::string, std::string> raw = config.pacman.extra;
+  std::map<std::string, std::string>& raw = config.pacman.extra;
   if (raw.count("tags") == 1) {
     std::string val = raw.at("tags");
     if (val.length() > 0) {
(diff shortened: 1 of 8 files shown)


src/liteclient.cc
```diff
@@ -34,6 +34,18 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, bool
...
+  const auto& repo_endpoint = config.uptane.repo_server;
+  std::string auth_creds_endpoint = Docker::RegistryClient::DefAuthCredsEndpoint;
+  if (!repo_endpoint.empty()) {
+    auto endpoint_pos = repo_endpoint.rfind('/');
+    if (endpoint_pos != std::string::npos) {
+      auth_creds_endpoint = repo_endpoint.substr(0, endpoint_pos);
+      auth_creds_endpoint.append("/hub-creds/");
+    }
+  }
+  raw["hub_auth_creds_endpoint"] = auth_creds_endpoint;
```
doanac, line 47: thanks for bubbling this up to the lite-client. It used to bother me being buried so deep down in the code.

doanac (approved):

PR #95, opened 2021-05-24
bootloader/masked: call fw_printenv without variable name

ricardosalveti:
Call fw_printenv with no-header (-n) when extracting the
bootfirmware_version content, otherwise it will be printed together with
the variable name, invalidating the string comparison logic to decide
when to set bootupgrade_available to 1.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/src/bootloader/rollbacks/masked.h
+++ b/src/bootloader/rollbacks/masked.h
@@ -37,7 +37,7 @@ class MaskedRollback : public Rollback {
       return;
     }
     std::string sink;
-    if (Utils::shell("fw_printenv bootfirmware_version", &sink) != 0) {
+    if (Utils::shell("fw_printenv -n bootfirmware_version", &sink) != 0) {
       LOG_WARNING << "Failed to read bootfirmware_version";
       return;
     }


src/bootloader/rollbacks/masked.h
```diff
@@ -37,7 +37,7 @@ class MaskedRollback : public Rollback {
       return;
     }
     std::string sink;
-    if (Utils::shell("fw_printenv bootfirmware_version", &sink) != 0) {
+    if (Utils::shell("fw_printenv -n bootfirmware_version", &sink) != 0) {
```
mike-sul, line 40: iiuc, it's not an issue for this case https://github.com/foundriesio/aktualizr-lite/blob/18996f6d276661ee7971d21a6ce7a2d67e4295d0/src/bootloader/rollbacks/fiovb.h#L43 ?

mike-sul (approved):

mike-sul:
@ldts It would be great to add some test(s) for this use-case if it's feasible at all.

src/bootloader/rollbacks/masked.h
ricardosalveti: That is something that I will check today.

igoropaniuk: I think that's not the case, fiovb doesn't print variable name, otherwise I would have stumbled upon the same issue when I was testing my closed imx8mmevk (but anyway worth checking)

@ricardosalveti

igoropaniuk (approved):
LGTM

PR #97, opened 2021-06-02
apps: don't check if Target Apps are running

mike-sul:
aklite doesn't need to check if Apps' containers are running and
(re-)start them. Let docker-compose/docker to monitor and maintain container status.
https://github.com/compose-spec/compose-spec/blob/master/spec.md#restart

### DIFF
--- a/src/docker/composeappengine.cc
+++ b/src/docker/composeappengine.cc
@@ -69,9 +69,12 @@ bool ComposeAppEngine::isRunning(const App& app) const {
 }
 
 bool ComposeAppEngine::isInstalled(const App& app) const {
-  // make sure the App root dir and compose file exists
+  // make sure the App root dir and compose file exists and a version of downloaded App macthes a given one
   bool result = boost::filesystem::exists(appRoot(app)) &&
-                boost::filesystem::exists(appRoot(app) / Docker::ComposeAppEngine::ComposeFile);
+                boost::filesystem::exists(appRoot(app) / Docker::ComposeAppEngine::ComposeFile) && checkVersion(app);
+  // TODO: check if a given app was really "installed", i.e. "up -d" or "up -d --no-start" was excuted for it
+  // and corresponding container images are present on a system
+  // and corresponding containers are created (no need to check if running)
   return result;
 }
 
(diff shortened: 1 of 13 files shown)


src/docker/composeappengine.cc
```diff
@@ -68,6 +68,16 @@ bool ComposeAppEngine::isRunning(const App& app) const {
...
 }
 
+bool ComposeAppEngine::isInstalled(const App& app) const {
+  // make sure the App root dir and compose file exists and a version of downloaded App macthes a given one
+  bool result = boost::filesystem::exists(appRoot(app)) &&
+                boost::filesystem::exists(appRoot(app) / Docker::ComposeAppEngine::ComposeFile) && checkVersion(app);
+  // TODO: check if a given app was really "installed", i.e. "up -d" or "up -d --no-start" was excuted for it
+  // and corresponding container images are present on a system
+  // and corresponding containers are created (no need to check if running)
+  return result;
```
doanac, line 78: I was going to ask the we remove the `isRunning` method. I *think* that's dead code now. However, these TODO's make me think that maybe part of isRunning needs to be merged into isInstalled to do what you are outlining. In away this TODO is really asking for an `isRunning`. I wonder if we should keep the `isRunning` name for this instead?

mike-sul: I've been considering it too and decided to leave it as it is now until the TODO is implemented.
First of all, `isRunning()` is still used by many tests to make sure that App is really "running", the docker-compose mock used in tests cannot control mocked Apps.

Secondly, as you are implying, `isRunning` is still can be used but will have a different meaning and corresponding implementation.
I think `isInstalled` name matches better what the method actually does and will do in the future. So, I think we will need both, the TODO functionality is split between isInstalled and isRunning.
In any case, I think this is subject to another PR and maybe RFD doc, meanwhile the current implementation addresses the "one-shot" app issue, so it's not restarted on each update cycle.

mike-sul:
I am closing this PR as it turned out to be the wrong direction at all after careful e2e testing of different types of restart policies in conjunction with different apps and archs.

PR #98, opened 2021-06-07
bootloader: fiovb: don't set bootupgrade_available

igoropaniuk:
```
Don't set bootupgrade_available flag in UpdateNotify(), as
during that step we're not able to detect if we have boot firmware
updates (this is decided conditionally in InstallNotify()).

Reported-by: Tim Anderson <tim.anderson@foundries.io>
Signed-off-by: Igor Opaniuk <igor.opaniuk@foundries.io>
```


Tim-Anderson (approved):
That looks to be the correct change. LGTM

mike-sul (approved):

PR #101, opened 2021-06-14
Store and make use of App update state

mike-sul:
Track App update state and make use of it, so
* network usage is optimized, aklite doesn't re-fetch what has been already fetched;
* simplify debugging and troubleshooting;
* extend the context of report events with App state info.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -207,7 +207,7 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
     res = OstreeManager::install(target);
     if (res.result_code.num_code == data::ResultCode::Numeric::kInstallFailed) {
       LOG_ERROR << "Failed to install OSTree target, skipping Docker Compose Apps";
-      res.description += "\n# Apps running:\n" + runningApps();
+      res.description += "\n# Apps running:\n" + getRunningAppsInfoForReport();
       return res;
     }
     const_cast<ComposeAppManager*>(this)->installNotify(target);
@@ -275,7 +275,7 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
     }
   }
 
-  res.description += "\n# Apps running:\n" + runningApps();
+  res.description += "\n# Apps running:\n" + getRunningAppsInfoForReport();
   return res;
 }
 
@@ -295,7 +295,7 @@ data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target
   }
 
   if (data::ResultCode::Numeric::kNeedCompletion != ir.result_code.num_code) {
-    ir.description += "\n# Apps running:\n" + runningApps();
+    ir.description += "\n# Apps running:\n" + getRunningAppsInfoForReport();
   }
   return ir;
 }
@@ -334,5 +334,21 @@ void ComposeAppManager::handleRemovedApps(const Uptane::Target& target) const {
 }
 
 std::string ComposeAppManager::getCurrentHash() const { return sysroot_->getCurDeploymentHash(); }
+Json::Value ComposeAppManager::getRunningAppsInfo() const { return app_engine_->getRunningAppsInfo(); }
+std::string ComposeAppManager::getRunningAppsInfoForReport() const {
+  std::string result;
+  const auto apps_info = getRunningAppsInfo();
+  for (Json::ValueConstIterator ii = apps_info.begin(); ii != apps_info.end(); ++ii) {
+    result += ii.key().asString() + ": " + (*ii)["uri"].asString() + "; state: " + (*ii)["state"].asString() + "\n";
+    ;
 
-std::string ComposeAppManager::runningApps() const { return app_engine_->runningApps(); }
+    Json::Value services = (*ii)["services"];
+    for (Json::ValueConstIterator jj = services.begin(); jj != services.end(); ++jj) {
+      result += "\t" + jj.key().asString() + ": " + (*jj)["hash"].asString();
+      result += "; image: " + (*jj)["image"].asString();
+      result += "; state: " + (*jj)["state"].asString();
+      result += "; status: " + (*jj)["status"].asString() + "\n";
+    }
+  }
+  return result;
+}
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -282,15 +282,10 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
 data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target& target) {
   auto ir = OstreeManager::finalizeInstall(target);
 
-  const auto& current_apps = getApps(target);
-  for (const auto& app_pair : current_apps) {
-    const auto& app_name = app_pair.first;
-    auto need_start_flag = cfg_.apps_root / app_name / Docker::ComposeAppEngine::NeedStartFile;
-    if (boost::filesystem::exists(need_start_flag)) {
-      if (ir.result_code.num_code == data::ResultCode::Numeric::kOk) {
-        app_engine_->run({app_pair.first, app_pair.second});
-      }
-      boost::filesystem::remove(need_start_flag);
+  if (ir.result_code.num_code == data::ResultCode::Numeric::kOk) {
+    // "finalize" (run) Apps that were pulled and created before reboot
+    for (const auto& app_pair : getApps(target)) {
+      app_engine_->run({app_pair.first, app_pair.second});
     }
   }
 
--- a/src/docker/composeappengine.cc
+++ b/src/docker/composeappengine.cc
@@ -12,28 +12,120 @@ ComposeAppEngine::ComposeAppEngine(boost::filesystem::path root_dir, std::string
     : root_{std::move(root_dir)},
       compose_{std::move(compose_bin)},
       docker_client_{std::move(docker_client)},
-      registry_client_{std::move(registry_client)} {
-  boost::filesystem::create_directories(root_);
-}
+      registry_client_{std::move(registry_client)} {}
 
 bool ComposeAppEngine::fetch(const App& app) {
-  boost::filesystem::create_directories(appRoot(app));
-  if (download(app)) {
-    LOG_INFO << "Validating compose file";
-    if (cmd_streaming(compose_ + "config", app)) {
-      LOG_INFO << "Pulling containers";
-      return cmd_streaming(compose_ + "pull --no-parallel", app);
+  boost::filesystem::create_directories(appRoot(app) / MetaDir);
+
+  bool result = false;
+
+  try {
+    AppState state(app, appRoot(app), true);
+
+    if (!download(app)) {
+      state.setState(AppState::State::kDownloadFailed);
+      return result;
+    }
+    state.setState(AppState::State::kDownloaded);
+
+    if (!verify(app)) {
+      state.setState(AppState::State::kVerifyFailed);
+      return result;
+    }
+    state.setState(AppState::State::kVerified);
+
+    if (!pullImages(app)) {
+      state.setState(AppState::State::kPullFailed);
+      return result;
+    }
+    state.setState(AppState::State::kPulled);
+
+    result = true;
+  } catch (const std::exception& exc) {
+    LOG_WARNING << "Failed to get/set App state, fallback to forced fetch: " << exc.what();
+
+    if (download(app) && verify(app) && pullImages(app)) {
+      result = true;
     }
   }
-  return false;
+
+  return result;
 }
 
 bool ComposeAppEngine::install(const App& app) {
-  boost::filesystem::ofstream flag_file(appRoot(app) / NeedStartFile);
-  return cmd_streaming(compose_ + "up --remove-orphans --no-start", app);
+  bool result = false;
+
+  if (!boost::filesystem::exists(appRoot(app))) {
+    LOG_ERROR << "App dir doesn't exist, cannot install App that hasn't been fetched";
+    return result;
+  }
+
+  try {
+    AppState state(app, appRoot(app));
+
+    switch (state()) {
+      case AppState::State::kPulled:
+      case AppState::State::kInstallFail: {
+        if (!installApp(app)) {
+          state.setState(AppState::State::kInstallFail);
+          break;
+        }
+        state.setState(AppState::State::kInstalled);
+      }
+      case AppState::State::kInstalled:
+      case AppState::State::kStarted:
+        result = true;
+        break;
+      default:
+        LOG_ERROR << "Cannot install App that hasn't been fetched";
+    }
+  } catch (const std::exception& exc) {
+    LOG_WARNING << "Failed to get/set App state, fallback to forced install: " << exc.what();
+
+    if (installApp(app)) {
+      result = true;
+    }
+  }
+  return result;
 }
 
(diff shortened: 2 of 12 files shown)


src/docker/composeappengine.cc
```diff
@@ -193,4 +320,77 @@ void ComposeAppEngine::extractAppArchive(const App& app, const std::string& arch
...
+    throw std::runtime_error("Failed to write to file, " + filepath_ + ", error: " + strerror(errno));
+  }
+  fsync(fd_);
+}
+
+void ComposeAppEngine::AppState::File::write(const std::string& val) {
+  lseek(fd_, 0, SEEK_SET);
+  ssize_t wr = ::write(fd_, val.data(), val.size());
+  if (wr == -1) {
+    throw std::runtime_error("Failed to write to file, " + filepath_ + ", error: " + strerror(errno));
```
doanac, line 373: I think `std::system_error` would be a more natural fit. like:
~~~
throw std::system_error(errno, std::generic_category(), "Unable to write file");
~~~

src/docker/composeappengine.cc
```diff
@@ -193,4 +320,77 @@ void ComposeAppEngine::extractAppArchive(const App& app, const std::string& arch
...
+  }
```
doanac, line 374: if the length of `val` shrinks, I think you'll need a call to something like `ftruncate`. but it seems like a safer/simpler approach would be writing to a tempfile and then doing a rename of the file. this would prevent a user making a concurrent call to read the file from getting inconsistent data.

src/docker/composeappengine.cc
```diff
@@ -193,4 +320,77 @@ void ComposeAppEngine::extractAppArchive(const App& app, const std::string& arch
   }
 }
 
+ComposeAppEngine::AppState::AppState(const App& app, const boost::filesystem::path& root, bool store_version) try
+    : version_file_ {
+  (root / MetaDir / VersionFile).string()
+}
+, state_file_{(root / MetaDir / StateFile).string()} {
+  version_ = version_file_.readStr();
```
doanac, line 328: i'm guessing you've thought through this, but its sort of feeling like this would be easier and more atomic if we had a single state file that had both values.

src/docker/composeappengine.cc
```diff
@@ -12,28 +12,120 @@ ComposeAppEngine::ComposeAppEngine(boost::filesystem::path root_dir, std::string
...
+      case AppState::State::kUnknown:
+      case AppState::State::kDownloadFailed: {
+        if (!download(app)) {
+          state.setState(AppState::State::kDownloadFailed);
+          break;
+        }
+        state.setState(AppState::State::kDownloaded);
+      }
+      case AppState::State::kDownloaded:
+      case AppState::State::kVerifyFailed: {
```
doanac, line 34: if verify failed, shouldn't we try to fetch again?

doanac: ah - i guess you need VerifyFailed in both places.

doanac (commented):
this makes me a little nervous. i like removing the logic to re-download the compose-app but always skipping the verification of the compose file and pulling the containers (which is a no-op unless something bad has happened) seems more dangerous than what we have now.

Maybe we just need less state to this. I'm think all we want is: DOWNLOADED, STARTED. Then - you always fetch( basically a no-op unless things are really bad). And you always start if not STARTED. I *think* that would allow you to not need the "INSTALLED" state.

mike-sul:
@doanac This https://github.com/foundriesio/aktualizr-lite/pull/102 is effectively a lock for the version/state files.

src/docker/composeappengine.cc
mike-sul: I was planning to do the renaming stuff, this is still WIP.

src/docker/composeappengine.cc
mike-sul: Not sure so. First of all, a version changes only once, while there are many states, so writing hash/URL on each state update is kind of useless. Also, writing both URI and state into a single file makes business logic slightly more complicated.
In terms of atomicity, it is atomic from AppState user perspective.

src/docker/composeappengine.cc
mike-sul: What for?

mike-sul:
> i like removing the logic to re-download the compose-app but always skipping the verification of the compose file and pulling the containers (which is a no-op unless something bad has happened) seems more dangerous than what we have now.

Both, `docker-compose config` (forking heavy python process) and `docker pull` (get token from DG by using aktualizr-get and fetching manifest ) are relatively expensive operations. 
What's the danger do you see?

mike-sul:
> Maybe we just need less state to this. I'm think all we want is: DOWNLOADED, STARTED. Then - you always fetch( basically a no-op unless things are really bad). And you always start if not STARTED. I _think_ that would allow you to not need the "INSTALLED" state.

Having states for each phase is, at least, useful in terms of debugging/troubleshooting, I was planning to add the current state into the report event. Having it in the report event helps to figure out the fetch/installation failure, not just fetch or install failed, but more specific, app fetch failed, image pull failed, etc.

@doanac Let's talk about it today after the standup.

mike-sul:
Also, I am not sure I understand why you are saying this `Then - you always fetch( basically a no-op unless things are really bad)`, `docker-compose pull` is a quite expensive operation even if all images are already present on a system.

1. Spawning a new process;
2. Loading python stuff;
3. Running the docker cred helper and aktualizr-get;
4. Making calls to dockerd;
5. dockerd makes several calls to Docker Registry.

src/docker/composeappengine.cc
doanac: I assume verifyfailed could mean one of 2 things: Computing the hash failed (no sense retrying). The actual payload is wrong. In this case we should try and download it again, or the device will never apply the update?

src/docker/composeappengine.cc
mike-sul: done

src/docker/composeappengine.cc
mike-sul: done

src/docker/composeappengine.cc
mike-sul: as discussed

mike-sul:
With the last commit we get the following output. I think it would be better if aklite sends actually serialized json to the beackend instead of a string. In this case, fioctl and UI will receive the info about running apps in the form of json so they can decide how to render/visualize the info output.

```
2021-06-16T11:32:53+00:00 : EcuDownloadStarted(intel-corei7-64-lmp-925)
 Details:
 | Sync active Target Apps
2021-06-16T11:33:05+00:00 : EcuDownloadCompleted(intel-corei7-64-lmp-925) -> Succeed
2021-06-16T11:33:05+00:00 : EcuInstallationStarted(intel-corei7-64-lmp-925)
2021-06-16T11:33:07+00:00 : EcuInstallationCompleted(intel-corei7-64-lmp-925) -> Succeed
 Details:
 | OSTree hash already installed, same as current
 | # Apps installed:
 | hub.foundries.io/msul-dev01/app-05@sha256:ca57113ff6242c17ad09e892448556a6916a28807c988c42e0763f5ec4e640ed
 | hub.foundries.io/msul-dev01/app-07@sha256:ee881b3b2542b8fa4a370bddb99fad607c827014c89a2cfd0cc338ba7e539df2
 | # Apps running:
 | app-05: hub.foundries.io/msul-dev01/app-05@sha256:ca57113ff6242c17ad09e892448556a6916a28807c988c42e0763f5ec4e640ed; state: Started
 | 	python-www: 4de16414efc53b3c84a640108bb7e7c6890d665f58358e058c195fc2041bc0b7; image: sha256:b0a2e960f732fdea26ccfaea95db03d18d4a32ca4d74f6ea1d3e2e53f8865424; state: running; status: Up 12 minutes
 | app-07: hub.foundries.io/msul-dev01/app-07@sha256:ee881b3b2542b8fa4a370bddb99fad607c827014c89a2cfd0cc338ba7e539df2; state: Started
 | 	nginx-01: 5e2f32d07385d348654361ce0021541f6d3d2801416434bcff6c6b1990f6bd80; image: hub.foundries.io/msul-dev01/nginx@sha256:c83ec3a0e87e2f3039c8234b157381af41081d42109bb35f418b851356e5e846; state: running; status: Up 3 hours
 | 	nginx-02: b920517847f29f65b975e441751225e6028f330f591592c5a5eaa72386da65a4; image: docker.io/library/nginx@sha256:a97eb9ecc708c8aa715ccfb5e9338f5456e4b65575daf304f108301f3b497314; state: running; status: Up 3 hours
 | 	python-www: 614b6e67f54940065fe305ce572dfe9b020d7f390a8fdc9fb70ee8f7624946d6; image: hub.foundries.io/msul-dev01/app-07@sha256:3ee70a119a620086afbfa8705bf222edd09b9ae9d41653749752c04089cad21d; state: running; status: Up 18 minutes
 | 

```

mike-sul:
@doanac I think that's all that I wanted to implement in this round/iteration of aklite improving.

src/docker/composeappengine.cc
```diff
@@ -12,28 +12,111 @@ ComposeAppEngine::ComposeAppEngine(boost::filesystem::path root_dir, std::string
...
+      return result;
+    }
+    state.setState(AppState::State::kPulled);
+
+    result = true;
+  } catch (const std::exception& exc) {
+    LOG_WARNING << "Failed to get/set App state, fallback to forced fetch: " << exc.what();
+
+    if (download(app) && verify(app) && pullImages(app)) {
+      result = true;
```
doanac, line 48: should we also try to call `state.setState(AppState::State::kPulled);` here and hope for the best?

src/docker/composeappengine.cc
```diff
@@ -12,28 +12,111 @@ ComposeAppEngine::ComposeAppEngine(boost::filesystem::path root_dir, std::string
...
   }
-  return false;
+
+  return result;
 }
 
 bool ComposeAppEngine::install(const App& app) {
-  boost::filesystem::ofstream flag_file(appRoot(app) / NeedStartFile);
-  return cmd_streaming(compose_ + "up --remove-orphans --no-start", app);
+  boost::filesystem::create_directories(appRoot(app) / MetaDir);
```
doanac, line 56: can this be called without this existing? If it doesn't exist should we even bother trying to install?

src/docker/composeappengine.cc
```diff
@@ -44,6 +127,19 @@ void ComposeAppEngine::remove(const App& app) {
 }
 
 bool ComposeAppEngine::isRunning(const App& app) const {
+  if (!boost::filesystem::exists(appRoot(app) / MetaDir)) {
```
doanac, line 130: maybe AppState should be able to see this and just return an unknown state?

src/docker/composeappengine.cc
```diff
@@ -193,4 +315,78 @@ void ComposeAppEngine::extractAppArchive(const App& app, const std::string& arch
...
+  int wr = ::write(fd, data, size);
+  if (-1 == wr) {
+    close(fd);
+    throw std::system_error(errno, std::system_category(), std::string("Failed to read from a file: ") + path_);
+  }
+  fsync(fd);
+  close(fd);
+
+  ::rename(tmp_file.c_str(), path_.c_str());
+  ::remove(tmp_file.c_str());
```
doanac, line 420: remove shouldn't be needed unless rename failed and we ignore that, so i'm no sure its worth calling remove?

src/docker/composeappengine.h
```diff
@@ -2,6 +2,7 @@
 #define AKTUALIZR_LITE_COMPOSE_APP_ENGINE_H_
 
 #include <string>
+#include <unordered_map>
```
doanac, line 5: not sure i see where you've added code to use an unordered_map?

src/composeappmanager.cc
```diff
@@ -334,5 +334,21 @@ void ComposeAppManager::handleRemovedApps(const Uptane::Target& target) const {
...
+  for (Json::ValueConstIterator ii = apps_info.begin(); ii != apps_info.end(); ++ii) {
+    result += ii.key().asString() + ": " + (*ii)["uri"].asString() + "; state: " + (*ii)["state"].asString() + "\n";
+    ;
+
+    Json::Value services = (*ii)["services"];
+    for (Json::ValueConstIterator jj = services.begin(); jj != services.end(); ++jj) {
+      result += "\t" + jj.key().asString() + ": " + (*jj)["hash"].asString();
+      result += "; image: " + (*jj)["image"].asString();
+      result += "; state: " + (*jj)["state"].asString();
+      result += "; status: " + (*jj)["status"].asString() + "\n";
```
doanac, line 345: i worried status wouldn't be interesting, but it kind of is. this will let you see if things restarted that shouldn't have?

src/docker/composeappengine.cc
```diff
@@ -168,7 +169,47 @@ bool ComposeAppEngine::isRunning(const App& app) const {
...
+
+    for (Json::ValueIterator ii = containers.begin(); ii != containers.end(); ++ii) {
+      Json::Value val = *ii;
+
+      std::string app_name = val["Labels"]["com.docker.compose.project"].asString();
+      if (app_name.empty()) {
+        continue;
+      }
+
+      App app{app_name, ""};
```
doanac, line 186: seems like `App app` is only used for calling `state(app,..)`. Maybe just keep it scoped in that conditional block

doanac (commented):
looks good. just some small comments

src/composeappmanager.cc
mike-sul: yes, and not only that. In the case of failure, it will indicate the exact state at which it had happened. In general, more context is always better.

src/docker/composeappengine.cc
mike-sul: Not sure. First of all the `state` object is not available in the given context.
Also, fallback to this code happens only if the state object initialization or setting a state yields an exception, so invoking state.setState() will yield an exception most likely, it's kind of to the same again.

src/docker/composeappengine.cc
mike-sul: Sure, it doesn't make sense in the case of using only from the compose app manager, which might break some tests, but you're right we should return with an error. On the other hand, such case should never happened if used through the compose app manager.

src/docker/composeappengine.cc
mike-sul: AppState will yield an exception in this case which will lead to the fallback scenario which is checking running Apps through the docker client.

src/docker/composeappengine.cc
mike-sul: Yes, I actually forgot to check the return value for `rename`.

src/docker/composeappengine.h
mike-sul: Some left over, I tried to be smart with the state management but then decided that it's not worth doing it at this stage.

mike-sul:
@doanac updated according to the comments.

src/docker/composeappengine.cc
doanac: Just feels like some API friction here. It seems like you have to know the internals of AppState here to call it correctly? If AppState switched its internals to use `/var/run....` this code becomes invalid?

mike-sul: sure it should be just `exists(appRoot(app)` as in `run` and `install`

mike-sul: updated

doanac (approved):

PR #102, opened 2021-06-15
Aklite lock

mike-sul:
Make sure that just one instance of 'aklite daemon' or 'aklite update' is running simultaneously.

### DIFF
--- a/src/liteclient.h
+++ b/src/liteclient.h
@@ -26,7 +26,7 @@ class Lock {
 
 class LiteClient {
  public:
-  LiteClient(Config& config_in, const std::shared_ptr<AppEngine>& app_engine = nullptr, bool finalize = true,
+  LiteClient(Config& config_in, const std::shared_ptr<AppEngine>& app_engine = nullptr,
              const std::shared_ptr<P11EngineGuard>& p11 = nullptr);
 
   Config config;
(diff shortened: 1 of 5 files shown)


src/liteclient.h
```diff
@@ -40,7 +40,7 @@ class LiteClient {
 
   bool checkForUpdatesBegin();
   void checkForUpdatesEnd(const Uptane::Target& target);
-  bool complete(bool finalize);
+  bool complete();
```
doanac, line 43: I like this change. While we are at this, maybe its time to give this method a better name since its now public. e.g. `complete_previous_install` or something.

src/main.cc
```diff
@@ -14,6 +15,35 @@
...
+    }
+  }
+  FileLock(const FileLock&) = delete;
+  FileLock(const FileLock&&) = delete;
+  FileLock& operator=(const FileLock&) = delete;
+  FileLock& operator=(const FileLock&&) = delete;
+
+ private:
+  const char* const path_{nullptr};
+  boost::filesystem::ofstream file_;
```
doanac, line 43: Is this needed so that the file gets automagically created before instantiate the boost::file_lock? Seems like you could do that all in the constructor and not need to keep this open fd around?

src/liteclient.h
mike-sul: Jorge wanted something shorter :)... I would prefer to stick to the aktualizr's terms, which is finalize_installation :).

src/main.cc
mike-sul: It has to be created before `lock_` is initialized in the initialization list.

src/liteclient.h
doanac: I think `finalize_install{ation}` actually makes more sense also.

mike-sul: renamed.

doanac (approved):

PR #103, opened 2021-06-15
Remove the download/update lock

mike-sul:
The current download and update locks are not used.
Since we added the aklite system-wide lock that prevents from running two simultaneous update sessions they become obsolete/redundant.


doanac (commented):
this looks good once we confirm no customer has used this.

doanac (approved):

PR #104, opened 2021-06-17
bootloader: update fw if bootfirmware_version is not set

Tim-Anderson:
On a closed board, if the bootfirmware_version is not set
it means that an upgrade is need to properly set the version.

Signed-off-by: Tim Anderson <tim.anderson@foundries.io>

### DIFF
--- a/src/bootloader/rollbacks/fiovb.h
+++ b/src/bootloader/rollbacks/fiovb.h
@@ -39,7 +39,7 @@ class FiovbRollback : public Rollback {
     std::string sink;
     if (Utils::shell("fiovb_printenv bootfirmware_version", &sink) != 0) {
       LOG_WARNING << "Failed to read bootfirmware_version";
-      return;
+      sink = std::string();
     }
     LOG_INFO << "Current firmware version: " << sink;
     if (sink.compare(version) != 0) {


ricardosalveti:
For context the different behavior when comparing with uboot env is what caused the issue, as fiovb will return an error when the variable doesn't exists, which is not the case for fw_printenv.

igoropaniuk (approved):
LGTM

src/bootloader/rollbacks/fiovb.h
```diff
@@ -39,7 +39,7 @@ class FiovbRollback : public Rollback {
     std::string sink;
     if (Utils::shell("fiovb_printenv bootfirmware_version", &sink) != 0) {
       LOG_WARNING << "Failed to read bootfirmware_version";
-      return;
+      sink = std::string();
```
mike-sul, line 42: So, `sink` won't be empty if the fiovb command fails? Maybe it would be useful to print it before nulling if it contains some useful error messages.

mike-sul (approved):

src/bootloader/rollbacks/fiovb.h
ricardosalveti: @Tim-Anderson did you check if sink ended up with an error there? I wonder if fiovb_printenv will dump the error to stderr instead.

Tim-Anderson: To clarify. The optee error "E/TA: ..." is output to the console only. The "READ_PVALUE failed: 0xffff0006 / 4"  and the "Cant print the environment" are both going to stderr so stdout will have no output.

ricardosalveti (approved):
LGTM

PR #106, opened 2021-06-18
ostree: cancel ostree download session if hung

mike-sul:
Add a download watchdog to workaround the issue in libostree related to resuming of download after network goes down for a longer period of time (~ 15-20 minutes) and then goes up again.
TODO: add more details about the issue.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -311,13 +311,10 @@ bool LiteClient::checkImageMetaOffline() {
   return true;
 }
 
-std::pair<bool, Uptane::Target> LiteClient::downloadImage(const Uptane::Target& target,
-                                                          const api::FlowControlToken* token) {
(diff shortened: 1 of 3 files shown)


mike-sul:
@doanac It doesn't help to cancel aklite/ostree download session when network is down, it just simply hangs (at least I wait for more than 1 hour). But if network goes up the fix makes ostree cancel the download session and aklite will start a new one if there are still download tries (3), if not the target download, as well as an overall Target installation, fails and aklite will try to install it again on the next update cycle.

mike-sul:
@doanac Unfortunately the "fix" is not a fix, I just wasn't patient enough to wait for > 15/20  minutes.

mike-sul:
close in favor of https://github.com/foundriesio/meta-lmp/commit/2b742b96341a245b2a1c66bb93a837345e533c63

PR #108, opened 2021-06-24
apps: use DEBUG level msg if no app state file

mike-sul:
The warning message might scarry users as they appear during the transition to the latest aklite version that supports App state tracking. Specifically, if aklite is updated and one or more of the Apps are not updated then the state files won't be present on a system so isRunning() check cannot be accomplished by means of the state files. The PR just set DEBUG level for the message informing about it.


doanac (approved):
thanks. that scared me just the other day.

PR #110, opened 2021-07-13
apps: check downloaded App contains a compose file

mike-sul:
Make sure that a new Target App, that akite downloads, includes a compose file (docker-compose.yml).

### DIFF
--- a/src/docker/composeappengine.cc
+++ b/src/docker/composeappengine.cc
@@ -143,7 +143,15 @@ bool ComposeAppEngine::isRunning(const App& app) const {
   bool started_state = false;
   try {
     AppState state(app, appRoot(app));
-    started_state = state() == AppState::State::kStarted;
+    if (app.uri == state.version()) {
+      started_state = state() == AppState::State::kStarted;
+    } else {
+      // The state file exists but it describes a state of some other App not the one
+      // that specified via the param. It can happen if a new Target App download or start fails
+      // and aklite checks if the current App is running
+      LOG_DEBUG << "Failed to get/set App state, fallback to checking the dockerd state";
+      started_state = true;
+    }
   } catch (const std::exception& exc) {
     LOG_DEBUG << "Failed to get/set App state, fallback to checking the dockerd state: " << exc.what();
     started_state = true;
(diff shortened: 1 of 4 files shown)


mike-sul:
@doanac did you have a chance to look into it?

src/docker/composeappengine.cc
```diff
@@ -351,6 +360,13 @@ bool ComposeAppEngine::checkAvailableStorageSpace(const boost::filesystem::path&
   return true;
 }
 
+void ComposeAppEngine::verifyAppArchive(const App& app, const std::string& archive_file_name) {
+  if (!cmd_streaming("tar -tf " + archive_file_name + " " + ComposeFile, app) &&
+      !cmd_streaming("tar -tf " + archive_file_name + " ./" + ComposeFile, app)) {
```
doanac, line 365: are both checks really required? seems like the `./` check will never do anything?

src/docker/composeappengine.cc
```diff
@@ -143,7 +143,15 @@ bool ComposeAppEngine::isRunning(const App& app) const {
...
     AppState state(app, appRoot(app));
-    started_state = state() == AppState::State::kStarted;
+    if (app.uri == state.version()) {
+      started_state = state() == AppState::State::kStarted;
+    } else {
+      // The state file exists but it describes a state of some other App not the one
+      // that specified via the param. It can happen if a new Target App download or start fails
+      // and aklite checks if the current App is running
+      LOG_DEBUG << "Failed to get/set App state, fallback to checking the dockerd state";
+      started_state = true;
```
doanac, line 153: i'm not quite understanding why we don't do `started_state = state() == AppState::State::kStarted;` here also?

doanac (commented):
i just have some questions to help me understand this better.

src/docker/composeappengine.cc
mike-sul: files can be prepended with `./` prefix in the app archive, both options are valid so I am checking both. The first check expects the compose file without the prefix, which is our default case, so under normal circumstances, just one condition is executed.

src/docker/composeappengine.cc
mike-sul: Because a client calling isRunning() is asking about the App a version of which (current target) differs from a version of App (new Target) state of which is a recorder/stored in the state file.
It can happen when aklite tried to download a new Target's App and failed or a download succeeded but start failed and after that somehow isRunning is called for a current Target, not a new one. Or if a rollback happens after a reboot then the state file refers to a state of the new Target App while the liteclient asks if running the current App... 
also, this verification https://github.com/foundriesio/aktualizr-lite/blob/94577e0300948069912993984dab021849a3c9cb/tests/aklite_test.cc#L93 will fail if we do it.

doanac (approved):

PR #112, opened 2021-08-11
aklite: add ability to manage non-booted rootfs

mike-sul:
- Add ability to run aklite on a local host in the case of `ostree`
package manager usage. I.e. teach the ostree manager to manage a
standalone/non-booted ostree-based rootfs.

- Also, it makes the ostree package manager use the Fio's bootloader instead of the
libaktualizr's native one. The bootloader update won't work if the
libaktualizr's bootloader implementation is used.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,6 +18,7 @@ set(SRC helpers.cc
 
 set(HEADERS helpers.h
         composeappmanager.h
+        rootfstreemanager.h
         docker/composeappengine.h
         docker/composeapptree.h
         docker/composeinfo.h
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -9,6 +9,7 @@
 
 #include "composeappmanager.h"
 #include "crypto/keymanager.h"
+#include "rootfstreemanager.h"
 #include "target.h"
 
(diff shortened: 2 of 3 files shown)


mike-sul:
@pattivacek FYI

doanac (approved):

pattivacek:
Finally got a chance to test this. Looks good until I actually try to install an update, then it fails with `error: Unable to install update: No merge deployment`. I can probably hack on my local repo a bit to make that work, but that's not my priority for now. Thanks again!

mike-sul:
> Looks good until I actually try to install an update, then it fails with `error: Unable to install update: No merge deployment`

Was the pull successful? Sometimes ostree cannot pull some of the objects in `bare-user-only` because of the permission related issue. So, I just change `bare-user-only` to `bare-user`, pull commit and then bring it back to `bare-user-only` :).

mike-sul:
It looks like it's  another issue, most likely the repo/commit you pulled is not exactly rootfs that ostree expects. This is the guide https://github.com/foundriesio/aktualizr-lite/blob/master/how-to-run-locally.md#setup-a-system-rootfs-that-aktualizr-lite-will-be-managing how to make ostree happy about the rootfs.

pattivacek:
Nice, thanks. I forgot I'd taken some shortcuts before. I followed the instructions more carefully and I was able to get it to work this time without a problem. Cool!

PR #113, opened 2021-08-18
ostree pull from GCS with fallback to DG/ostreeproxy

mike-sul:
Pull ostree commit objects from directly from the GCS bucket(s). If remotes that points to the buckets are not available or pull from GCS fails then fallback to fetching via Device Gateway or OSTreeproxy, depending on the device/aklite configuration.

### DIFF
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -1,0 +1,60 @@
+#include "rootfstreemanager.h"
+#include "ostree/repo.h"
+
+bool RootfsTreeManager::fetchTarget(const Uptane::Target &target, Uptane::Fetcher &fetcher, const KeyManager &keys,
+                                    const FetcherProgressCb &progress_cb, const api::FlowControlToken *token) {
+  (void)fetcher;
+
+  std::vector<Remote> remotes = {{remote, config.ostree_server, {{"X-Correlation-ID", target.filename()}}, true}};
+
+  getAdditionalRemotes(remotes, target.filename());
+
+  data::InstallationResult pull_err{data::ResultCode::Numeric::kUnknown, ""};
+  for (const auto &remote : remotes) {
+    LOG_INFO << "Fetchig ostree commit " + target.sha256Hash() + " from " + remote.baseUrl;
+    if (!remote.isRemoteSet) {
+      setRemote(remote.name, remote.baseUrl);
+    }
+    pull_err = OstreeManager::pull(config.sysroot, remote.baseUrl, keys, target, token, progress_cb,
+                                   remote.isRemoteSet ? nullptr : remote.name.c_str(), remote.headers);
+    if (pull_err.isSuccess()) {
+      break;
+    }
+    LOG_INFO << "Failed to fetch from " + remote.baseUrl + ", err: " + pull_err.description;
+  }
+
+  if (!pull_err.isSuccess()) {
+    LOG_ERROR << "Failed to fetch ostree commit " + target.sha256Hash() + ", err: " + pull_err.description;
+  }
+
+  return pull_err.isSuccess();
+}
+
+void RootfsTreeManager::getAdditionalRemotes(std::vector<Remote> &remotes, const std::string &target_name) {
+  const auto resp = http_client_->post(gateway_url_ + "/download-urls", Json::Value::null);
+
+  if (!resp.isOk()) {
+    LOG_WARNING << "Failed to obtain download URLs from Gateway, fallback to dowload via gateway/proxy server: "
+                << resp.getStatusStr();
+    return;
+  }
+
+  const auto respJson = resp.getJson();
+  for (Json::ValueConstIterator it = respJson.begin(); it != respJson.end(); it++) {
+    remotes.emplace(
+        remotes.begin(), Remote{"gcs",
+                                (*it)["download_url"].asString(),
+                                {
+                                    {"X-Correlation-ID", target_name},
+                                    {"Authorization", std::string("Bearer ") + (*it)["access_token"].asString()},
+                                }}
+
+    );
+  }
+}
+
+void RootfsTreeManager::setRemote(const std::string &name, const std::string &url) {
+  OSTree::Repo repo{sysroot_->path() + "/ostree/repo"};
+
+  repo.addRemote(name, url, "", "", "");
+}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,6 +3,7 @@ set(TARGET ${MAIN_TARGET_LIB})
 
 set(SRC helpers.cc
         composeappmanager.cc
+        rootfstreemanager.cc
         docker/composeappengine.cc
         docker/composeapptree.cc
         docker/composeinfo.cc
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -53,9 +53,8 @@ ComposeAppManager::ComposeAppManager(const PackageConfig& pconfig, const Bootloa
                                      const std::shared_ptr<INvStorage>& storage,
                                      const std::shared_ptr<HttpInterface>& http,
                                      std::shared_ptr<OSTree::Sysroot> sysroot, AppEngine::Ptr app_engine)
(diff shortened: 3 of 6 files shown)


src/rootfstreemanager.cc
```diff
@@ -0,0 +1,15 @@
+#include "rootfstreemanager.h"
+
+bool RootfsTreeManager::fetchTarget(const Uptane::Target &target, Uptane::Fetcher &fetcher, const KeyManager &keys,
+                                    const FetcherProgressCb &progress_cb, const api::FlowControlToken *token) {
+  if (!target.IsOstree()) {
```
doanac, line 8: not your bug, but it seems like best way to handle this would be to put such a check in ostreemanager::pull logic since its the method requiring this.  no point fixing here i guess.

src/rootfstreemanager.cc
```diff
@@ -1,15 +1,34 @@
...
-  return OstreeManager::pull(config.sysroot, config.ostree_server, keys, target, token, progress_cb, headers).success;
+  return OstreeManager::pull(config.sysroot, config.ostree_server, keys, target, token, progress_cb,
+                             serverUrl.empty() ? nullptr : gcsRemote.c_str(), headers)
+      .success;
 }
 
-void RootfsTreeManager::getPullHeaders(RequestHeaders &headers) {}
+std::string RootfsTreeManager::getServerUrl(RequestHeaders &auth_header) {
+  (void)auth_header;
+  return "";
```
doanac, line 28: seems sane so far. I guess the next step will be deciding how to fill this logic in?

src/rootfstreemanager.cc
mike-sul: ok, just wanted to minimize changes to libaktualizr, but I think your're right, it's better to put into ostreemanager::pull although I feel like we will bypass it sooner or later since we have its all functional blocks in the scope of aklite project.

src/rootfstreemanager.cc
mike-sul: yep. I tested it locally by hard-coding the bucket URL and oauth token, works as expected.
The next step is to add a new endpoint to device-gateway or ostreeproxy that will provide an oauth token and download URL(s) to devices.

mike-sul: I think it makes sense to fallback to downloading via ostreeproxy in any case.

mike-sul:
Now, it works against this version of ostreehub & proxy https://github.com/foundriesio/ostreehub/pull/20.

src/rootfstreemanager.cc
```diff
@@ -24,11 +24,27 @@ bool RootfsTreeManager::fetchTarget(const Uptane::Target &target, Uptane::Fetche
...
+    return (*it)["download_url"].asString();
+  }
+
   return "";
 }
 
 void RootfsTreeManager::setRemote(const std::string &name, const std::string &url) {
   OSTree::Repo repo{sysroot_->path() + "/ostree/repo"};
+
+  LOG_DEBUG << "Setting remote to " << url;
```
doanac, line 48: i bet we'll wind up wanting this to be LOG_INFO for support

mike-sul: updated

doanac: i think things look pretty good. lets try and get this staged so we can all start playing with it.

src/rootfstreemanager.cc
mike-sul: updated.

mike-sul:
@doanac I tested the given version for the all possible scenarios (devices with old configurations, the ostree commit is not available on the new multi-regional buckets, only on the current one, etc) .

doanac (commented):
looking good. I usually don't say this, but this might be easier to read if you squash it now - its not a huge change in the end.

I'm sure you've thought about this and I know its a difficult part of the code to test ... Is there anything here we could add test coverage for or maybe our current coverage will still exercise these paths ?

mike-sul:
> looking good. I usually don't say this, but this might be easier to read if you squash it now - its not a huge change in the end.
> 
> I'm sure you've thought about this and I know its a difficult part of the code to test ... Is there anything here we could add test coverage for or maybe our current coverage will still exercise these paths ?

First of all, I've tested manually all key use-cases (including fallback), just would like to merge it sooner so the dev  factories will start dogfooding.

Sure, I thought about the automated tests, it shouldn't be tough to extend the existing tests with tests that cover the code path introduced in the given PR. 
I am just thinking about priority, specifically, whether I should do "factory reset" stuff first of all.

mike-sul:
@doanac squashed all PR commits.

doanac (approved):

doanac:
@mike-sul - was just doing a little testing and I'm not sure how you really intend this logic to pan out in the short term. I noticed my device was pulling from the EU bucket which i found odd. I added a patch that I think we probably want just to help support:
~~~
$ git diff
diff --git a/src/rootfstreemanager.cc b/src/rootfstreemanager.cc
index 668fcb6..663dd75 100644
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -8,6 +8,9 @@ bool RootfsTreeManager::fetchTarget(const Uptane::Target &target, Uptane::Fetche
   std::vector<Remote> remotes = {{remote, config.ostree_server, {{"X-Correlation-ID", target.filename()}}, true}};
 
   getAdditionalRemotes(remotes, target.filename());
+  for (const auto &r : remotes) {
+    LOG_INFO << "Using " << r.baseUrl << " as an ostree source";
+  }
 
   data::InstallationResult pull_err{data::ResultCode::Numeric::kUnknown, ""};
   for (const auto &remote : remotes) {
~~~

Let me know if you think that's sane, and I'll do a PR.

Then - I'm wonder about how we order the remotes. We call `remotes.emplace(` which kind of reverses the defined order they are in. Should we `emplace_back` or just swap the US/EU order to keep things working more aligned with our current production - or just switch to making more things go through EU?

mike-sul:
>was just doing a little testing and I'm not sure how you really intend this logic to pan out in the short term. I noticed my device was pulling from the EU bucket which i found odd.

According to the plan https://github.com/foundriesio/rftm/tree/main/adrs/0007#phase-0 it was supposed to be just one bucket (the US one) for the initial phase, then once it's tested and verified we are supposed to make decisions on the items described  in https://github.com/foundriesio/rftm/tree/main/adrs/0007#decision-and-implementation-plan, including "How to do Load Balancing across multiple GCS buckets (and ostreeproxy clusters if we decide that we need one in a few locations)?".

I just did extra step and decided to do with multi-bucket for the phase 0 and do dogfooding for it. I think the bucket order is not really important for the dogfooding. Once we done with testing then we need to make decision on it as described in the implementation plan. Implementation options are described in the document.

In addition to the described options, while we are making on decision on the dynamic load balancing, we could simply just 
1.  enable one bucket, just pick the one where we have the biggest amount of devices with poor networking (LTE);
2. just leave two buckets, and pick the first one where we have the biggest amount of devices with poor networking (LTE);
3. add a new parameter to lmp-device-register and corresponding one to sota.toml that specifies a location, so aklite uses it to sort the bucket list.

In general, I think we are not looking for speed, we are looking for reliability and robustness (HA, FT) first of all. I think it becomes important for LTE/4G/3G cases.

mike-sul:
>Let me know if you think that's sane, and I'll do a PR.

It looks redundant to me as the remote base URL is printed here https://github.com/foundriesio/aktualizr-lite/blob/a40a69b8dae0b4fcf7360c6d132ed6d294016864/src/rootfstreemanager.cc#L14.

doanac:
> > Let me know if you think that's sane, and I'll do a PR.
> 
> It looks redundant to me as the remote base URL is printed here
> 
> https://github.com/foundriesio/aktualizr-lite/blob/a40a69b8dae0b4fcf7360c6d132ed6d294016864/src/rootfstreemanager.cc#L14
> .

alright. i'll ignore for now. the problem i was having was i didn't know *why*. I'm guessing as we make the remote selection more dynamic the logging to answer this question will come along naturally though.

PR #115, opened 2021-09-09
Add new aktualizr-lite API

doanac:
This patchset introduces an easy-to-use API into aktualizr-lite to allow users to produce their own custom daemons with specialized logic for controlling how updates are processed.

The API is currently missing support for ComposeApps as well as support for managing secondary ECUs. These will come in future changes once this initial API is agreed on.

### DIFF
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -11,6 +11,41 @@
 class LiteClient;
 
 /**
+ * A high-level representation of a TUF Target in terms applicable to a
+ * FoundriesFactory.
+ */
+class TufTarget {
+ public:
+  TufTarget(std::string name, std::string sha256, int version, Json::Value custom)
+      : name_(std::move(name)), sha256_(std::move(sha256)), version_(version), custom_(std::move(custom)) {}
+
+  /**
+   * Return the TUF Target name. This is the key in the targets.json key/value
+   * singed.metadata dictionary.
+   */
+  const std::string &Name() const { return name_; }
+  /**
+   * Return the sha256 OStree hash of the Target.
+   */
+  const std::string &Sha256Hash() const { return sha256_; }
+  /**
+   * Return the FoundriesFactory CI build number or in TUF, custom.version.
+   */
+  int Version() const { return version_; }
+
+  /**
+   * Return TUF custom data for a Target.
+   */
+  const Json::Value &Custom() const { return custom_; }
+
+ private:
+  std::string name_;
+  std::string sha256_;
+  int version_;
+  Json::Value custom_;
+};
+
+/**
  * AkliteClient provides an easy-to-use API for users wanting to customize
  * the behavior of aktualizr-lite.
  */
@@ -34,6 +69,11 @@ class AkliteClient {
    * Return the active aktualizr-lite configuration.
    */
   boost::property_tree::ptree GetConfig() const;
+
+  /**
+   * Return the Target currently running on the system.
+   */
+  TufTarget GetCurrent() const;
 
   /**
    * Default files/paths to search for sota toml when configuration client.
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -65,6 +65,20 @@ class CheckInResult {
 };
 
 /**
+ * The response from an AkliteClient call to Install
+ */
+class InstallResult {
+ public:
+  enum class Status {
+    Ok = 0,
+    NeedsCompletion,
+    Failed,
+  };
+  Status status;
+  std::string description;
+};
+
+/**
  * The response from an AkliteClient call to Download()
  */
 class DownloadResult {
@@ -76,8 +90,10 @@ class DownloadResult {
   };
   Status status;
   std::string description;
+  std::function<InstallResult()> Install;
 };
 
+std::ostream &operator<<(std::ostream &os, const InstallResult &res);
 std::ostream &operator<<(std::ostream &os, const DownloadResult &res);
 
 /**
@@ -125,6 +141,11 @@ class AkliteClient {
    * Download and verify the content of the given target.
    */
   DownloadResult Download(const TufTarget &t, std::string reason = "");
+
+  /**
+   * Install the Target
+   */
+  InstallResult Install(const TufTarget &t);
 
   /**
    * Check if the Target has been installed but failed to boot. This would
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -1,0 +1,45 @@
+// Copyright (c) 2021 Foundries.io
+// SPDX-License-Identifier: Apache-2.0
+
+#include <string>
+
+#include <boost/filesystem.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include "json/json.h"
+
+class LiteClient;
+
+/**
+ * AkliteClient provides an easy-to-use API for users wanting to customize
+ * the behavior of aktualizr-lite.
+ */
+class AkliteClient {
+ public:
+  /**
+   * Construct a client instance pulling in config files from the given
+   * locations. ex:
+   *
+   *   AkliteClient c(AkliteClient::CONFIG_DIRS)
+   *
+   * @param config_dirs The list of files/directories to parse sota toml from.
+   */
+  AkliteClient(const std::vector<boost::filesystem::path> &config_dirs);
+  /**
+   * Used for unit-testing purposes.
+   */
+  AkliteClient(std::shared_ptr<LiteClient> client) : client_(client) {}
+
+  /**
+   * Return the active aktualizr-lite configuration.
+   */
+  boost::property_tree::ptree GetConfig() const;
+
+  /**
+   * Default files/paths to search for sota toml when configuration client.
+   */
+  static std::vector<boost::filesystem::path> CONFIG_DIRS;
+
+ private:
+  std::shared_ptr<LiteClient> client_;
+};
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -106,6 +106,13 @@ class AkliteClient {
   TufTarget GetCurrent() const;
 
   /**
+   * Check if the Target has been installed but failed to boot. This would
+   * make this be considered a "rollback target" and one we shouldn't consider
+   * installing.
+   */
+  bool IsRollback(const TufTarget &t) const;
+
+  /**
    * Default files/paths to search for sota toml when configuration client.
    */
   static std::vector<boost::filesystem::path> CONFIG_DIRS;
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -65,6 +65,22 @@ class CheckInResult {
 };
 
 /**
+ * The response from an AkliteClient call to Download()
+ */
+class DownloadResult {
+ public:
+  enum class Status {
+    Ok = 0,
+    DownloadFailed,
+    VerificationFailed,
+  };
+  Status status;
+  std::string description;
+};
+
+std::ostream &operator<<(std::ostream &os, const DownloadResult &res);
+
+/**
  * AkliteClient provides an easy-to-use API for users wanting to customize
  * the behavior of aktualizr-lite.
  */
@@ -104,6 +120,11 @@ class AkliteClient {
    * Return the Target currently running on the system.
    */
   TufTarget GetCurrent() const;
+
+  /**
+   * Download and verify the content of the given target.
+   */
+  DownloadResult Download(const TufTarget &t, std::string reason = "");
 
   /**
    * Check if the Target has been installed but failed to boot. This would
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -46,6 +46,25 @@ class TufTarget {
 };
 
 /**
+ * The response from an AkliteClient call to CheckIn()
+ */
+class CheckInResult {
+ public:
+  enum class Status {
+    Ok = 0,    // check-in was good
+    OkCached,  // check-in failed, but locally cached meta-data is still valid
+    Failed,    // check-in failed and there's no valid local meta-data
+  };
+  CheckInResult(Status status, std::vector<TufTarget> targets) : status(status), targets_(std::move(targets)) {}
+  Status status;
+  const std::vector<TufTarget> &Targets() const { return targets_; }
+  TufTarget GetLatest() const { return targets_.back(); }
+
+ private:
+  std::vector<TufTarget> targets_;
+};
+
+/**
  * AkliteClient provides an easy-to-use API for users wanting to customize
  * the behavior of aktualizr-lite.
  */
@@ -66,6 +85,17 @@ class AkliteClient {
   AkliteClient(std::shared_ptr<LiteClient> client) : client_(client) {}
 
   /**
+   * Performs a "check-in" with the device-gateway. A check-in consists of:
+   *  1) Report sota.toml. Only do once.
+   *  2) Report network info. Only done once unless it changes aftert startup
+   *  3) Report hardware info. Only done once.
+   *  4) ask device-gateway for a new root.json - normally a 404.
+   *  5) ask device-gateway for timestamp and snapshot metadata.
+   *  6) pull down a new targets.json if needed
+   */
+  CheckInResult CheckIn() const;
+
+  /**
    * Return the active aktualizr-lite configuration.
    */
   boost::property_tree::ptree GetConfig() const;
@@ -82,4 +112,5 @@ class AkliteClient {
 
  private:
   std::shared_ptr<LiteClient> client_;
+  mutable bool configUploaded_{false};
 };
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(TARGET_EXE aktualizr-lite)
+set(TARGET_LIB aktualizr_lite)
 set(TARGET ${MAIN_TARGET_LIB})
 
 set(SRC helpers.cc
@@ -41,6 +42,7 @@ set(HEADERS helpers.h
 
 add_executable(${TARGET_EXE} main.cc)
 add_library(${TARGET} OBJECT ${SRC})
+add_library(${TARGET_LIB} SHARED ${SRC} api.cc)
 
 if(ALLOW_MANUAL_ROLLBACK)
   add_definitions(-DALLOW_MANUAL_ROLLBACK)
@@ -62,14 +64,16 @@ set(INCS
 
 target_include_directories(${TARGET} PRIVATE ${INCS})
 target_include_directories(${TARGET_EXE} PRIVATE ${INCS})
+target_include_directories(${TARGET_LIB} PRIVATE ${AKLITE_DIR}/include ${INCS})
 
 target_link_libraries(${TARGET} aktualizr_lib)
+target_link_libraries(${TARGET_LIB} aktualizr_lib)
 target_link_libraries(${TARGET_EXE} ${TARGET})
-
 
 # TODO: consider cleaning up the overall "install" elements as it includes
 # redundant targets (e.g aktualizr-secondary, aktualizr-cert-provider, etc)
 install(TARGETS ${TARGET_EXE} RUNTIME DESTINATION bin COMPONENT ${TARGET_EXE})
+install(TARGETS ${TARGET_LIB} LIBRARY DESTINATION lib COMPONENT ${TARGET_LIB})
 
 # enable creating clang-tidy targets for each source file (see aktualizr/CMakeLists.txt for details)
-aktualizr_source_file_checks(main.cc ${SRC} ${HEADERS})
+aktualizr_source_file_checks(main.cc api.cc ../include/aktualizr-lite/api.h ${SRC} ${HEADERS})
--- a/src/api.cc
+++ b/src/api.cc
@@ -22,3 +22,14 @@ boost::property_tree::ptree AkliteClient::GetConfig() const {
   boost::property_tree::ini_parser::read_ini(ss, pt);
   return pt;
 }
+
+TufTarget AkliteClient::GetCurrent() const {
+  auto current = client_->getCurrent();
+  int ver = -1;
+  try {
+    ver = std::stoi(current.custom_version(), nullptr, 0);
+  } catch (const std::invalid_argument& exc) {
+    LOG_ERROR << "Invalid version number format: " << current.custom_version();
+  }
+  return TufTarget(current.filename(), current.sha256Hash(), ver, current.custom_data());
+}
--- a/src/api.cc
+++ b/src/api.cc
@@ -1,11 +1,26 @@
 #include "aktualizr-lite/api.h"
 
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
 #include "helpers.h"
 #include "libaktualizr/config.h"
 #include "liteclient.h"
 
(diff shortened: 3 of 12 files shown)


doanac:
Here's an example main.cc I'm using for testing: https://paste.debian.net/1211026/

mike-sul (commented):
I think it makes sense to put the API header(s) under `include/aklite/` or `include/libaklite/` so a client will do `#include "aklite/<api-header>" and add `-I /<aklite-path>/include`. This is kind of one of the patterns and iirc this is how libaktualizr does.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <string>
+
+#include <boost/filesystem.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include "json/json.h"
+
+class LiteClient;
```
mike-sul, line 11: Is it part of the API?

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * A high-level representation of a TUF Target in terms applicable to a
+ * FoundriesFactory.
+ */
+class TufTarget {
```
mike-sul, line 17: It looks like a structure at a first glance, not sure if it's bad or good though :).

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+class CheckInResult {
+ public:
+  enum class Status {
+    Ok,        // check-in was good
+    OkCached,  // check-in failed, but locally cached meta-data is still valid
+    Failed,    // check-in failed and there's no valid local meta-data
+  };
+  CheckInResult(Status status) : status(status) {}
+  Status status;
+  std::vector<TufTarget> Targets() const { return targets_; }
```
mike-sul, line 60: Maybe the return type can be `const std::vector<>& ` so the vector copying is avoided.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+class TufTarget {
+ public:
+  TufTarget(std::string name, std::string sha256, int version, Json::Value custom)
+      : name_(std::move(name)), sha256_(std::move(sha256)), version_(version), custom_(std::move(custom)) {}
+
+  /**
+   * Return the TUF Target name. This is the key in the targets.json key/value
+   * singed.metadata dictionary.
+   */
+  std::string Name() const { return name_; }
```
mike-sul, line 26: I suppose the getters could return `const std::string&`?

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+  Json::Value custom_;
+};
+
+/**
+ * The response from an AkliteClient call to CheckIn()
+ */
+class CheckInResult {
+ public:
+  enum class Status {
+    Ok,        // check-in was good
```
mike-sul, line 54: I don't think it's really important, just a matter of a coding guideline, the first value should be explicitly defined `Ok = 0`

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+ * The response from an AkliteClient call to Download()
+ */
+class DownloadResult {
+ public:
+  enum class Status {
+    Ok,
+    DownloadFailed,
+    VerificationFailed,
+  };
+  DownloadResult(Status status, std::string description = "") : status(status), description(std::move(description)) {}
```
mike-sul, line 78: I think the ctor could be redundant in this case, something like `DownloadResult res{<status>, <descr>}` or `return DownloadResult{<status>, <descr>}` should work.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+   * @param config_dirs The list of files/directories to parse sota toml from.
+   */
+  AkliteClient(const std::vector<boost::filesystem::path> &config_dirs);
+  /**
+   * Used for unit-testing purposes.
+   */
+  AkliteClient(std::shared_ptr<LiteClient> client) : client_(client) {}
+
+  /**
+   * Performs a "check-in" with the device-gateay. A check-in consists of:
```
mike-sul, line 119: typo `gateay` --> `gateway`

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+ public:
+  /**
+   * Construct a client instance pulling in config files from the given
+   * locations. ex:
+   *
+   *   AkliteClient c(AkliteClient::CONFIG_DIRS)
+   *
+   * @param config_dirs The list of files/directories to parse sota toml from.
+   */
+  AkliteClient(const std::vector<boost::filesystem::path> &config_dirs);
```
mike-sul, line 112: Probably, it makes sense, at the final stage of the API development, document the params that AkliteClinet accepts, the params defined inside of the toml files.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+   */
+  AkliteClient(std::shared_ptr<LiteClient> client) : client_(client) {}
+
+  /**
+   * Performs a "check-in" with the device-gateay. A check-in consists of:
+   *  1) ask device-gateway for a new root.json - normally a 404.
+   *  2) ask device-gateway for timestamp and snapshot metadata.
+   *  3) pull down a new targets.json if needed
+   */
+  CheckInResult CheckIn(std::string hwid = "") const;
```
mike-sul, line 124: What does `hwid` param is needed for? Isn't it defined in the one of toml files passed during the client instantiation? Or this is for the option that implies that each ECU has it's own target?

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Return the active aktualizr-lite configuration.
+   */
+  boost::property_tree::ptree GetConfig() const;
+
+  /**
+   * Report network and hwinfo to server. This only reports data if the
+   * information has changed locally.
+   */
+  void SendTelemetry() const;
```
mike-sul, line 135: This is something that might be use-case/client specific, I mean each client might wanna define it's own set of telemetry data to send.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Report the full aktualizr-lite active configuration to the
+   * device-gateway. Should be called once during start-up .
+   */
+  void ReportAkliteConfiguration() const;
```
mike-sul, line 141: It sounds like the akliteclient internal stuff, exposing it to clients might be an error prone.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Default files/paths to search for sota toml when configuration client.
+   */
+  std::vector<boost::filesystem::path> CONFIG_DIRS = {"/usr/lib/sota/conf.d", "/var/sota/sota.toml",
+                                                      "/etc/sota/conf.d/"};
+
+  /**
+   * Return a random UUID to be used for a Download/Install operation
+   */
+  static std::string GenerateCorrelationId(const TufTarget &t);
```
mike-sul, line 174: It could be an internal stuff, aklite client could generate it internally if a client doesn't specify it in the Download() and  Install() calls?

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+
+ private:
+  std::shared_ptr<LiteClient> client_;
```
mike-sul, line 177: It doesn't look like something that should be part of the external API. Maybe this API should be pure interface (just set of pure virtual functions) or abstract class. So, this `std::shared_ptr<LiteClient> client_;` becomes a member of the interface/base-class subclass?

src/CMakeLists.txt
```diff
@@ -41,6 +42,7 @@ set(HEADERS helpers.h
 
 add_executable(${TARGET_EXE} main.cc)
 add_library(${TARGET} OBJECT ${SRC})
+add_library(${TARGET_LIB} SHARED ${SRC} api.cc)
```
mike-sul, line 45: `akite.cc` instead of `api.cc`?

include/aklite.h
doanac: our backend isn't capable of arbitrary telemetry?

include/aklite.h
doanac: the same value needs to be in both calls to correlate the download and install.

doanac:
> I think it makes sense to put the API header(s) under include/aklite/ or include/libaklite/ so a client will do #include "aklite/<api-header>" and add -I //include`. This is kind of one of the patterns and iirc this is how libaktualizr does.

those are the internal headers for aktualizr. the external headers are at https://github.com/uptane/aktualizr/tree/master/include

include/aklite.h
doanac: It's something a user needs control of. they need to decide both if and when to send. I thought about trying to embed it in the SendTelemetry method. Makes things a little messy because the client object become stateful (has to keep a "has_sent_config" state), but that's another option.

mike-sul (commented):
Did the first round. Looks good to in general. The key comments are:

1. the `LiteClient` shouldn't be referenced from the external API/header;
2. what about ECU registration and sending report events for secondary ECU updates? I guess this is going to be added on the next iteration?
3.

include/aklite.h
mike-sul: then what the point to expose this functionality, just send it internally whenever akliteclient decides to do so.

doanac:
> what about ECU registration and sending report events for secondary ECU updates? I guess this is going to be added on the next iteration?

yep - the cover letter notes that. i want to get this settled before i add another new aspect to it.

include/aklite.h
mike-sul: if a client decides not to send it at all, won't it break out backend logic?

include/aklite.h
mike-sul: what makes a client set the same value of a cor ID for these both methods?

include/aklite.h
doanac: no - we just wouldn't have the data.

mike-sul:
> > I think it makes sense to put the API header(s) under include/aklite/ or include/libaklite/ so a client will do #include "aklite/" and add -I //include`. This is kind of one of the patterns and iirc this is how libaktualizr does.
> 
> those are the internal headers for aktualizr. the external headers are at https://github.com/uptane/aktualizr/tree/master/include

Not sure I understand the response. In the example https://paste.debian.net/1211026/ I see `#include "aklite.h"`, normally it should be "aklite/aklite.h" or <aklite/aklite.h>, so the header file should be in `src/aklite/ ` folder.

include/aklite.h
doanac: I suppose we could put it in the CheckIn method.

include/aklite.h
doanac: I thought a bit about that. Starts to feel overkill. We wind up having to define everything twice. this makes a forward declaration to a private class. The contents of liteclient are opaque to the user. 

One unwritten decision I made while doing this - I'm not aiming for ABI compatibility. So, we could remove that in the future and clients would work properly after re-compiling.

Maybe once this API really settles down we should think about going more elaborate, but I kind of like the minimalism for now.

pattivacek (commented):
Looks pretty good to me! I'll try it out here soon.

include/aklite.h
pattivacek: At least in our use case, we'll probably always want to check for everything. It's slightly hard for me to imagine that ever not being the case, but I suppose it's plausible.

include/aklite.h
mike-sul: Removing  the LiteClient from the interface makes it even more minimalistic :). 

 >We wind up having to define everything twice
Why?

mike-sul: >Maybe once this API really settles down we should think about going more elaborate
Yes, it's not critical at this stage. We can elaborate it later. IMHO, any external API/interface should be pure virtual :), otherwise further development and support will be very hacky :).

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+class LiteClient;
+
+/**
+ * A high-level representation of a TUF Target in terms applicable to a
+ * FoundriesFactory.
+ */
+class TufTarget {
+ public:
+  TufTarget(std::string name, std::string sha256, int version, Json::Value custom)
+      : name_(std::move(name)), sha256_(std::move(sha256)), version_(version), custom_(std::move(custom)) {}
```
pattivacek, line 20: Might be nice to have some sort of "empty" Target initializer. For libaktualizr we have `static Target Unknown();`.

include/aklite.h
doanac: fixup = ccb842e4586a4c67cffbac69a80a0aff8c1f0238

include/aklite.h
doanac: ccb842e4586a4c67cffbac69a80a0aff8c1f0238

include/aklite.h
doanac: fixed with:
1824f23 (HEAD -> libaklite) fixup! api: Add Install method
af14ccd fixup! api: Add Download method
3be157e fixup! api: Add CheckIn method

include/aklite.h
doanac: 347cbcf1c80a9f47ea0ebd15ea73dcf61d676ade

include/aklite.h
doanac: 179aa36 (HEAD -> libaklite, origin/libaklite) fixup! api: Add Install method
beb11cd fixup! api: Add Download method

doanac:
force pushed a rebased/squashed patch set with the trivial stuff cleaned up

include/aklite.h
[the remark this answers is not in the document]
doanac: The most common source of friction and bugs I get when using this code base has always been how stateful objects are without you knowing. You get code where you have to know to call methodA before methodB is safe. Or having to remember to call our magic method https://github.com/foundriesio/aktualizr-lite/blob/master/src/main.cc#L325.

Setting that in TufTarget is going to lead to it getting set internally, the update working, the user doing `current = latest` and then the next update won't get a good correlation id.

One alternative I was thinking is having the DownloadResponse include a method `Install`. Then it could all be encapsulated there. The problem I'm seeing is that the main reason people want to customize our agent is because they have very unique ways they want to control the update. e.g. maybe download, wait for some time - perhaps the device or update agent would restart. and then install. I'm not sure this compromise would help them. But maybe I'm dreaming complications that don't exist - nobody has specifically asked for that.

src/CMakeLists.txt
doanac: i changed the header to `include/aktualizr-lite/api.h` so i think this name makes more sense now.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+    Failed,
+  };
+  InstallResult(Status status, std::string description = "") : status(status), description(std::move(description)) {}
+  Status status;
+  std::string description;
+};
+
+/**
+ * AkliteClient provides an easy-to-use API for users wanting to customize
+ * the behavior of aktualiz-lite.
```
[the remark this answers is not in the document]
doanac: 09f8673972a276ad03dc0a078b0cad85253dae1c

include/aklite.h
doanac: I've just force pushed an approach that consolidates SendTelemetry and ReportAkliteConfiguration into the CheckIn method:

 d2a5e23...5fd8825

I also stumbled upon a bug we have with reportHwInfo on intel:  https://github.com/foundriesio/aktualizr-lite/pull/115/commits/0232f9acf951ea939642f1f21f6a3337e6b0dd27

include/aklite.h
doanac: its been put into CheckIn.

include/aklite.h
doanac: One approach mike and I have thought for secondaries would be to embedded the secondary hash in the custom data of the primary's Target. Then you have atomic Targets and no need for this filter.

However, its probably worth the effort to allow it to filter for more than one hwid. As it stands now, its not good for either approach. I'll think about how to fix that.

include/aklite.h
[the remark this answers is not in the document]
doanac: What do you all think of something like this change:

 59726560c529483fe5288bb7f277700d4d614a08

Its kind of interesting. We really don't want people to call Install if Download hasn't been called, so this kind of forces things to be done correctly.

include/aktualizr-lite/api.h
```diff
@@ -0,0 +1,165 @@
...
+class DownloadResult {
+ public:
+  enum class Status {
+    Ok = 0,
+    DownloadFailed,
+    VerificationFailed,
+  };
+  Status status;
+  std::string description;
+  std::function<InstallResult()> Install;
```
mike-sul, line 94: That's clever.

mike-sul: btw, `correlation ID` is an attribute of Uptane::Target, as well as it's a column  in in the `installed_versions` table.

src/api.cc
```diff
@@ -0,0 +1,155 @@
...
+  Uptane::HardwareIdentifier hwid(client.config.provision.primary_ecu_hardware_id);
+  for (const auto& t : client.allTargets()) {
+    if (t.filename() == name) {
+      return std::make_shared<Uptane::Target>(t);
+    }
+  }
+  return nullptr;
+}
+
+InstallResult _install(LiteClient& client, Uptane::Target& t) {
```
mike-sul, line 100: This is the python way to name private methods, I suppose we can call it a normal way in C++ :).

mike-sul (commented):
LGTM at first glance, I am going actually to try it, build and run locally.

include/aklite.h
doanac: i've removed the hwid for now. I think when i do the next phase to support secondary ecu, we'll find a need to change this slightly.

include/aklite.h
```diff
@@ -0,0 +1,178 @@
...
+   * Install the Target
+   */
+  InstallResult Install(const TufTarget &t, const std::string &correlation_id);
+
+  /**
+   * Check if the Target has been installed but failed to boot. This would
+   * make this be considered a "rollback target" and one we shouldn't consider
+   * installing.
+   */
+  bool IsRollback(const TufTarget &t) const;
```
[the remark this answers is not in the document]
doanac: i'm not a fan of the wording either. However, we've made this such a common term when talking to people now, that I'm afraid calling it something better would just lead to more confusion.

doanac:
@mike-sul @pattivacek - this latest revision should address all the major original concerns. I'm going to spend some time testing it today on a real device.

mike-sul:
Maybe it makes sense to include the example main into the project (e.g. under test folder), I managed to built it locally but keep getting `Segmentation fault` when trying to run it.

include/aktualizr-lite/api.h
```diff
@@ -0,0 +1,170 @@
...
+    OkCached,  // check-in failed, but locally cached meta-data is still valid
+    Failed,    // check-in failed and there's no valid local meta-data
+  };
+  CheckInResult(Status status) : status(status) {}
+  Status status;
+  const std::vector<TufTarget> &Targets() const { return targets_; }
+  TufTarget GetLatest() const { return targets_.back(); }
+
+ private:
+  friend class AkliteClient;
```
mike-sul, line 64: I assume, the ctor could have additional/second param, the target  vector. Then,  CheckIn() could `move` the vector defined in its local context in return statement. As result, the `friend` can be removed.

src/api.cc
```diff
@@ -0,0 +1,149 @@
...
+    }
+    for (const auto& it : t.hardwareIds()) {
+      if (it == hwidToFind) {
+        result.targets_.emplace_back(t.filename(), t.sha256Hash(), ver, t.custom_data());
+        break;
+      }
+    }
+  }
+
+  std::sort(result.targets_.begin(), result.targets_.end(), compareTargets);
```
mike-sul, line 63: I suppose if std::set or std::unordered_set is used with a custom comparator  then no need in a separate sort call.

src/api.cc
```diff
@@ -0,0 +1,149 @@
...
+  boost::property_tree::ptree pt;
+  boost::property_tree::ini_parser::read_ini(ss, pt);
+  return pt;
+}
+
+TufTarget AkliteClient::GetCurrent() const {
+  auto current = client_->getCurrent();
+  int ver = -1;
+  try {
+    ver = std::stoi(current.custom_version(), nullptr, 0);
```
mike-sul, line 80: It looks like the try/catch and conversion could happen in the TufTarget's ctor if it's done in many places.

src/api.cc
```diff
@@ -0,0 +1,149 @@
...
+  } catch (const std::invalid_argument& exc) {
+    LOG_ERROR << "Invalid version number format: " << current.custom_version();
+  }
+  return TufTarget(current.filename(), current.sha256Hash(), ver, current.custom_data());
+}
+
+static InstallResult install(LiteClient& client, const Uptane::Target& t) {
```
mike-sul, line 87: So, we actually need the install method in the interface? Maybe, the install functor in the download result will suffice?

src/api.cc
```diff
@@ -0,0 +1,149 @@
...
+  client.logTarget("Installing: ", t);
+
+  auto rc = client.install(t);
```
mike-sul, line 90: Just wondering why we return `data::ResultCode::Numeric`  instead of `data::InstallationResult` in `LiteClient::install`, including explanation of the installation failure would be really useful. :)

src/api.cc
mike-sul: sorry, I see why it's needed now. I just confused it with `AkliteClient::Install()`.

tests/apiclient_test.cc
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Class MockAppEngine
+ *
+ */
+class MockAppEngine : public AppEngine {
+ public:
+  MockAppEngine(bool default_behaviour = true) {
+    if (!default_behaviour) return;
+
+    ON_CALL(*this, fetch).WillByDefault(Return(true));
```
mike-sul, line 37: It's not related to the given effort/PR, I just think that we can avoid copy-pasting the mock app engine definition.

mike-sul (approved):
@doanac looks great, just minor and not exactly related or important comments for the current phase.

src/api.cc
doanac: i agree, i'm just trying to not turn this PR into an effort for fixing the liteclient API. having this API in place should make changing the liteclient signature easier

include/aktualizr-lite/api.h
doanac: latest force push fixes this in the CheckIn commit.

doanac:
>Maybe it makes sense to include the example main into the project (e.g. under test folder), I managed to built it locally but keep getting Segmentation fault when trying to run it.

The latest version's final commit is an example program i'm running locally to update a device.

src/api.cc
doanac: I don't have a strong opinion. I went down a bit of a rabbit hole on pros/cons for this. sounds like std::set has some downsides but I think its when you hit a much larger size of elements than we're dealing with. The other thing I was starting to think was doing one sort at the end might be faster than doing lots of sorted-inserts. 

Let me know what you think. If you want std::set - do i have to be concerned with doing any special `==` operator handling to make that container work properly?

mike-sul: I think it's not really important  at this stage and may be in general taking into account a number of targets . 
The std::sort does object copying or movement (if it supports it). In this case, the `TufTarget` consists of std::string and Json::Value that supports "move", so the sort should be optimal in terms of object copying during sorting in the array.

doanac:
@mike-sul - its passing basic tests at home. i think its probably okay to merge. i'll leave that to you and start looking into the secondary ecu apis next.

mike-sul:
> its passing basic tests at home. i think its probably okay to merge.

sure, it works for me too. @pattivacek please, let me know if it's OK with you to merge it.

include/aklite.h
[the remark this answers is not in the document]
pattivacek: That's a pretty cool idea as far as I'm concerned.

pattivacek:
LGTM, too. I'll need to test it more but it's fine with me to merge and I'll post an issue if something isn't right.

PR #121, opened 2021-09-22
api: Include jsoncpp headers with install

doanac:
These are required by api.h


doanac:
@pattivacek - you may hit this as you build out your custom client.

mike-sul (approved):
I suppose we don't really need the given headers on a device, unless it's needed for some dev flow that includes native compilation (qemu, device) and/or bitbake/Yocto is not smart enough to include paths to the given headers during building of a client that depends on them.

pattivacek:
> @pattivacek - you may hit this as you build out your custom client.

Indeed. I solved it by hacking the installation path in libaktualizr. I actually had to do that to get it to install correctly. I did some work recently to improve the installation mechanism in the aktualizr repo, and when I rebased your changes on top of that, I saw there were a few conflicts. I think we need a better way to make the paths relative in aktualizr's CMake. Slightly annoying but if it's gonna be a subproject we have to do something like that.

> I suppose we don't really need the given headers on a device, unless it's needed for some dev flow that includes native compilation (qemu, device) and/or bitbake/Yocto is not smart enough to include paths to the given headers during building of a client that depends on them.

Right, you only need to put headers like that into the `-dev` package. I think libaktualizr is already doing that and aktualizr-lite as well.

PR #123, opened 2021-09-27
Skip App fetching if already fetched

mike-sul:
Make sure that App content (manifest and/or its archive) is (re-)fetched only if it's not stored or damaged (hash) on a file system.


doanac (approved):
this is a nice change. been wanting something like this for a long time

PR #124, opened 2021-09-28
Fix app manifest storing and verification

mike-sul:
- store a received App manifest as it is, don't apply any modification/canonization;
- calculate a manifest hash by using actual unmodified/non-canonized  representation of an App manifest json;
- minor logging improvement.


doanac (approved):
Looks good. I'd almost asked for the HEAD commit in this PR during the original review.

PR #125, opened 2021-09-28
Add Target verification step to Compose App Manager

mike-sul:
Make sure that App is verified (docker-compose config) just after successful download.
We used to do it for Compose Apps at the end of "fetch" phase.
Now, we want to support App update rollback, so we need to distinguish between true download and Target verification&installation  failures. In the first case, aklite can keep doing download attempts to make sure that Target Apps are fully downloaded. In the later case, there is no sense in update retrying, thus a rollback should be performed.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -58,6 +58,22 @@ bool RestorableAppEngine::fetch(const App& app) {
   return res;
 }
 
+bool RestorableAppEngine::verify(const App& app) {
+  bool res{false};
+  try {
+    const Uri uri{Uri::parseUri(app.uri)};
+    const auto app_dir{apps_root_ / uri.app / uri.digest.hash()};
+    TemporaryDirectory app_tmp_dir;
+    installApp(app_dir, app_tmp_dir.Path());
+    LOG_DEBUG << app.name << ": verifying App: " << app_dir << " --> " << app_tmp_dir.Path();
+    res = EXIT_SUCCESS ==
+          boost::process::system(compose_cmd_ + " config -q", boost::process::start_dir = app_tmp_dir.Path());
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "failed to verify App; app: " + app.name + "; uri: " + app.uri + "; err: " + exc.what();
+  }
+  return res;
+}
+
 bool RestorableAppEngine::install(const App& app) {
   bool res{false};
   try {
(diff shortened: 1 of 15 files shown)


mike-sul:
This PR is kind of prerequisite for "App rollback" feature.

src/docker/restorableappengine.cc
```diff
@@ -58,6 +58,22 @@ bool RestorableAppEngine::fetch(const App& app) {
...
 
+bool RestorableAppEngine::verify(const App& app) {
+  bool res{false};
+  try {
+    const Uri uri{Uri::parseUri(app.uri)};
+    const auto app_dir{apps_root_ / uri.app / uri.digest.hash()};
+    TemporaryDirectory app_tmp_dir;
+    installApp(app_dir, app_tmp_dir.Path());
+    LOG_DEBUG << app.name << ": verifying App: " << app_dir << " --> " << app_tmp_dir.Path();
+    res = EXIT_SUCCESS ==
```
doanac, line 69: any reason this doesn't call verifyComposeApp also?

doanac (commented):
I'm trying to wrap my mind around calling `docker-compose config` 3 times with 3 different implementations. Can you explain this a little more? This will cause some churn on something like an imx6 where running that command is pretty slow

src/docker/restorableappengine.cc
mike-sul: `docker-compose config` is supposed to be called once for ComposeAppEngine and twice for RestorableAppEngine.

ComposeAppEngine: just after App archive is pulled and extracted within "fetch" phase.

RestorableAppEngine
1. Within the "verify target" phase which is invoked between "fetch" and "install". Failure of `docker-compose config` is not considered as an exceptional situation, this is just part of the business logic normal flow. That's why `verifyComposeApp` is not reused since it yields an exception.

2. As part of "install" step, just after App content is extracted from the skopeo store into the compose app install dir (`/var/sota/compose-apps/<app-name>`). We just make sure that the extracted  compose file is valid. It might happen that power failure happens during the extraction, so the resultant docker-compose.yml might become broken.

I think, #2 is  optional as the following `docker-compose up` actually does verify docker-compose.yml implicitly anyway. The current approach is just more accurate and precise in terms of determining a type of error.

doanac: I think accuracy rules with this project. I am curious why item 1 from above isn't exceptional. I think for it to fail it would have to change from valid content to invalid content between two update steps, right? In other word, I'd think 2 is mandatory but 1 is optional.

mike-sul: >In other word, I'd think 2 is mandatory but 1 is optional.

Not sure I understand what do you mean exactly, let's discuss it offline :).

PR #126, opened 2021-09-28
api: Add option to assert single instance

doanac:
When clients run in a daemon mode, they'll need to assert they are the
only copy of the process running. Tools like lmp-device-register also
check for this.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -1,5 +1,6 @@
 #include "aktualizr-lite/api.h"
 
+#include <sys/file.h>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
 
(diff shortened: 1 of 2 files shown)


doanac:
@mike-sul - not super pleased with the naming or mechanics of this. let me know what you think

src/api.cc
```diff
@@ -55,6 +56,21 @@ AkliteClient::AkliteClient(const std::vector<boost::filesystem::path>& config_di
   client_->finalizeInstall();
 }
 
+void AkliteClient::AssertSingleInstance() {
+  // Leave this open for the remainder of the process to keep the lock held
+  int fd = open("/var/lock/aklite.lock", O_CREAT | O_RDONLY, 0444);
```
mike-sul, line 61: Is it require the opposite operation - unlocking?

mike-sul:
Maybe this something like this https://github.com/foundriesio/aktualizr-lite/blob/aa7dbfdeccce96dee20b35eaf526ee02422eda61/src/main.cc#L20 can be used? The idea is to use an RAII like object that is composed into `InstallContext` or just act a a scoped lock within the state changing methods like `CheckIn`, `Download` and `Install`.
As alternative the Api can just expose an utility RAII like class to lock/unlock. Then, a client can decides by itself when to lock and lock since it might be use-case specific logic.

src/api.cc
doanac: no. a flock dies when the file is closed (or process exits).

doanac:
> The idea is to use an RAII like object that is composed into InstallContext or just act a a scoped lock within the state changing methods like CheckIn, Download and Install.
As alternative the Api can just expose an utility RAII like class to lock/unlock. Then, a client can decides by itself when to lock and lock since it might be use-case specific logic.

I started with that approach. You could pretty much make the CheckIn and Installer methods obtain a lock. The problem I thought about is that you could wind up with 2 daemons running at the same time in different intervals. After looking at how we use it in daemon_main and lmp-device-register looks for it, I started to realize its not really used as a critical-section/mutex primitive - it's used to ensure one copy of a daemon is running (or not running in the case of lmp-device-register).

mike-sul:
It looks to me as rather use-case specific stuff and I am not sure it should be part of the API at all. Maybe just put the lock implementation into the example and that's it.

Or, actually do it, but hide from a client, in the api implementation internals. In this case the ctor should be extended with additional param to denote whether it's RW (update) or just R (info) instance, iirc the flock/fd api allows specifying different types of locks, rw, r.

I incline towards the first approach, just reference implementation in the example.

doanac:
> It looks to me as rather use-case specific stuff and I am not sure it should be part of the API at all. 

making lmp-device-register depend on it via the configurable $SOTA_CLIENT variable implies some contractual obligations in order to make things work correctly.

I think hiding it via the constructor makes sense. It's really an implementation and system dependency that's bubbling up. It's causing real friction in my factory now though.

doanac:
I've force pushed a different approach. I'm not sure this is the one, but maybe it helps move us towards the right one.

mike-sul:
> I've force pushed a different approach. I'm not sure this is the one, but maybe it helps move us towards the right one.

the direction makes sense to me.

doanac:
@mike-sul - do we tweak this more or merge it?

mike-sul:
> do we tweak this more or merge it?

It's up to you and Patti, I think it's good enough for the current iteration.

mike-sul (approved):
Just rebase on master before merging.

doanac:
rebased on master

PR #127, opened 2021-09-28
Create a new "aklite server"

doanac:
This patch set allows aktualizr-lite to run as a REST API service exposed via a Unix domain socket. It's a similar approach to something like the docker daemon. This would allow users to create custom sota clients in higher level languages like Python and not have to use the C++ client API.

### DIFF
--- a/src/server-main.cc
+++ b/src/server-main.cc
@@ -60,6 +60,23 @@ static void check_in(const AkliteClient& client, const httplib::Request& req, ht
   json_resp(res, code, data);
 }
 
+static void check_apps_in_sync(const AkliteClient& client, const httplib::Request& req, httplib::Response& res,
+                               CurrentInstaller* cur_installer) {
+  (void)req;
+  LOG_DEBUG << "check_apps_in_sync called";
+
+  auto installer = client.CheckAppsInSync();
+  if (installer == nullptr) {
+    json_resp(res, 204, Json::Value());
+  } else {
+    cur_installer->id++;
+    cur_installer->installer = std::move(installer);
+    Json::Value data;
+    data["installer-id"] = cur_installer->id;
+    json_resp(res, 200, data);
+  }
+}
+
 static void create_installer(const AkliteClient& client, const httplib::Request& req, httplib::Response& res,
                              CurrentInstaller* cur_installer) {
   (void)req;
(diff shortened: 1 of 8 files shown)


doanac:
@mike-sul - this is still a little rough, but it's far enough along to start getting high level feedback on the approach.

mike-sul:
I wonder what clang-tidy and valgrind thinks about it :).

doanac:
CI should be happy now.

cajun-rat:
Http as the RPC interface makes a lot of sense. It might be a good idea to make the API expressible in swagger/openapi: this would make it easy to generate stubs to call the API from statically typed languages (Golang, etc).

Did you investigate [boost::beast ](https://www.boost.org/doc/libs/1_77_0/libs/beast/doc/html/index.html) as the http server? 

Pros:

- We already have a boost dependency, so get it for free
- Maybe possible to use it for backend duties too?  At some point I would like to extend uptane_generator to support basic self-hosted use cases. I don't know the relative maturity of httplib and boost::beast.

Cons

- boost::beast is relatively new. The first release was 1.66, and 1.67 is in Debian oldstable, but I think there have been some breaking API changes. If we are not going back that far or intend to put this behind a compile flag, then this probably isn't a problem.
- Lower level API. There is no `svr.Get("/targets/current",...` in boost::beast

cajun-rat:
@doanac I just realised this on the ak-lite not Aktualizr repo. I'll leave my comment here for posterity, but it pretty much nonsense :)

src/server-main.cc
```diff
@@ -0,0 +1,336 @@
...
+        Json::Value data;
+        data["error"] = e.what();
+        json_resp(res, 500, data);
+      }
+    });
+
+    std::mutex client_mutex;
+    CurrentInstaller installer{0, nullptr};
+
+    svr.Get("/check_in", [&client, &client_mutex](const httplib::Request& req, httplib::Response& res) {
```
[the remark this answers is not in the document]
mike-sul: Maybe it should return just an array/list , no need in `targest` element in the response json.

src/server-main.cc
```diff
@@ -0,0 +1,336 @@
...
+    });
+    svr.Get("/config", [&client, &client_mutex](const httplib::Request& req, httplib::Response& res) {
+      std::lock_guard<std::mutex> guard(client_mutex);
+      get_config(client, req, res);
+    });
+    svr.Get("/targets/current", [&client, &client_mutex](const httplib::Request& req, httplib::Response& res) {
+      std::lock_guard<std::mutex> guard(client_mutex);
+      get_current_target(client, req, res);
+    });
+    svr.Get("/targets/rollback/(\\S+)", [&client, &client_mutex](const httplib::Request& req, httplib::Response& res) {
```
mike-sul, line 299: I am not sure I get this interface?

src/server-main.cc
```diff
@@ -0,0 +1,336 @@
...
+      std::lock_guard<std::mutex> guard(client_mutex);
+      get_rollback_target(client, req, res);
+    });
+    svr.Post("/targets/installer",
```
mike-sul, line 303: Maybe `HTTP POST /targets/<target-name>/installers`?

mike-sul:
@doanac I think it makes sense to close this PR and just leave the branch?

PR #129, opened 2021-10-01
Forward-declare as much as possible in liteclient.h.

pattivacek:
This makes it much easier to use elsewhere. Note that the destructor has
to be declared explicitly to prevent auto-inlining, which prevents
forward declaration with unique_ptr.

There were several transitive header dependencies that had to be filled
in appropriately as well.

This _might_ depend on https://github.com/uptane/aktualizr/pull/19, but I think it's just our test code that depends on both.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -5,8 +5,10 @@
 #include <boost/uuid/uuid_io.hpp>
 
 #include "helpers.h"
+#include "http/httpclient.h"
 #include "libaktualizr/config.h"
 #include "liteclient.h"
+#include "primary/reportqueue.h"
 
 std::vector<boost::filesystem::path> AkliteClient::CONFIG_DIRS = {"/usr/lib/sota/conf.d", "/var/sota/sota.toml",
                                                                   "/etc/sota/conf.d/"};
--- a/src/helpers.cc
+++ b/src/helpers.cc
@@ -4,6 +4,7 @@
 #include <boost/uuid/uuid_io.hpp>
 
 #include "composeappmanager.h"
+#include "storage/invstorage.h"
 
 static bool appListChanged(const Json::Value& target_apps, std::vector<std::string>& cfg_apps_in,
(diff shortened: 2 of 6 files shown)


pattivacek:
Annoyingly, this is still failing for me locally when I try to build the tests:
```
/usr/bin/ld: tests/CMakeFiles/t_liteclientHSM.dir/liteclientHSM_test.cc.o: in function `P11EngineGuard::P11EngineGuard(P11Config const&)':
/home/patti/Code/aktualizr-lite/build/../aktualizr/src/libaktualizr/crypto/p11engine.h:85: undefined reference to `P11Engine::P11Engine(P11Config)'
```
I'm guessing this is because I'm building with P11? I'm still trying to make sense of it.

mike-sul (approved):

mike-sul:
> I'm guessing this is because I'm building with P11?

What target do you build, I suppose `client-example`?

pattivacek:
> What target do you build, I suppose `client-example`?

`ninja build_tests`

mike-sul:
> > What target do you build, I suppose `client-example`?
> 
> `ninja build_tests`

hm, both `build_tests` and `custom-client` targets works for me, with P11 on `cmake -S . -B build-ninja -DCMAKE_BUILD_TYPE=Debug -GNinja -DBUILD_P11=ON`

pattivacek:
> I'm guessing this is because I'm building with P11? I'm still trying to make sense of it.

Oops, that was supposed to read "building WITHOUT P11".

mike-sul:
> > I'm guessing this is because I'm building with P11? I'm still trying to make sense of it.
> 
> Oops, that was supposed to read "building WITHOUT P11".

`t_liteclientHSM` should have been included conditionally. I'll fix it shortly.

mike-sul:
https://github.com/foundriesio/aktualizr-lite/pull/130

pattivacek:
Rebased. Can I get a merge if there's no objections?

PR #131, opened 2021-10-05
Updates to make custom-client build inside Yocto

doanac:
These changes make it easier to build the custom-client as-is inside yocto.


examples/custom-client-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required (VERSION 3.5)
+project(custom-sota-client)
+
+add_executable(custom-sota-client main.cc)
```
mike-sul, line 4: I just wonder why it cannot be another target (custom-sota-client) in the scope of the existing project (aklite).

unit-test
```diff
@@ -12,23 +12,36 @@ set -x
...
+echo "## Running aklite unit tests"
 cd $build
 ctest -j6 -V -L aklite
+ninja install
 cd $here
 rm -rf $build
 
+echo "## Compiling the custom-client-cxx example"
+build=$(mktemp -d)
+cp -r $here/aktualizr/third_party/jsoncpp/include/json /usr/include
```
mike-sul, line 37: It means that this script (unit-test) can be executed only as a root and within a container.

doanac: I thought that was already a requirement because the unit test runs wrapdocker - https://github.com/foundriesio/aktualizr-lite/blob/master/dev-shell#L24

mike-sul (approved):

PR #133, opened 2021-11-04
Check if App is really installed successfully

mike-sul:
This change makes sure that we properly determine whether a new App was successfully installed or not.
From aklite (update agent) perspective an App installation is successful if:
1.  `docker-compose up` exits with 0;
2.  All App's containers are created.
    
So, the PR makes sure the second criterion is met.

App/service running inside of Compose App's container(s)  might fail of course, too. But aklite doesn't take it into account during App installation since this is App specific business logic failure and we wanna keep aklite agnostic of it.
Thus, in this case, a user will be notified that the new App version was successfully installed, but the status report will indicate that one or more of the App's containers are not running including its state and exit/failure code&reason.

### DIFF
--- a/src/docker/composeappengine.cc
+++ b/src/docker/composeappengine.cc
@@ -66,7 +66,7 @@ bool ComposeAppEngine::install(const App& app) {
     switch (state()) {
       case AppState::State::kPulled:
       case AppState::State::kInstallFail: {
-        if (!installApp(app)) {
+        if (!installApp(app) || !areContainersCreated(app)) {
           state.setState(AppState::State::kInstallFail);
           break;
         }
@@ -82,7 +82,7 @@ bool ComposeAppEngine::install(const App& app) {
   } catch (const std::exception& exc) {
     LOG_WARNING << "Failed to get/set App state, fallback to forced install: " << exc.what();
 
-    if (installApp(app)) {
+    if (installApp(app) && areContainersCreated(app)) {
       result = true;
     }
   }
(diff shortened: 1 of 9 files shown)


src/docker/composeappengine.cc
```diff
@@ -322,6 +324,37 @@ bool ComposeAppEngine::start(const App& app) {
...
 }
 
+bool ComposeAppEngine::areContainersCreated(const App& app) {
+  bool result{false};
+  try {
+    ComposeInfo info((appRoot(app) / ComposeFile).string());
+    std::vector<Json::Value> services = info.getServices();
+    if (services.empty()) {
+      LOG_WARNING << "App: " << app.name << ", no services in docker file!";
+      return false;
```
doanac, line 334: maybe we need defensive code here, but shouldn't this really go into `verify` so that we can reject an app with no services?

doanac (commented):
nothing big. 

one thing to clarify to make sure I understand this.

When `docker-compose up` exits, containers will be *created* but probably not *running*. Thus you have logic for created and running. correct?

mike-sul:
> When `docker-compose up` exits, containers will be _created_ but probably not _running_. Thus you have logic for created and running. correct?

yes.

src/docker/composeappengine.cc
mike-sul: Yep, this is rather to make "defensive code".
It looks like something that not exactly related to the given PR, although makes sense to be improved. On the other side, our CI should prevent from such cases.
Anyway, I think, I am fine with adding a check for service number to the verification step.

mike-sul: I think it makes sense to address it in another PR because I see that we process the target verification failure incorrectly and there is not test for the given case, so all this together makes it quite a big change.

doanac: k. lets do LOG_ERROR here and address that in another PR.

mike-sul: force pushed it.

doanac (approved):

PR #135, opened 2021-11-10
api.h: Add constructor with commandline argument map.

pattivacek:


pattivacek:
Really weird/annoying how the formatter requires `Config &config` in headers but `Config& config` in the source files. I think it's the same way in aktualizr. Why does it do that?

doanac:
> Why does it do that?

I'd always assumed because "that's how Patti liked it". Now I know we've all disliked that :)

doanac (approved):

pattivacek:
> I'd always assumed because "that's how Patti liked it". Now I know we've all disliked that :)

I remember noticing that once before and thinking "that's odd" but I assumed I was misunderstanding something or there was a good reason I hadn't heard of yet. If anyone knows the story or can figure out how to change it to be consistent, I'd love to know it!

PR #143, opened 2021-12-03
apps: check expected size of a layer manifest

mike-sul:
We do not know in advance a size of an App's index/root manifest,
hence we just assume that it shouldn't be bigger than certain predefined value (16KB).
But, we know a layer manifest size since it is defined in the root level manifest (Merkle tree).
Therefore, we must ensure that a downloaded layer manifest's size matches the size specified in the root manifest.
Also, we must sure that data being downloaded don't exceed the given manifest's size,
libcurl does that we just pass a corresponding param down to it via the http client.

### DIFF
--- a/src/docker/docker.cc
+++ b/src/docker/docker.cc
@@ -65,21 +65,33 @@ RegistryClient::RegistryClient(std::shared_ptr<HttpInterface> ota_lite_client, s
       ota_lite_client_{std::move(ota_lite_client)},
       http_client_factory_{std::move(http_client_factory)} {}
 
-std::string RegistryClient::getAppManifest(const Uri& uri, const std::string& format) const {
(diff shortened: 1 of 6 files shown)


mike-sul:
@doanac It still doesn't work exactly as I wanted, so no need to review it yet.

mike-sul:
One of things that I wanted to implement is rather a subject for a dedicated ticket and PR (mainly because of challenges in testing), https://foundriesio.atlassian.net/browse/FFTK-1071. So, I will address it in another PR.

mike-sul:
@doanac I think, it's good to go as it is. Two other concerns that I have are subject of another tasks/PRs.
https://foundriesio.atlassian.net/browse/FFTK-1072

doanac (approved):

PR #144, opened 2021-12-06
apps: parameterized App recreation before reboot

mike-sul:
- Add a config param `create_containers_before_reboot` that allows
configuring aklite behavior if both ostree and Apps are updated.
Specifically, whether to recreate App's containers or not before a system reboot.
By default, App's containers are recreated (but not restarted).
It's worth noting that App containers are recreated just after reboot if
docker-compose detects any configuration changes in any case
(e.g. a new version of ostree/LmP modifies compose App content on boot).

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -305,8 +305,16 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
     // I have no idea via the package manager interface method install() is const which is not a const
     // method by its definition/nature
     auto& non_const_app_engine = (const_cast<ComposeAppManager*>(this))->app_engine_;
-    auto run_res = just_install ? non_const_app_engine->install({pair.first, pair.second})
-                                : non_const_app_engine->run({pair.first, pair.second});
+    bool run_res{false};
+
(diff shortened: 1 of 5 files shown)


mike-sul:
@doanac I added the tests I planned to add.

src/composeappmanager.cc
```diff
@@ -299,22 +303,32 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
...
+
+      if (!run_res) {
+        res = data::InstallationResult(data::ResultCode::Numeric::kInstallFailed, "Could not install app");
+      } else {
+        res.description += "\n" + pair.second;
+      }
     }
-  };
+  } else {
+    LOG_INFO << "Apps' containers will be re-created and started just after succesfull boot on the new ostree version";
```
doanac, line 325: should we add this into res.description to provide more install details to our backend?

mike-sul: makes sense, will do it.

mike-sul: updated.

doanac (approved):

PR #145, opened 2021-12-10
Complete check whether App is fetched

mike-sul:
- Traverse through each element of App's Merkle tree in order to verify whether it's completely fetched or not.
- Tests to verify the procedure that checks whether App is entirely fetched.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -242,7 +242,14 @@ void RestorableAppEngine::prune(const Apps& app_shortlist) {
         const auto image_root{app_dir / app_version_dir / "images" / image_uri.registryHostname / image_uri.repo /
                               image_uri.digest.hash()};
 
-        const auto image_manifest_desc{Utils::parseJSONFile(image_root / "index.json")};
+        const auto index_manifest{image_root / "index.json"};
+        if (!boost::filesystem::exists(index_manifest)) {
+          LOG_WARNING << "Failed to find an index manifest of App image: " << image << ", removing its directory";
+          boost::filesystem::remove_all(image_root);
+          continue;
+        }
+
+        const auto image_manifest_desc{Utils::parseJSONFile(index_manifest)};
         HashedDigest image_digest{image_manifest_desc["manifests"][0]["digest"].asString()};
         blob_shortlist.emplace(image_digest.hash());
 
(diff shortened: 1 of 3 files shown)


src/docker/restorableappengine.cc
```diff
@@ -517,15 +524,89 @@ bool RestorableAppEngine::isAppFetched(const App& app) const {
...
+                << "; index: " << index_manifest;
+      return false;
+    }
+
+    const auto manifest_desc{Utils::parseJSONFile(index_manifest)};
+    HashedDigest manifest_digest{manifest_desc["manifests"][0]["digest"].asString()};
+
+    const auto manifest_file{blobs_root_ / "sha256" / manifest_digest.hash()};
+    if (!boost::filesystem::exists(manifest_file)) {
+      LOG_DEBUG << app.name << ": missing App image manifest; image: " << image << "; manifest: " << manifest_file;
```
doanac, line 575: is this really LOG_DEBUG - almost feels like a LOG_WARN

src/docker/restorableappengine.cc
```diff
@@ -576,32 +576,59 @@ bool RestorableAppEngine::areAppImagesFetched(const App& app) const {
...
+    const auto config_hash{getContentHash(config_file)};
+    if (config_hash != config_digest.hash()) {
+      LOG_DEBUG << app.name << ": App image config hash mismatch; actual: " << config_hash
+                << "; expected: " << config_digest.hash();
+      return false;
+    }
 
-    // add layer blobs
+    // check layers, just check blobs' size since generation of their hashes might consumes
+    // too much CPU for a given device ???
```
doanac, line 605: size is probably fine cause docker is going to verify the hashes when they get imported, right?

doanac (commented):
looks fine. just some dumb questions

src/docker/restorableappengine.cc
mike-sul: I don't have strong opinion about it. I put LOG_DEBUG because it might be just a regular/normal case if a download is interrupted.

src/docker/restorableappengine.cc
mike-sul: Unfortunately, `skopeo copy` just fails if one of the existing blobs has incorrect hash. That's why I have to remove the blob size for which  is incorrect (#623).
So, the issue might happen if a size matches while hash is not, to prevent it we need to patch skopeo.

PR #146, opened 2021-12-10
AkliteClient: Unlock the lock before destructing.

pattivacek:


src/api.cc
```diff
@@ -89,6 +90,16 @@ AkliteClient::AkliteClient(const boost::program_options::variables_map& cmdline_
...
 }
 
+AkliteClient::~AkliteClient() {
+  // Release the lock to allow reobtaining with another instance.
+  int fd = open("/var/lock/aklite.lock", O_CREAT | O_RDONLY, 0444);
+  if (fd >= 0) {
+    flock(fd, LOCK_UN);
+    close(fd);
+    unlink("/var/lock/aklite.lock");
+  }
```
doanac, line 100: I bet this will all work by just calling `unlink("/var/lock/aklite.lock");` the active process isn't dropping the original flock, its just causing us to flock a new file/inode

pattivacek: Updated; I think you're right. I'll test shortly.

mike-sul (approved):

mike-sul:
@pattivacek LGTM, please, just make clang tidy happy.

pattivacek:
> just make clang tidy happy.

FYI I recommend clang-tidy-12 now. It looks like the default is still 10 here.

pattivacek:
Okay, CI passes now!

PR #150, opened 2022-01-10
Fix aktualizr secondary root

pattivacek:
Based on https://github.com/foundriesio/aktualizr-lite/pull/146 and including https://github.com/uptane/aktualizr/pull/50. Also includes https://github.com/uptane/aktualizr/pull/48; are there particularly changes required for supporting that? It wasn't clear to me yet, and my testing so far has worked fine, but the comments seem to indicate some adaptation is required.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "aktualizr"]
 	path = aktualizr
 	url = https://github.com/foundriesio/aktualizr
-	branch = 2021.11+fio
+	branch = 2022.1+fio


mike-sul:
> are there particularly changes required for supporting that?

"It works for our case if alt_remote is set what is not always the case because we do fallback to the config.ostree_server if optional/alternative remotes are not available. If it happens then the given code will set target.custom.uri as the remote URI which doesn't work for us because we use this field custom.uri for referring to the CI build that generated a given target."

pattivacek:
I think I need some help understanding what's going on with the failing test.

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "aktualizr"]
 	path = aktualizr
-	url = https://github.com/foundriesio/aktualizr
```
mike-sul, line 3: I think I need to create a new branch (2022.1+fio) in foundriesio/aktualizr for this change.

mike-sul: I think it has to do wit this change https://github.com/uptane/aktualizr/commit/0b49e03970ef4cd20d36f0ff7388b15cf39b0313.
Let me make corresponding/adjusting changes in aklite.

pattivacek: > I think I need to create a new branch (2022.1+fio) in foundriesio/aktualizr for this change.

I was wondering about that. Is that something I can do if I want to push changes like this, or does it have to be done by y'all?

> I think it has to do wit this change [uptane/aktualizr@0b49e03](https://github.com/uptane/aktualizr/commit/0b49e03970ef4cd20d36f0ff7388b15cf39b0313).

That's what I both feared and hoped.

> Let me make corresponding/adjusting changes in aklite.

Thanks! Let me know if I can help.

mike-sul: > I was wondering about that. Is that something I can do if I want to push changes like this, or does it have to be done by y'all?

@doanac What do you think about it?

mike-sul: @pattivacek You can refer to this branch https://github.com/foundriesio/aktualizr/tree/2022.1%2Bfio now.

mike-sul:
This change https://github.com/uptane/aktualizr/commit/7cdf9b4c080a1aedf74e420d4f6fae2f327e9d32 might cause an issue too because `if (target.MatchTarget(Uptane::Target::Unknown()))` never works if sql.db is absent or empty.

pattivacek:
> if sql.db is absent or empty

That shouldn't really ever happen, though, right? Still, seems like something perhaps @cajun-rat might want to know about.

mike-sul:
> That shouldn't really ever happen, though, right?

I meant `installed_versions`  record is not set in sql.db which is quite possible just after an initial system boot.
I think @cajun-rat did the right thing, I am just afraid that the fixed issue became a feature over the time :) and there is some code that expects getCurrent() to be Target::Unknown if neither "current_version" nor "installed_versions" matches the currently  deployed ostree hash.

.gitmodules
doanac: I don't mind if patti wants to push mergeups to our aktualizr  fork. Patti is more in tuned with what's happening there anyway.

mike-sul: I suppose we need to grant @pattivacek with permissions to create PRs to foundriesio/aktualizr, unless it's already set.

cajun-rat:
> > That shouldn't really ever happen, though, right?
> 
> I meant `installed_versions` record is not set in sql.db which is quite possible just after an initial system boot. I think @cajun-rat did the right thing, I am just afraid that the fixed issue became a feature over the time :) and there is some code that expects getCurrent() to be Target::Unknown if neither "current_version" nor "installed_versions" matches the currently deployed ostree hash.

Oops, I missed this when making those changes. I was pretty convinced that the only usage of getCurrent() was in reporting the Uptane manifest back to the server, so the change was safe in that the new hash was 'more correct' than the old hash.

When you test and find `getCurrent() == Target::Unknown`, what happens next? (i.e. where does that information flow to?)

pattivacek:
Looks like my changes were already included in #151, so I'm closing this. Thanks again!

mike-sul:
> When you test and find `getCurrent() == Target::Unknown`, what happens next? (i.e. where does that information flow to?)

It's used in two spots within aklite. In one place, it's just impact log message for one of the rarely used use-case https://github.com/foundriesio/aktualizr-lite/blob/a923fdb291e22911dd3e5b622a8b09aff81bd173/src/main.cc#L77.
In the second case it might impact customer unless they implement the callback handler fully correct https://github.com/foundriesio/aktualizr-lite/blob/a923fdb291e22911dd3e5b622a8b09aff81bd173/src/liteclient.cc#L175.

PR #152, opened 2022-01-11
Check an available capacity of the docker store before an App update

mike-sul:
Check whether there is sufficient amount of available storage in both the skopeo and docker store before an App update.
A storage size, required to accommodate a new App version, is calculated  precisely for the skopeo store (tar.gzip format) and approximately for the docker store.
The approximation is based on an average gzip compression ratio in the case of rootfs diffs/image layers.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -1,5 +1,6 @@
 #include "restorableappengine.h"
 
+#include <sys/statvfs.h>
 #include <limits>
 #include <unordered_set>
 
@@ -29,12 +30,15 @@ RestorableAppEngine::StorageSpaceFunc RestorableAppEngine::DefStorageSpaceFunc =
     };
 
 RestorableAppEngine::RestorableAppEngine(boost::filesystem::path store_root, boost::filesystem::path install_root,
+                                         boost::filesystem::path docker_root,
                                          Docker::RegistryClient::Ptr registry_client,
                                          Docker::DockerClient::Ptr docker_client, std::string client,
                                          std::string docker_host, std::string compose_cmd,
                                          StorageSpaceFunc storage_space_func)
     : store_root_{std::move(store_root)},
       install_root_{std::move(install_root)},
+      docker_root_{std::move(docker_root)},
+      docker_and_skopeo_same_volume_{areDockerAndSkopeoOnTheSameVolume(store_root_, docker_root_)},
       client_{std::move(client)},
       docker_host_{std::move(docker_host)},
       compose_cmd_{std::move(compose_cmd)},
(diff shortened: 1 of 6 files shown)


mike-sul:
@doanac This is my idea behind estimation of a required storage space within the docker store, i.e. approximate the required for App update storage space based on the following assumptions:
1. Layers in the skopeo and docker store are in-sync.
2. A layer compression ratio is constant across all layers.

src/docker/restorableappengine.cc
```diff
@@ -754,4 +790,32 @@ std::string RestorableAppEngine::getContentHash(const boost::filesystem::path& p
...
+  }
+  const auto dockerVolumeID{getPathVolumeID(docker_path.parent_path())};
+  if (!std::get<1>(dockerVolumeID)) {
+    LOG_WARNING << "Failed to obtain an ID of a docker store volume; path: " << docker_path
+                << ", err: " << strerror(errno);
+  }
+  // we assume that a docker data root and a skopeo store are located on the same volume in case of a failure to obtain
+  // a volume ID
+  const auto res{std::get<0>(skopeoVolumeID) == std::get<0>(dockerVolumeID)};
+  return res;
```
doanac, line 808: Is there some optimization or reason you don't just do:
```
return std::get<0>(skopeoVolumeID) == std::get<0>(dockerVolumeID);
```

src/docker/restorableappengine.cc
```diff
@@ -754,4 +790,32 @@ std::string RestorableAppEngine::getContentHash(const boost::filesystem::path& p
...
+  if (!std::get<1>(skopeoVolumeID)) {
+    LOG_WARNING << "Failed to obtain an ID of a skopeo store volume; path: " << skopeo_path
+                << ", err: " << strerror(errno);
+  }
+  const auto dockerVolumeID{getPathVolumeID(docker_path.parent_path())};
+  if (!std::get<1>(dockerVolumeID)) {
+    LOG_WARNING << "Failed to obtain an ID of a docker store volume; path: " << docker_path
+                << ", err: " << strerror(errno);
+  }
+  // we assume that a docker data root and a skopeo store are located on the same volume in case of a failure to obtain
```
doanac, line 805: would it be safer to just `return false` right after the LOG_WARNING's above? then you aren't relying on the failed return value to behave a certain way.

src/docker/restorableappengine.cc
mike-sul: it's just convenient when debugging.

src/docker/restorableappengine.cc
mike-sul: By default we assume that the stores are located on the same volume. If it fails for both we need it to be `true`. If for one of it it should be `false`.

doanac: ah - i see the `getPathVolumeID` controls this.

mike-sul:
@doanac I think it's good enough now for a final review.

doanac (approved):
nice changes from the initial draft

PR #153, opened 2022-01-12
Fix app rollback in the case of App update failure

mike-sul:


mike-sul:
@doanac This fixes an App rollback if App installation fails during "only App" update. The rollback itself was implemented long time ago, but it turned out that it didn't work quite properly because a bad/failing "only App update" Target wasn't treated as a failing Target so aklite kept trying to install it. This change fixes the logic  that determines whether Target is "bad" or "good" thus an overall App rollback works now. I added corresponding test as well as verified it from e2e perspective on a device.

doanac (approved):
interesting how "simple" that turned out to be. 

I think these logging changes are nice change as well.

PR #154, opened 2022-01-14
fixup! use generic multiplication with overflow function

mike-sul:
Use the generic/typeless multiplication with overflow checking function instead of the type specific option. The latter creates a challenge to make both clang-tidy and different gcc toolchains (archs) happy due to mess with integer typing in C++ world.


mike-sul:
@pattivacek 
I tested the fix for.
```
MACHINE              = "imx7ulpea-ucom"
TUNE_FEATURES        = "arm vfp cortexa7 neon thumb callconvention-hard"
```

doanac (approved):

PR #157, opened 2022-01-27
aktualizr: allow updating meta if the same ver

mike-sul:
Allow updating metadata if local and remote versions are the same
but they actually differ.
We need it for the use-case of switching tags on a device what may lead
to a need to update metadata of the same version.
Each tag has its own set of metadata (except root) so metadata
of different tags can have the same version but they actually differ
because they refer to different set of targets.

### DIFF
--- a/examples/custom-client-cxx/main.cc
+++ b/examples/custom-client-cxx/main.cc
@@ -71,6 +71,11 @@ int main(int argc, char **argv) {
       if (latest.Name() != current.Name() && !client.IsRollback(latest)) {
         std::string reason = "Updating from " + current.Name() + " to " + latest.Name();
         auto installer = client.Installer(latest, reason);
+        if (!installer) {
+          LOG_ERROR << "Found latest Target but failed to retrieve its metadata from DB, skipping update";
+          std::this_thread::sleep_for(std::chrono::seconds(interval));
+          continue;
+        }
         auto dres = installer->Download();
         if (dres.status != DownloadResult::Status::Ok) {
           LOG_ERROR << "Unable to download target: " << dres;
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "aktualizr"]
 	path = aktualizr
 	url = https://github.com/foundriesio/aktualizr
-	branch = 2022.1+fio
+	branch = 2022.2+fio
--- a/tests/apiclient_test.cc
+++ b/tests/apiclient_test.cc
@@ -206,6 +206,93 @@ TEST_F(ApiClientTest, Secondaries) {
   ASSERT_EQ(secondary_target.filename(), result.GetLatest("riscv").Name());
 }
 
+TEST_F(ApiClientTest, SwitchTag) {
+  auto liteclient = createLiteClient();
+  ASSERT_TRUE(targetsMatch(liteclient->getCurrent(), getInitialTarget()));
+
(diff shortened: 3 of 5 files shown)


doanac:
wow. that's subtle. It fixes the issue, so I'm fine with the change. Are we doing anything against the spirit of TUF here though?

pattivacek:
> Are we doing anything against the spirit of TUF here though?

I don't think so, TUF and Uptane seem to allow for metadata to change without bumping the version number. I was always uncomfortable with that, but I'm not aware of a security risk as a result of that.

mike-sul:
@pattivacek @vkhoroz I added a test for the given case.
@doanac I made tiny change to the custom/example client so it doesn't crash in the use-case this PR provides fix for.

mike-sul:
> > Are we doing anything against the spirit of TUF here though?
> 
> I don't think so, TUF and Uptane seem to allow for metadata to change without bumping the version number. I was always uncomfortable with that, but I'm not aware of a security risk as a result of that.

The spec says:
```
The version number of the trusted timestamp metadata file, if any, MUST be less than or equal to the version number of the new timestamp metadata file. If the new timestamp metadata file is older than the trusted timestamp metadata file, discard it, abort the update cycle, and report the potential rollback attack.
```
So, updating meta if `local_version == remote_version` does not violate the spec.

mike-sul:
There is one tiny and not critical spec violation, and it doesn't  relate to this PR.

The snapshot version check should happen before timestamp persisting and snapshot meta downloading, while we do this check after these steps.
```
5. The version number of the snapshot metadata file in the trusted timestamp metadata file, if any, MUST be less than or equal to its version number in the new timestamp metadata file. If not, discard the new timestamp metadata file, abort the update cycle, and report the failure.
```

examples/custom-client-cxx/main.cc
```diff
@@ -71,6 +71,11 @@ int main(int argc, char **argv) {
       if (latest.Name() != current.Name() && !client.IsRollback(latest)) {
         std::string reason = "Updating from " + current.Name() + " to " + latest.Name();
         auto installer = client.Installer(latest, reason);
+        if (!installer) {
+          LOG_ERROR << "Found latest Target but failed to retrieve its metadata from DB, skipping update";
+          std::this_thread::sleep_for(std::chrono::seconds(interval));
+          continue;
+        }
```
doanac, line 78: nice catch. thanks

doanac (approved):

vkhoroz (approved):

pattivacek:
> There is one tiny and not critical spec violation, and it doesn't relate to this PR.
> 
> The snapshot version check should happen before timestamp persisting and snapshot meta downloading, while we do this check after these steps.
> 
> ```
> 5. The version number of the snapshot metadata file in the trusted timestamp metadata file, if any, MUST be less than or equal to its version number in the new timestamp metadata file. If not, discard the new timestamp metadata file, abort the update cycle, and report the failure.
> ```

Yeah, fair point. Kind of a pain to get right, and I'm not sure there's any security risk with the current approach. I'm quite open to reconsidering that even if there isn't.

mike-sul:
> I'm not sure there's any security risk with the current approach. I'm quite open to reconsidering that even if there isn't.

I'm not sure neither. I think it can be changed in a scope of "making verifyXXXX()" effort,  as well as loading snapshot/targets metadata into the object context at its instantiation so `local_version` is not -1 just after reboot.

PR #158, opened 2022-01-31
ci-retest aklite: enforce ostree update if hashes differ

mike-sul:
Enforce an ostree update if current and the latest targets have different ostree hashes even if their versions match.

This is needed for the use-case when the same OE CI run is re-run one or more times and each time a new ostree and target are pushed to the backend.


doanac (approved):
thanks!

mike-sul:
ci-retest

PR #159, opened 2022-02-01
Change the target download interface

mike-sul:
This PR goal is twofold, report download error messages and make the first step towards getting rid of dependency on the package manager interface.

- Introduce a new interface of a target download that uses the api's download result type.
- Make the package managers implement the new interface.
- Change LiteClient so it uses the new download interface and includes a failure reason into a report event sent to the backend.
- Adjust LiteClient clients so they use the api's download result type.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -2,6 +2,8 @@
 
 #include <set>
 
+#include <boost/format.hpp>
+
 #include "bootloader/bootloaderlite.h"
 #include "docker/restorableappengine.h"
 #include "target.h"
@@ -210,8 +212,12 @@ DownloadResult ComposeAppManager::Download(const TufTarget& target) {
 
   for (const auto& pair : all_apps_to_fetch) {
     LOG_INFO << "Fetching " << pair.first << " -> " << pair.second;
-    if (!app_engine_->fetch({pair.first, pair.second})) {
-      res = {DownloadResult::Status::DownloadFailed, "Download failed TODO"};
+    const auto fetch_res{app_engine_->fetch({pair.first, pair.second})};
+    if (!fetch_res) {
+      const std::string errDesc{boost::str(boost::format("failed to fetch App; app: %s; uri: %s; err: %s") %
+                                           pair.first % pair.second % fetch_res.err)};
+      res = {DownloadResult::Status::DownloadFailed, errDesc};
+      break;
     }
   }
 
@@ -221,12 +227,13 @@ DownloadResult ComposeAppManager::Download(const TufTarget& target) {
 
 bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetcher& fetcher, const KeyManager& keys,
                                     const FetcherProgressCb& progress_cb, const api::FlowControlToken* token) {
+  (void)target;
   (void)fetcher;
   (void)token;
   (void)progress_cb;
   (void)keys;
 
-  return Download(Target::toTufTarget(target));
+  throw std::runtime_error("Using obsolete method of package manager: fetchTarget()");
 }
 
 TargetStatus ComposeAppManager::verifyTarget(const Uptane::Target& target) const {
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -214,9 +214,10 @@ DownloadResult ComposeAppManager::Download(const TufTarget& target) {
     LOG_INFO << "Fetching " << pair.first << " -> " << pair.second;
     const auto fetch_res{app_engine_->fetch({pair.first, pair.second})};
     if (!fetch_res) {
-      const std::string errDesc{boost::str(boost::format("failed to fetch App; app: %s; uri: %s; err: %s") %
-                                           pair.first % pair.second % fetch_res.err)};
-      res = {DownloadResult::Status::DownloadFailed, errDesc};
+      const std::string err_desc{boost::str(boost::format("failed to fetch App; app: %s; uri: %s; err: %s") %
+                                            pair.first % pair.second % fetch_res.err)};
+      LOG_ERROR << err_desc;
+      res = {DownloadResult::Status::DownloadFailed, err_desc};
       break;
     }
   }
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -75,8 +75,9 @@ ComposeAppManager::Config::Config(const PackageConfig& pconfig) {
 ComposeAppManager::ComposeAppManager(const PackageConfig& pconfig, const BootloaderConfig& bconfig,
                                      const std::shared_ptr<INvStorage>& storage,
                                      const std::shared_ptr<HttpInterface>& http,
-                                     std::shared_ptr<OSTree::Sysroot> sysroot, AppEngine::Ptr app_engine)
-    : RootfsTreeManager(pconfig, bconfig, storage, http, std::move(sysroot)),
+                                     std::shared_ptr<OSTree::Sysroot> sysroot, const KeyManager& keys,
+                                     AppEngine::Ptr app_engine)
+    : RootfsTreeManager(pconfig, bconfig, storage, http, std::move(sysroot), keys),
       cfg_{pconfig},
       app_engine_{std::move(app_engine)} {
   if (!app_engine_) {
@@ -182,25 +183,27 @@ bool ComposeAppManager::checkForAppsToUpdate(const Uptane::Target& target) {
   return cur_apps_to_fetch_and_update_.empty() && cur_apps_to_fetch_.empty();
 }
 
-bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetcher& fetcher, const KeyManager& keys,
-                                    const FetcherProgressCb& progress_cb, const api::FlowControlToken* token) {
-  if (!RootfsTreeManager::fetchTarget(target, fetcher, keys, progress_cb, token)) {
-    return false;
+DownloadResult ComposeAppManager::Download(const TufTarget& target) {
+  auto ostree_download_res{RootfsTreeManager::Download(target)};
+  if (!ostree_download_res) {
+    return ostree_download_res;
   }
+
+  DownloadResult res{DownloadResult::Status::Ok, ""};
+  const Uptane::Target uptane_target{Target::fromTufTarget(target)};
 
   if (cfg_.force_update) {
     LOG_INFO << "All Apps are forced to be updated...";
-    cur_apps_to_fetch_and_update_ = getApps(target);
+    cur_apps_to_fetch_and_update_ = getApps(uptane_target);
   } else if (!are_apps_checked_) {
     // non-daemon mode (force check) or a new Target to be applied in daemon mode,
     // then do full check if Target Apps are installed and running
     LOG_INFO << "Checking for Apps to be installed or updated...";
-    checkForAppsToUpdate(target);
+    checkForAppsToUpdate(uptane_target);
   }
 
   LOG_INFO << "Found " << cur_apps_to_fetch_and_update_.size() << " Apps to update";
 
-  bool passed = true;
   AppsContainer all_apps_to_fetch;
   all_apps_to_fetch.insert(cur_apps_to_fetch_and_update_.begin(), cur_apps_to_fetch_and_update_.end());
   all_apps_to_fetch.insert(cur_apps_to_fetch_.begin(), cur_apps_to_fetch_.end());
@@ -208,12 +211,22 @@ bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetche
   for (const auto& pair : all_apps_to_fetch) {
     LOG_INFO << "Fetching " << pair.first << " -> " << pair.second;
     if (!app_engine_->fetch({pair.first, pair.second})) {
-      passed = false;
+      res = {DownloadResult::Status::DownloadFailed, "Download failed TODO"};
     }
   }
 
   are_apps_checked_ = false;
-  return passed;
+  return res;
+}
+
+bool ComposeAppManager::fetchTarget(const Uptane::Target& target, Uptane::Fetcher& fetcher, const KeyManager& keys,
+                                    const FetcherProgressCb& progress_cb, const api::FlowControlToken* token) {
+  (void)fetcher;
+  (void)token;
+  (void)progress_cb;
+  (void)keys;
+
+  return Download(Target::toTufTarget(target));
 }
 
 TargetStatus ComposeAppManager::verifyTarget(const Uptane::Target& target) const {
--- a/src/docker/composeappengine.cc
+++ b/src/docker/composeappengine.cc
@@ -14,10 +14,10 @@ ComposeAppEngine::ComposeAppEngine(boost::filesystem::path root_dir, std::string
       docker_client_{std::move(docker_client)},
       registry_client_{std::move(registry_client)} {}
 
-bool ComposeAppEngine::fetch(const App& app) {
+AppEngine::Result ComposeAppEngine::fetch(const App& app) {
   boost::filesystem::create_directories(appRoot(app) / MetaDir);
 
-  bool result = false;
+  Result result{false};
 
   try {
     AppState state(app, appRoot(app), true);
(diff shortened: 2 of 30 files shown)


mike-sul:
@doanac Please, look at this PR and check if you agree with an approach I took. In addition to the current changes, I 'd like to add catching stderr logs produced by `docker-compose`/`skopeo` and adding them into the report events. The most useful err logs are produced by these tools, not doing it makes this effort a bit useless, imho.

src/downloader.h
```diff
@@ -0,0 +1,20 @@
+#ifndef AKTUALIZR_LITE_DOWNLOADER_H_
+#define AKTUALIZR_LITE_DOWNLOADER_H_
+
+#include "aktualizr-lite/api.h"
+
+class Downloader {
+ public:
+  virtual DownloadResult Download(const TufTarget& target) = 0;
```
doanac, line 8: I'm guessing there are some pros/cons for using TufTarget here instead of Uptane::Target here. I get this worry we may go through a phase of code where we have different call-sites having to convert between the two. Can you explain the rationale and if you have an idea for a boundary between the two?

src/composeappmanager.cc
```diff
@@ -210,8 +212,12 @@ DownloadResult ComposeAppManager::Download(const TufTarget& target) {
...
   for (const auto& pair : all_apps_to_fetch) {
     LOG_INFO << "Fetching " << pair.first << " -> " << pair.second;
-    if (!app_engine_->fetch({pair.first, pair.second})) {
-      res = {DownloadResult::Status::DownloadFailed, "Download failed TODO"};
+    const auto fetch_res{app_engine_->fetch({pair.first, pair.second})};
+    if (!fetch_res) {
+      const std::string errDesc{boost::str(boost::format("failed to fetch App; app: %s; uri: %s; err: %s") %
+                                           pair.first % pair.second % fetch_res.err)};
+      res = {DownloadResult::Status::DownloadFailed, errDesc};
+      // Do we want to continue with Apps fetching after a failure to fetch one of the Apps?
```
doanac, line 220: probably isn't a bad idea. one problem with this logic - if there are >1 errors, you overwrite the previous `res.message`

src/docker/composeappengine.cc
```diff
@@ -42,11 +42,7 @@ bool ComposeAppEngine::fetch(const App& app) {
 
     result = true;
   } catch (const std::exception& exc) {
-    LOG_WARNING << "Failed to get/set App state, fallback to forced fetch: " << exc.what();
-
-    if (download(app) && verify(app) && pullImages(app)) {
-      result = true;
-    }
```
doanac, line 49: not sure i follow why this logic would change?

src/docker/composeappengine.cc
```diff
@@ -42,11 +42,7 @@ bool ComposeAppEngine::fetch(const App& app) {
```
doanac, line 42: for all the code above this like
```
if (!download(app)) {
      state.setState(AppState::State::kDownloadFailed);
      return result;
}
```
Should we set `result.err` to something more descriptive?

doanac (commented):
i like where this is headed

doanac:
>  I 'd like to add catching stderr logs produced by docker-compose/skopeo and adding them into the report events. The most useful err logs are produced by these tools, not doing it makes this effort a bit useless, imho.

I'd wanted to do the same thing when i introduced api.h, but quickly realized that all this plumbing was required. Thanks!

src/downloader.h
mike-sul: The only reason why I use TufTarget because we've has a plan to move towards using the api for all clients, specifically, we planned to move aklite's main.cc to the api usage?.
If it's so, then eventually it will just one component that does the conversion - LiteClient, interface of which will have just TufTarget, the conversion will be encapsulated in LiteClient implementation.

src/composeappmanager.cc
mike-sul: yes, we can concatenate error messages. I am just not sure it's worth trying to fetch other Apps if one of them fails, especially if we consider Target as atomic.

src/docker/composeappengine.cc
mike-sul: The overall logic hasn't changed as result of the AppState::setState() implementation change, it doesn't throw exception now, so the net effect of both implementations is the same.

This "state" logic of composeappengine is really weird I wish I hadn't implemented it at all, it's just optimizes a number fetch tries but makes code more complicated :).

src/docker/composeappengine.cc
mike-sul: download() should return just void since it throws an error if any error. For verify() and pullImages() we really don't know what's the error reason since we don't capture stderr of `docker-compose` process yet. Once, I implement the stderr capturing I'll update this code correspondingly. Either all these three methods will throw exception or return something like tuple<bool, errMsg>.

src/composeappmanager.cc
doanac: that's fine be me. odds are if the first one fails, the 2nd one will also.

src/downloader.h
mike-sul: Yep, it makes sense. The reason I did so is that I thought that Downloader ifc would become a part of the API some day  so a user can define/inject its own  downloader :). But, I think, I am kind of projecting the experience I got at HERE, their API is more flexible and allows their clients many different use-cases (e.g. custom download, download and installation are decoupled).
Anyway, I think you are right, it doesn't make sense to do it now, so the conversion will be encapsulated in liteclient.*.

src/composeappmanager.cc
mike-sul: just added `break` if the app fetch fails.

src/downloader.h
mike-sul: Actually I have second thoughts on this matter, I think the current implementation is a good trade-off as for an initial step towards minimizing and encapsulating dependency on aktualizr in general and not using the package manager interface at all in particular.
I can imagine many implementation options so intermediate steps towards the goal, so let's talk about it verbally to avoid confusion and assumptions about each other thoughts.

src/exec.h
```diff
@@ -0,0 +1,45 @@
...
+
+    io_service.run();
+
+    bool wait_successfull = child_process.wait_for(std::chrono::seconds(900));
+    if (!wait_successfull) {
+      throw std::runtime_error("Timeout occured while waiting for a child process completion");
+    }
+
+  } catch (const std::exception& exc) {
+    throw std::runtime_error("Failed to spawn process " + cmd + " exited with an error: " + exc.what());
```
doanac, line 29: i don't guess its possible for there to be stderr here, but I think there might be for line 25 above? so maybe we should try and include it in the exception?

src/docker/composeappengine.cc
```diff
@@ -22,22 +24,13 @@ AppEngine::Result ComposeAppEngine::fetch(const App& app) {
   try {
     AppState state(app, appRoot(app), true);
 
-    if (!download(app)) {
-      state.setState(AppState::State::kDownloadFailed);
-      return result;
-    }
+    download(app);
```
doanac, line 27: this is hard for me to reason about. we used to set the state fail to kDownloadFailed and now we aren't. I assume this means setting that file to failed states never meant anything useful? i.e we only look at it for good states?

mike-sul: Yes, those *Failed states are useless, I'll remove them.

mike-sul:
> Also - this commit feels like its making functional changes as well (how the state file gets handle) - maybe those could be split up?

well, I'll try to avoid functional changes, those that I made are driven by the switch to `exec` function that yields an exception in contrast to the previous one that returned a boolean type and no any exception.

doanac:
> What kind of unit test do you have in mind, verifying whether the report event includes detailed field?

I was thinking just a couple of tests for `exec` function itself where we look at `ex.what` to see if it collects things correctly. maybe i'm overthinking it, but this type function is always deceptively difficult to correctly.

mike-sul:
> a couple of tests for `exec` function itself where we look at `ex.what` to see if it collects things correctly

makes sense, I'll add two types of tests, for `exec` itself and checking if the status report event includes an error message.

mike-sul:
Added tests for the `exec` and stderr catching.

mike-sul:
@doanac I added unit tests as well as tested it e2e on a device.
The backend gets something like the following

```
2022-02-07T08:46:16+00:00 : EcuDownloadStarted(intel-corei7-64-lmp-1235)
 Details:
 | Sync active Target Apps
2022-02-07T08:46:42+00:00 : EcuDownloadCompleted(intel-corei7-64-lmp-1235) -> Failed!
 Details:
 | failed to fetch App; app: app-07; uri: hub.foundries.io/msul-dev01/app-07@sha256:cccc3475501792d4eeb043b728c9a0c8417fbe9f62146625610377e11bcf572d; err: failed to pull image
 | 	cmd: /home/mike/work/foundries/projects/skopeo/bin/skopeo copy -f v2s2 --dest-shared-blob-dir /home/mike/work/foundries/factories/msul-dev01/local-devices/msul-local-i7-001/reset-apps/blobs docker://hub.foundries.io/msul-dev01/nginx@sha256:67960643d8f50607a1df975581d29921bde4991e2b97bf3bc0c00b4fddd7b05b oci:/home/mike/work/foundries/factories/msul-dev01/local-devices/msul-local-i7-001/reset-apps/apps/app-07/cccc3475501792d4eeb043b728c9a0c8417fbe9f62146625610377e11bcf572d/images/hub.foundries.io/msul-dev01/nginx/67960643d8f50607a1df975581d29921bde4991e2b97bf3bc0c00b4fddd7b05b
 | 	err: time="2022-02-07T10:46:42+02:00" level=fatal msg="initializing source docker://hub.foundries.io/msul-dev01/nginx@sha256:67960643d8f50607a1df975581d29921bde4991e2b97bf3bc0c00b4fddd7b05b: reading manifest sha256:67960643d8f50607a1df975581d29921bde4991e2b97bf3bc0c00b4fddd7b05b in hub.foundries.io/msul-dev01/nginx: errors:\ndenied: requested access to the resource is denied\nunauthorized: authentication required\n"

```

I am just thinking whether we need to limit/strip an overall error message at aklite if it gets too big, e.g. 

```
2022-02-07T08:35:27+00:00 : EcuDownloadStarted(raspberrypi4-64-lmp-1235)
 Details:
 | Sync active Target Apps
2022-02-07T08:35:50+00:00 : EcuDownloadCompleted(raspberrypi4-64-lmp-1235) -> Failed!
 Details:
 | failed to fetch App; app: app-07; uri: hub.foundries.io/msul-dev01/app-07@sha256:cccc3475501792d4eeb043b728c9a0c8417fbe9f62146625610377e11bcf572d; err: failed to pull App images
 | 	cmd: /usr/bin/docker-compose pull --no-parallel
 | 	err: Pulling nginx-01 (hub.foundries.io/msul-dev01/nginx@sha256:67960643d8f50607a1df975581d29921bde4991e2b97bf3bc0c00b4fddd7b05b)...
 | /usr/bin/docker-credential-fio-helper: line 22: /usr/bin/aktualizr-get: No such file or directory
 | Traceback (most recent call last):
 |   File "/usr/lib/python3.9/site-packages/docker/credentials/store.py", line 79, in _execute
 |     output = subprocess.check_output(
 |   File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
 |     return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
 |   File "/usr/lib/python3.9/subprocess.py", line 528, in run
 |     raise CalledProcessError(retcode, process.args,
 | subprocess.CalledProcessError: Command '['/usr/bin/docker-credential-fio-helper', 'get']' returned non-zero exit status 127.
 | 
 | During handling of the above exception, another exception occurred:
 | 
 | Traceback (most recent call last):
 |   File "/usr/lib/python3.9/site-packages/docker/auth.py", line 264, in _resolve_authconfig_credstore
 |     data = store.get(registry)
 |   File "/usr/lib/python3.9/site-packages/docker/credentials/store.py", line 35, in get
 |     data = self._execute('get', server)
 |   File "/usr/lib/python3.9/site-packages/docker/credentials/store.py", line 93, in _execute
 |     raise errors.process_store_error(e, self.program)
 | docker.credentials.errors.StoreError: Credentials store docker-credential-fio-helper exited with "".
 | 
 | During handling of the above exception, another exception occurred:
 | 
 | Traceback (most recent call last):
 |   File "/usr/bin/docker-compose", line 12, in <module>
 |     sys.exit(main())
 |   File "/usr/lib/python3.9/site-packages/compose/cli/main.py", line 81, in main
 |     command_func()
 |   File "/usr/lib/python3.9/site-packages/compose/cli/main.py", line 203, in perform_command
 |     handler(command, command_options)
 |   File "/usr/lib/python3.9/site-packages/compose/metrics/decorator.py", line 18, in wrapper
 |     result = fn(*args, **kwargs)
 |   File "/usr/lib/python3.9/site-packages/compose/cli/main.py", line 844, in pull
 |     self.project.pull(
 |   File "/usr/lib/python3.9/site-packages/compose/project.py", line 763, in pull
 |     service.pull(ignore_pull_failures, silent=silent)
 |   File "/usr/lib/python3.9/site-packages/compose/service.py", line 1251, in pull
 |     return progress_stream.get_digest_from_pull(event_stream)
 |   File "/usr/lib/python3.9/site-packages/compose/progress_stream.py", line 99, in get_digest_from_pull
 |     for event in events:
 |   File "/usr/lib/python3.9/site-packages/compose/service.py", line 1218, in _do_pull
 |     output = self.client.pull(repo, **pull_kwargs)
 |   File "/usr/lib/python3.9/site-packages/docker/api/image.py", line 411, in pull
 |     header = auth.get_config_header(self, registry)
 |   File "/usr/lib/python3.9/site-packages/docker/auth.py", line 47, in get_config_header
 |     authcfg = resolve_authconfig(
 |   File "/usr/lib/python3.9/site-packages/docker/auth.py", line 324, in resolve_authconfig
 |     return authconfig.resolve_authconfig(registry)
 |   File "/usr/lib/python3.9/site-packages/docker/auth.py", line 235, in resolve_authconfig
 |     cfg = self._resolve_authconfig_credstore(registry, store_name)
 |   File "/usr/lib/python3.9/site-packages/docker/auth.py", line 280, in _resolve_authconfig_credstore
 |     raise errors.DockerException(
 | docker.errors.DockerException: Credentials store error: StoreError('Credentials store docker-credential-fio-helper exited with "".')
 | 

```

src/docker/composeappengine.cc
mike-sul: Removed the failed states.

src/downloader.h
mike-sul: Sure, eventually the conversion will be encapsulated in implementation of the interface(s) that abstract LiteClient from aktualizr.

doanac:
> I am just thinking whether we need to limit/strip an overall error message at aklite if it gets too big, e.g.

I've been on the fence with that as well. We should probably loop in @vkhoroz to see how it might affect backend management of device-events. However, as general rule of thumb, keeping this limited to some arbitrary limit like 256 or 512bytes seems reasonable and still informative enough to help an operator

vkhoroz:
> > I am just thinking whether we need to limit/strip an overall error message at aklite if it gets too big, e.g.
> 
> I've been on the fence with that as well. We should probably loop in @vkhoroz to see how it might affect backend management of device-events. However, as general rule of thumb, keeping this limited to some arbitrary limit like 256 or 512bytes seems reasonable and still informative enough to help an operator

Yes, I think we should limit it somehow, and I have some plans how to enforce these limits on the back-end side too.
So it's nice to be proactive on the client side even before this gots limited on back-end.

I have no opinion about the exact limit value and will leave it up to @mike-sul who has the full grasp on the aklite specifics.
Probably, we should also think about this from the UI perspective to prevent a really long message from clobbering the page.

doanac (approved):
this is good progress. thanks

mike-sul:
> Probably, we should also think about this from the UI perspective to prevent a really long message from clobbering the page.

I've just checked the UI and looks like it handles long messages well.

mike-sul:
> I have no opinion about the exact limit value and will leave it up to @mike-sul who has the full grasp on the aklite specifics.

I think 1024 bytes should suffice.

doanac:
> I think 1024 bytes should suffice.

This may go down in history as your Bill Gates moment :)

PR #163, opened 2022-02-28
ostree: follow redirect while getting remote URLs

mike-sul:

### DIFF
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -60,7 +60,7 @@ void RootfsTreeManager::installNotify(const Uptane::Target& target) {
 }
 
 void RootfsTreeManager::getAdditionalRemotes(std::vector<Remote>& remotes, const std::string& target_name) {
-  const auto resp = http_client_->post(gateway_url_ + "/download-urls", Json::Value::null);
+  const auto resp = http_client_->post(gateway_url_ + "/download-urls", Json::Value::null, true);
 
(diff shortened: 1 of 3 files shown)


src/rootfstreemanager.cc
```diff
@@ -60,7 +60,7 @@ void RootfsTreeManager::installNotify(const Uptane::Target& target) {
 }
 
 void RootfsTreeManager::getAdditionalRemotes(std::vector<Remote>& remotes, const std::string& target_name) {
-  const auto resp = http_client_->post(gateway_url_ + "/download-urls", Json::Value::null);
+  const auto resp = http_client_->post(gateway_url_ + "/download-urls", Json::Value::null, true);
```
pattivacek, line 63: Why not make following redirects the default? I think we already have a limit in place for the number of redirects. What's the risk?

mike-sul: TBH, I don't know. I just wanted to make sure that I won't break anything else. 
Let's try and enable it by default. The question is: Should it be default for any request (set in ctor) or default specifically for POST?

pattivacek: I'd say in can be in the ctor, as long as the setting to limit the number of redirects is also set there. I think redirects are generally safe, as long as there aren't thousands of them.

mike-sul:
Closing this PR in favor of https://github.com/foundriesio/aktualizr-lite/pull/164

PR #166, opened 2022-04-15
apps: Use `docker compose` by default

mike-sul:
Switch from the python based `docker-compose` to golang based
`docker compose` owned by docker org.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -106,7 +106,10 @@ ComposeAppManager::ComposeAppManager(const PackageConfig& pconfig, const Bootloa
   if (!app_engine_) {
     auto docker_client{std::make_shared<Docker::DockerClient>()};
     auto registry_client{std::make_shared<Docker::RegistryClient>(http, cfg_.hub_auth_creds_endpoint)};
-    const auto compose_cmd{boost::filesystem::canonical(cfg_.compose_bin).string() + " "};
+    std::string compose_cmd{boost::filesystem::canonical(cfg_.compose_bin).string() + " "};
+    if (cfg_.compose_bin.filename().compare("docker") == 0) {
+      compose_cmd += "compose ";  // "make it `docker compose` command
+    }
     const std::string skopeo_cmd{boost::filesystem::canonical(cfg_.skopeo_bin).string()};
     const std::string docker_host{"unix:///var/run/docker.sock"};
 
--- a/unit-test
+++ b/unit-test
@@ -5,7 +5,7 @@ cd $here
 
 if [ $# -eq 0 ] ; then
 	echo "Launching unit tests in docker"
-	TEST=1 exec ./dev-shell ./unit-test indocker
+	TEST=1 exec ./dev-shell sudo -u testuser ./unit-test indocker
 fi
 
 set -x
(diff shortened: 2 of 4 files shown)


mike-sul:
[retest]

src/composeappmanager.cc
```diff
@@ -106,7 +106,10 @@ ComposeAppManager::ComposeAppManager(const PackageConfig& pconfig, const Bootloa
   if (!app_engine_) {
     auto docker_client{std::make_shared<Docker::DockerClient>()};
     auto registry_client{std::make_shared<Docker::RegistryClient>(http, cfg_.hub_auth_creds_endpoint)};
-    const auto compose_cmd{boost::filesystem::canonical(cfg_.compose_bin).string() + " "};
+    std::string compose_cmd{boost::filesystem::canonical(cfg_.compose_bin).string() + " "};
+    if (cfg_.compose_bin.filename().compare("docker") == 0) {
+      compose_cmd += "compose ";  // "make it `docker compose` command
+    }
```
doanac, line 112: this will make it impossible to run a new aklite with an old platform. Its probably the easier fix, but we might also leave this as is and instead of check for "file exists" at startup, we check if `$cmd -h` works.

@ricardosalveti - you think we'll ever have someone wanting to run this and not also wanting to update their platform?

mike-sul: Why impossible? a user still can use `docker-compose` by specifying it as a value of `docker_compose_bin` in his local env or additional toml snippet in meta-sub (as we agreed in the slack discussion).

doanac: n/m i missed the "+ compose" part.

ricardosalveti: Only for folks in dunfell that might not necessarily have docker compose v2 available, but as @mike-sul said, we can override the default via toml.

mike-sul:
[ci-retest]

mike-sul:
[ci-retest]

mike-sul:
@doanac I ran into really odd issue, see https://ci.foundries.io/projects/aklite/builds/691/, https://ci.foundries.io/projects/aklite/builds/692/ as an example.
```
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
108:   what():  boost::filesystem::remove: Operation not permitted: "/tmp/aktualizr-999c-5974-55dc-158b/1b1e-ad74-dir/sysrepo/ostree/deploy/lmp/deploy/70e41f0dc0868a786529bfe531c3972c8565ae4da74fc7733c0c772a1ca8b977.0/boot"
``` 
It took some time for me to fix it. I wonder how the tests work before at all.
In any case, it's better to run them as a non root user and it looks like the run becomes faster after the fix :).

unit-test
```diff
@@ -5,7 +5,7 @@ cd $here
 
 if [ $# -eq 0 ] ; then
 	echo "Launching unit tests in docker"
-	TEST=1 exec ./dev-shell ./unit-test indocker
+	TEST=1 exec ./dev-shell sudo -u testuser ./unit-test indocker
```
doanac, line 8: weird, but it works, so ship it.

mike-sul: Yes, but I managed to consistently reproduce it locally. Something must have changed recently, maybe in docker for ubuntu.

I need at least one approval to merge it.

doanac (approved):

PR #168, opened 2022-04-25
No space reaction

mike-sul:
- Identify an insufficient space error if ostree or restorable App pull fails;
- Don't retry the pull if such the error is detected;
- Don't try new updates until aklite is restarted (or device is rebooted).

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -378,13 +378,19 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
           LOG_INFO << "Latest Target: " << target_to_install.filename() << " is a failing Target (aka known locally)."
                    << " Skipping its installation.";
         }
+        data::ResultCode::Numeric rc{data::ResultCode::Numeric::kOk};
         if (!client.appsInSync()) {
           client.checkForUpdatesEnd(target_to_install);
-          do_app_sync(client);
+          rc = do_app_sync(client);
+          if (rc == data::ResultCode::Numeric::kOk) {
+            LOG_INFO << "Device is up-to-date";
+          } else {
+            LOG_ERROR << "Failed to sync the current Target: " << target_to_install.filename();
+          }
         } else {
+          LOG_INFO << "Device is up-to-date";
           client.checkForUpdatesEnd(Uptane::Target::Unknown());
         }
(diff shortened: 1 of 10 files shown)


src/main.cc
```diff
@@ -310,6 +316,13 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
         continue;
       }
 
+      if (no_space) {
+        LOG_WARNING
+            << "Insufficient storage space was detected, skipping any updates until aktualizr-lite is restarted";
+        std::this_thread::sleep_for(std::chrono::seconds(interval));
+        continue;
```
doanac, line 323: I'm not sure we should just stop retrying. Nobody is going to have logic to restart the daemon. Is there some way we could try and detect if free disk space had increased, so that we could start trying to install again?

mike-sul: >Nobody is going to have logic to restart the daemon.

Well, I am not sure about it. Aeler kind of does it, their App does control the aklite daemon. In general, I think customers should use aklite via the API/lib or command line tools. In this case it's easier to have control over upadte, e.g. let a device user to decide when is the best time to perform an update or notify a user about insufficient space so he/she has a chance to clean a storage.


> Is there some way we could try and detect if free disk space had increased, so that we could start trying to install again?

I've thought about. I think there are many ways how to do that, but I am not that it belongs to aklite. This is up to an application that uses aklite to decide how to react on such situation, hot to clean storage. if it is so then that application can start aklite again.

In any case, the key goal of this change is to overcome the issue that Milosh discovered.

We may allow an update again when a new Target appears.

mike-sul: @mike-scott @Tim-Anderson What do you think aklite should do in this case (not enough space to download update),

1. Don't keep trying to download an update until aklite is restarted or a device is rebooted.

2. Check whether available storage has increased since "not enough space" was detected and try to do an update again then.

Tim-Anderson:
I would suggest #2

mike-sul:
@doanac I removed the last commit. So, now aklite will keep trying to update, but only once per update cycle and won't do download try for each ostree remote/origin/bucket just for the first one.

doanac (approved):

mike-sul:
[ci-retest]

PR #169, opened 2022-04-27
Check rootfs free space before next download attempt after "no space" failure

mike-sul:
- Store currently available free space if an ostree download fails because insufficient available storage.
- Check whether free space has increased before the next ostree download attempt. Skip the download if a free space has not been increased, unless a new Target to be downloaded.

### DIFF
--- a/src/ostree/repo.cc
+++ b/src/ostree/repo.cc
@@ -146,4 +146,53 @@ void Repo::checkout(const std::string& commit_hash, const std::string& src_dir, 
   }
 }
 
+void Repo::setConfigItem(const std::string& section, const std::string& key, const std::string& value) {
+  g_autoptr(GError) error = nullptr;
+  g_autoptr(GKeyFile) config = nullptr;
+
+  config = ostree_repo_copy_config(repo_);
+  if (config == nullptr) {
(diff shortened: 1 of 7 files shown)


mike-sul:
That's not easy to test this new functionality because free storage changes all the time while tests are running in parallel.

mike-sul:
@doanac I've just disabled the test that verifies the new functionality if free space has increased. The problem is that storage space changes al the time during running many tests in parallel.
I've considered and experimented with an approach that allows to "inject" the free space function to rootfs manager. It would provide the test ability to fake actual storage size and have full control over it.
But, it requires too many changes to the code that is part of a usual flow, not sure it's worth doing it.

mike-sul:
> but this previous value would still be there and reject the OTA. Is that possible?

Not exactly, If a new Target arrives then the pacman will try to download it unconditionally for the first time.
The stored size is a target specific.

mike-sul:
I suppose it's fine to implement it in the scope of the daemon/main.cc if

1. we are OK that the API clients will handle this case by themselves;
2. we are OK that aklite will retry download after restart/reboot so we keep the state just in the daemon runtime context/RAM without persisting it on storage.

PR #172, opened 2022-05-04
Offline update PoC

mike-sul:
Dirty PoC of offline update support at aklite aimed to get learn all subtle issues and/or roadblocks towards a proper implementation.


mike-sul:
@doanac This dirty PoC is good enough for me to get an idea how the overall offline update will work. I am going to put some doc to outline and discuss e2e solution.

PR #173, opened 2022-05-05
apps: Fallback to python's compose if no docker's

mike-sul:


src/composeappmanager.cc
```diff
@@ -110,6 +110,12 @@ ComposeAppManager::ComposeAppManager(const PackageConfig& pconfig, const Bootloa
     if (cfg_.compose_bin.filename().compare("docker") == 0) {
       compose_cmd += "compose ";  // "make it `docker compose` command
     }
+    std::string out;
+    if (Utils::shell(compose_cmd, &out, true) != EXIT_SUCCESS) {
```
doanac, line 114: I think we can be more clear by not providing a default in composemanager.h. Then in this file we dynamically decide what the default is in order of preference(golang then python). So sort of:
```
if (raw.count("docker_compose_bin") == 0) {
    if exists("/usr/lib/docker/cli-plugins/docker-compose") {
      we use golang version
   } else if exists("/usr/bin/docker-compose") {
     we use python version
   } else {
     this is really bad
   }
}
```

mike-sul: Also, how aklite can know that `docker compose` is installed in `"/usr/lib/docker/cli-plugins/docker-compose"` and the same for `docker-compose`. For example, in my local setup I use different locations.

So, I suggest to not make assumption about actual location of `docker` and `docker-compose`.

doanac (approved):

mike-sul:
So we get the following with this version against the latest meta-lmp:

```
docker: 'compose' is not a docker command.
See 'docker --help'
warning: The docker utility specified in the config does not support `compose` mode: /usr/bin/docker compose 
warning: Falling back to the python docker-compose: /usr/bin/docker-compose.python3-docker-compose 
  debug: Compose utility: `/usr/bin/docker-compose.python3-docker-compose `

```

PR #176, opened 2022-06-07
App early startup utility

mike-sul:
The utility is aimed for running preloaded Restorable Apps before device registration and networking is available.

- The "hack" sub-command adds the preloaded Apps to a docker store repository, so dockerd doesn't need to contact Registry.
- The run sub-command injects the preloaded Apps' images into docker store and starts them.


apps/fioapp/hack_cmd.h
```diff
@@ -0,0 +1,48 @@
...
+#include "cmd.h"
+
+#include <boost/algorithm/string/classification.hpp>
+#include <boost/algorithm/string/split.hpp>
+#include "logging/logging.h"
+
+namespace apps {
+namespace fioapp {
+
+extern int hackDockerStore(const std::vector<std::string>& shortlist, const std::string& store_root,
```
doanac, line 13: should we put this in a header to avoid `extern` stuff that might get brittle?

doanac (commented):
i have the most annoying review comment ever - I'm not sure we should use the word "fioapp". That's *really* generic. I'd almost think something called fioapp was a reliability and serviceability tool we included or our "app store".

apps/fioapp/hack_cmd.h
mike-sul: I just didn't want to add a whole new file for sake of a single function :).

doanac:
maybe `compose-apps-early`. then you get `compose-apps-early hack|start` and it sort of fits with the wrapper script? Or `aklite-apps-early`?

doanac:
> I am fine with it, not sure I like early in the name though.

that's seems reasonable.

mike-sul:
So, `aklite-apps`?

doanac (approved):
i like how this has changed. looks good

mike-sul:
I added a commit to store an app uri https://github.com/foundriesio/aktualizr-lite/pull/176/commits/5e942653f944222c838bb41cb006c16c2eff7d2f

Now, it's ready to be merged.

PR #178, opened 2022-06-24
apps: stop removed Apps before creating new one

mike-sul:
Apps that have been removed from a config or are not present in a new
Target should be stopped before creating/starting new Apps because they
may interfere with each other, e.g. the same port is re-used.

- Add a new method to AppEngine to stop App.
- Stop removed Apps before creating and/or starting new Apps.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -335,6 +335,9 @@ data::InstallationResult ComposeAppManager::install(const Uptane::Target& target
   const bool just_install = res.result_code == data::ResultCode::Numeric::kNeedCompletion;  // system update
 
   if (!just_install || cfg_.create_containers_before_reboot) {
+    // Stop disabled Apps before creating or starting new Apps since they may interfere with each other (e.g. the same
+    // port is used).
+    stopDisabledComposeApps(target);
     // make sure we install what we fecthed
     if (!cur_apps_to_fetch_and_update_.empty()) {
       res.description += "\n# Apps installed:";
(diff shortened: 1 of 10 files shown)


mike-sul:
> this looks almost identical to pr 177, did you mean to create this ?

Yes, 177 is PR to the v86 branch, this is the fix's cherry-pick to the current head of master.

src/composeappmanager.cc
```diff
@@ -456,6 +462,34 @@ void ComposeAppManager::handleRemovedApps(const Uptane::Target& target) const {
...
+        auto& non_const_app_engine = (const_cast<ComposeAppManager*>(this))->app_engine_;
+        non_const_app_engine->stop({name, ""});
+      }
+    }
+  }
+}
+
 // Handle the case like:
 //  1) sota.toml is configured with 2 compose apps: "app1, app2"
 //  2) update is applied, so we are now running both app1 and app2
```
doanac, line 495: we now have `removeDisabledComposeApps` and `stopDisabledComposeApps`. They almost do the same thing. Do we need both?

src/docker/composeappengine.cc
```diff
@@ -151,6 +151,15 @@ AppEngine::Result ComposeAppEngine::run(const App& app) {
...
 }
 
+void ComposeAppEngine::stop(const App& app) {
+  try {
+    const auto root_dir{appRoot(app)};
+    exec(compose_ + "down", "failed to bring App down", boost::process::start_dir = root_dir);
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "docker-compose was unable to bring down: " << exc.what();
+  }
+}
```
doanac, line 161: this is 99% the same as `remove`. You can remove most of `remove`'s logic and just call this now.

src/composeappmanager.cc
mike-sul: Yes, we need both.

mike-sul: I brought it to a common method.

src/docker/composeappengine.cc
mike-sul: It doesn't make sense to reuse `stop` from `remove` because we need to put try/catch into `remove` anyway (remove_all can throw exc). So, what's the point to run `this->stop()` if we can run `exec(compose down)` directly.

PR #183, opened 2022-07-18
aklite: Import pre-provisioned root meta

mike-sul:
Import pre-provisioned root role metadata from a file system into the
aklite's DB if:
 - no any root meta found in DB;
 - root meta are found on a file system.

The import procedure detects whether it's a prod or CI device and imports correspondent root metadata.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -12,6 +12,7 @@
 #include "crypto/p11engine.h"
 #include "helpers.h"
 #include "http/httpclient.h"
+#include "offline/client.h"
 #include "primary/reportqueue.h"
 #include "rootfstreemanager.h"
 #include "storage/invstorage.h"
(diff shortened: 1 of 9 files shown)


mike-sul:
This change goes together with
https://github.com/foundriesio/ci-scripts/pull/202/files
https://github.com/foundriesio/meta-lmp/pull/740

src/liteclient.cc
```diff
@@ -261,6 +263,49 @@ void LiteClient::notify(const Uptane::Target& t, std::unique_ptr<ReportEvent> ev
...
+    // if libaktualizr's default value then set aklite's default value
+    tuf_meta_src = "/usr/lib/sota/tuf";
+  }
+  if (!boost::filesystem::exists(tuf_meta_src)) {
+    LOG_INFO << "Root metadata are not provisioned to FS, proceed with fetching them from Device Gateway considering "
+                "Trust On First Use";
+    return;
+  }
+
+  const std::string prod_bc_value{"production"};
```
doanac, line 283: how does this work for offline updates where the device doesn't have a certificate?

mike-sul: The cert is needed solely to determine a device type here, so we know which of the root meta set to import, ci or prod. In the case of an offline update a device just fetches what it was pointed to.

doanac: > In the case of an offline update a device just fetches what it was pointed to.

Can you elaborate here, I still don't understand. It looks like for an offline update, there will be no client cert, the LiteClient constructor will call this method, the `key_manager->getBC` will throw an error, we'll log it and then wind up defaulting to prod=false.

mike-sul: So, no certs means that this is CI device.

doanac: i believe the current offline updates requirements doc says devices will not be registered.

mike-sul: I thought it makes sense only for dev devices, let's discuss it offline today.

doanac: how about you default to production *if* the production directory has content? Then people only doing CI targets work by default, but people that have production will do the best option by default.

mike-sul: I changed it to the two-level check as we discussed offline.

mike-sul: I resolve this conversation since I reworked implementation considerably.

mike-sul:
@doanac I reworked it according to what we discussed offline.

The logic for offline and online cases is different and separated so we can improve/change the offline one if needed without touching the online logic which is good enough already.

src/liteclient.cc
```diff
@@ -261,6 +262,82 @@ void LiteClient::notify(const Uptane::Target& t, std::unique_ptr<ReportEvent> ev
...
+  if (prod && !boost::filesystem::exists(import_src / "1.root.json")) {
+    LOG_WARNING << "This production device is not provisioned with production root role metadata, download them from "
+                   "Device Gateway (TOFU)";
+    return;
+  }
+
+  if (importRootMeta(import_src)) {
+    LOG_INFO << "Successfully imported " << (prod ? "production" : "CI") << " root role metadata from " << import_src;
+  } else {
+    LOG_ERROR << "Failed to import " << (prod ? "production" : "CI") << " root role metadata from " << import_src;
```
doanac, line 337: Should we include something about "TOFU" here also to indicate things are still going to proceed?

doanac (commented):
looks good. I just have a couple of logging nits.

mike-sul:
@doanac Updated PR according to the comments as well as enabled implicit testing of the new functionality.

src/liteclient.cc
```diff
@@ -261,6 +262,83 @@ void LiteClient::notify(const Uptane::Target& t, std::unique_ptr<ReportEvent> ev
...
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "A device certificate is not found or failed to parse it: " << exc.what();
+    LOG_ERROR << "Cannot determine a device type (prodution or CI). Skip root metadata import, download them from "
+                 "Device Gateway (TOFU)";
+    return;
+  }
+
+  boost::filesystem::path import_src{std::get<1>(import) / (prod ? "prod" : "ci")};
+  if (prod && !boost::filesystem::exists(import_src / "1.root.json")) {
+    LOG_WARNING << "This production device is not provisioned with production root role metadata, download them from "
```
doanac, line 329: i hate to nit too much, but I think we should say: `will download them from` or `downloading them from`. The way that reads, it almost sounds like its a task the user must do manually.

mike-sul: updated.

doanac (approved):

PR #185, opened 2022-08-01
apps: Obtain and send info about running Apps

mike-sul:

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -535,11 +535,8 @@ Json::Value ComposeAppManager::getAppsState() const {
 
   try {
     auto apps{app_engine_->getRunningAppsInfo()};
-    if (!apps.isNull()) {
-      const auto hash{sysroot()->getDeploymentHash(OSTree::Sysroot::Deployment::kCurrent)};
-      apps_state["deviceTime"] = TimeStamp::Now().ToString();
-      apps_state["ostree"] = hash;
 
+    if (!apps.isNull()) {
       for (Json::ValueIterator ii = apps.begin(); ii != apps.end(); ++ii) {
         const auto app_name{ii.key().asString()};
         const Json::Value& services{(*ii)["services"]};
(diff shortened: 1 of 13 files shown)


src/docker/dockerclient.cc
```diff
@@ -57,6 +57,82 @@ std::tuple<bool, std::string> DockerClient::getContainerState(const Json::Value&
...
+
+int DockerClient::getContainerExitCode(const std::string& id) {
+  const std::string cmd{"http://localhost/containers/" + id + "/json"};
+  auto resp = http_client_->get(cmd, HttpInterface::kNoLimit);
+  if (!resp.isOk()) {
+    throw std::runtime_error("Request to dockerd has failed: " + cmd);
+  }
+  const auto container_info{resp.getJson()};
+  return container_info["State"]["ExitCode"].asInt();
+}
```
doanac, line 78: i know its a draft, but it would be nice if we could make that docker engine api call once and get both values.

mike-sul: We don't both, either health or exit code is requested.

mike-sul:
I wouldn't like to invest more into the api part until someone explicitly ask for it (interface methods to fetch Apps state and send it to DG).

src/composeappmanager.cc
```diff
@@ -530,6 +530,46 @@ std::string ComposeAppManager::getRunningAppsInfoForReport() const {
...
+          if (service["health"] == "unhealthy") {
+            // if it's unhealthy then an overall App is considered unhealthy
+            (*ii)["state"] = "unhealthy";
+            break;
+          }
+        }
+
+        if (!(*ii).isMember("uri")) {
+          // App's URI is stored in the `state` in the case of a regular Compose App engine
+          (*ii)["uri"] = "";  // TODO: figure out an app's URI
```
doanac, line 560: are you still trying to find a way to do this or are you planning to leave it as "TODO" for this PR and do later?

mike-sul: The later.

mike-sul: It actually works for `compose apps` case not `restorable one`, we store compose app "state" in a file which contains an app uri.

doanac (approved):
Looks good.

PR #186, opened 2022-08-11
aklite: Send Apps status only if changed

mike-sul:
Send Apps status to Device Gateway only if it has changed since the last time it was checked.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -586,13 +586,18 @@ void LiteClient::reportAppsState() {
     return;
   }
   const auto apps_state{compose_pacman->getAppsState()};
-
   if (apps_state.isNull()) {
     LOG_WARNING << "Failed to obtain Apps state, skipping sending it to Device Gateway";
     return;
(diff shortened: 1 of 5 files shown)


src/liteclient.cc
```diff
@@ -586,13 +586,19 @@ void LiteClient::reportAppsState() {
     return;
   }
   const auto apps_state{compose_pacman->getAppsState()};
-
   if (apps_state.isNull()) {
     LOG_WARNING << "Failed to obtain Apps state, skipping sending it to Device Gateway";
     return;
   }
+  if (!apps_state_.isNull() && apps_state_.isMember("apps") && apps_state.isMember("apps") &&
```
doanac, line 593: looks good. merge it like this if you want, but i wonder if this would be a little cleaner if you moved these checks into `compareAppsStats`

mike-sul: makes sense.

doanac (approved):

PR #187, opened 2022-08-19
aklite: Limit event number sent to Device Gateway

mike-sul:
- Add ability to limit a number of events to include into a single
  request to Device Gateway.
- Limit event number sent in a single HTTP POST /events request.
- Handle 413 response to HTTP POST /events request by re-trying with
  fewer events in a body until just one event is sent.
- If 413 to a request that carries just one event then just drop the
  event and stop trying to send it.
- Add unit tests for the aforementioned functionality.

### DIFF
--- a/src/liteclient.h
+++ b/src/liteclient.h
@@ -107,6 +107,8 @@ class LiteClient {
 
   std::shared_ptr<Downloader> downloader_;
   Json::Value apps_state_;
+  const int report_queue_run_pause_s_{10};
+  const int report_queue_event_limit_{6};
 };
 
 #endif  // AKTUALIZR_LITE_CLIENT_H_
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -107,7 +107,8 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, cons
   if (!uptane_fetcher_) {
     uptane_fetcher_ = std::make_shared<Uptane::Fetcher>(config, http_client);
   }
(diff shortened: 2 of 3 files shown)


src/liteclient.h
```diff
@@ -107,6 +107,8 @@ class LiteClient {
 
   std::shared_ptr<Downloader> downloader_;
   Json::Value apps_state_;
+  const int report_queue_run_pause_s_{10};
+  const int report_queue_even_limit_{6};
```
doanac, line 111: `even_limit` -> `event_limit`

doanac (commented):
i see the libaktualizr changes were labeled "fio extras", they are probably worth sharing upstream.

src/liteclient.h
mike-sul: Updated.

mike-sul:
> i see the libaktualizr changes were labeled "fio extras", they are probably worth sharing upstream.

OK, what's correct label?

doanac:
> > i see the libaktualizr changes were labeled "fio extras", they are probably worth sharing upstream.
> 
> OK, what's correct label?

`fio toup` ("to upstream")

mike-sul:
In addition to the current changes I am trying to add a proper handling of 413 in the report queue, so it won't keep trying to send "big" events.

mike-sul:
@doanac This is PR https://github.com/foundriesio/aktualizr/pull/57 to the `aktualizr` repo, I think it should make the change review easier.

doanac (approved):

PR #189, opened 2022-08-25
Enable nerdctl

mike-sul:
Adjust and extend aklite to make it working with `nerdctl` (`containerd`).


mike-sul:
This change depends on the patch to nerdctl https://github.com/foundriesio/meta-lmp/pull/796.

doanac (commented):
I have two things I'm on the fence about.

1. I like making this easy to use. But we probably don't want this everywhere. ifdef'ing this in and having a `PACKAGECONFIG` option in the bitbake recipe is probably a little nicer for now.
2. Given the number of "TODOs" we'll be encountering, I wonder if we should keep this in its own branch that we keep rebased on master for now. If we split this commit in 2, the first have could go to master and the 2nd have (containerd stuff) could be that branch. I *think* that would allow us to keep things going w/o having to deal with merge conflicts.

mike-sul:
What problem it causes just keeping in on master?
I don't have any plans to add any functionality to it. It does what we wanted, i.e. do the dev testing.
Also, it doesn't impact normal use-cases anyhow.

mike-sul:
Tested on qemu, work as expected for my basic Apps, so we can use it for dev testing.

mike-sul:
> I have two things I'm on the fence about.

I think the trade off is:

1. Introduce cmake def, e.g. `BUILD_AKLITE_WITH_NERDCTL`. 
2. Do `PACKAGECONFIG[nerdctl] = "-DBUILD_AKLITE_WITH_NERDCTL=ON,-BUILD_AKLITE_WITH_NERDCTL=OFF,"
3. Keep all changes in master.

doanac:
+1 on your suggestion. keep the code in master but make people opt-in for it.

mike-sul:
@doanac Restructured it into two commits and made the nerdctl functionality optional.

mike-sul:
ci-retest

mike-sul:
Tested it with https://github.com/foundriesio/meta-lmp/pull/799

doanac (approved):

PR #196, opened 2022-09-15
Image and container pruning

mike-sul:
appengine: Move pruning to an engine client impl
    
Don't call the image and container pruning commands directly from an app
engine. Instead, do it indirectly via the app engine client interface.
A specific implementation of the pruning is encapsulated within a
specific implementation of a container engine client.


doanac:
looks fine, but you have one empty commit in this that must have been for some CI debugging or something.

mike-sul:
> but you have one empty commit in this that must have been for some CI debugging

fixed, thx.

doanac (approved):

PR #198, opened 2022-09-28
Compare boot frmware versions to unset or increase the boot upgrade flag

mike-sul:
Change the way the value of the bootloader flag `bootupgrade_available`
is defined just after successful Target/ostree installation.
    
    - Read the current boot fw version from the rootfs that a device
      is currently booted on, instead of getting it from the boot flag.
    
    - If the boot fw version found in the new ostree deployment is
      different than the current one, then increase `bootupgrade_available`.
    
    - If the boot fw version found in the new ostree deployment is
      the same as the current then decrease `bootupgrade_available`,
      unless it's already equal zero.

### DIFF
--- a/src/bootloader/bootloaderlite.cc
+++ b/src/bootloader/bootloaderlite.cc
@@ -9,10 +9,12 @@
 
 namespace bootloader {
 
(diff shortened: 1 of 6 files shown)


mike-sul:
@Tim-Anderson @igoropaniuk Please, check out if it works for you.

PR #199, opened 2022-10-03
bootloader: Remove redundant "rollback" classes

mike-sul:
- Remove redundant "rollback" interfaces and classes since they basically implement the same logic as the `Bootloader` class implemented within libaktaulizr.

- The only difference/specialization is in implementation of `installNotify` which is not used in libaktualizr at all.

- Therefore, this change removes all "rollback" classes and just overrides `installNotify` in BootloaderLite. The overridden `installNotify` encapsulates the logic that was spread across the "rollback" classes.

- Added `bootloader` namespace to encapsulate the bootloader-related logic in it, and to align it with the namespacing approach of aklite, (each subdirectory implies namespace).


mike-sul:
@igoropaniuk @Tim-Anderson Just FYI. This is just refactoring aimed to simplify an overall implementation, the business logic of the boot flag management is exactly the same as it is now in master.

doanac (approved):
its a lot simpler. logically it looks the same, i'll trust it works

PR #202, opened 2022-10-13
boot: Block update if boot fw update is in progress

mike-sul:
Ban ostree update if there is ongoing boot fw update and implicitly initiate device reboot by returning the `need completion` return code. The reboot is need to confirm and finalize the boot fw update.

### DIFF
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -5,6 +5,23 @@
 #include "ostree/repo.h"
 #include "target.h"
 
+RootfsTreeManager::RootfsTreeManager(const PackageConfig& pconfig, const BootloaderConfig& bconfig,
+                                     const std::shared_ptr<INvStorage>& storage,
+                                     const std::shared_ptr<HttpInterface>& http,
+                                     std::shared_ptr<OSTree::Sysroot> sysroot, const KeyManager& keys)
+    : OstreeManager(pconfig, bconfig, storage, http, new bootloader::BootloaderLite(bconfig, *storage, sysroot)),
+      sysroot_{std::move(sysroot)},
+      boot_fw_update_status_{new bootloader::BootloaderLite(bconfig, *storage, sysroot)},
+      http_client_{http},
+      gateway_url_{pconfig.ostree_server},
+      keys_{keys} {
+  const std::string update_block_attr_name{"ostree_update_block"};
+  if (pconfig.extra.count(update_block_attr_name) == 1) {
+    std::string val{pconfig.extra.at(update_block_attr_name)};
+    update_block_ = val != "0" && val != "false";
+  }
+}
+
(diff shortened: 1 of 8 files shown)


mike-sul:
@igoropaniuk This is the change to aklite we discussed - Disallow ostree update while there is an ongoing boot fw update that requires confirmation & finalization through device reboot.
This change does both, forbids ostree update and makes the aklite daemon to exit and reboot device.

mike-sul:
@doanac FYI

src/rootfstreemanager.cc
```diff
@@ -83,10 +83,18 @@ data::InstallationResult RootfsTreeManager::install(const Uptane::Target& target
...
   if ((current.sha256Hash() != target.sha256Hash()) ||
       (!sysroot()->getDeploymentHash(OSTree::Sysroot::Deployment::kPending).empty() &&
        sysroot()->getDeploymentHash(OSTree::Sysroot::Deployment::kPending) != target.sha256Hash())) {
+    // If the boot fw update is in progress and it is an installation of Target ostree of which differs
+    //  from the ostree a device is booted on then
+    if (boot_fw_update_status_->isUpdateInProgress() && (current.sha256Hash() != target.sha256Hash())) {
+      LOG_WARNING << "Boot fw update is in progress."
+                     " A device must be rebooted to confirm and finalize the boot fw update"
+                     " before installation of a new Target with ostree/rootfs change";
+      return data::InstallationResult(data::ResultCode::Numeric::kNeedCompletion, "");
```
doanac, line 96: doesn't our main loop interpret this response as "install went well, reboot now"?

mike-sul: Yep, that's exactly what we need in this case.

mike-sul: We need to reboot twice. At first to confirm and finalize ostree. And then before installation of another new ostree version we need to do another reboot to confirm and finalize boot fw update.

doanac: is this going to trick our backend into thinking the device is about install this new target and not that it needs to do something for the previous one?

mike-sul: the previous one is already installed.
aklite will install this new Target just after reboot.

mike-sul: I just would like to get confirmation from @igoropaniuk that this idea works at all, before maturing it.

doanac: i think your explanation makes sense. i think maybe just add some comment text to help explain.

mike-scott (commented):
Hoping to set this up and test it a bit tomorrow on real HW.

Tim-Anderson (approved):
LGTM

igoropaniuk (approved):
LGTM

PR #203, opened 2022-10-25
aklite-offline: Support update if no Apps or dockerless system

mike-sul:
Make sure the docker client path/file exists before its usage.

1. This is workaround to make the `aklite-offline` work on a device that does not include `/usr/bin/docker` but has `dockerd`.

2. More accurate support of just ostree update if docker installed or no Apps are present in the offline update content.

### DIFF
--- a/apps/aklite-offline/cmds.cpp
+++ b/apps/aklite-offline/cmds.cpp
@@ -14,13 +14,19 @@ int InstallCmd::installUpdate(const Config& cfg_in, const boost::filesystem::pat
     switch (install_res) {
       case offline::PostInstallAction::NeedReboot: {
         ret_code = 100;
-        std::cout << "Please reboot a device and run `run` command to apply installation and start the updated Apps\n";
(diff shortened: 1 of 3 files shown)


apps/aklite-offline/cmds.cpp
```diff
@@ -14,7 +14,8 @@ int InstallCmd::installUpdate(const Config& cfg_in, const boost::filesystem::pat
     switch (install_res) {
       case offline::PostInstallAction::NeedReboot: {
         ret_code = 100;
-        std::cout << "Please reboot a device and run `run` command to apply installation and start the updated Apps\n";
+        std::cout << "Please reboot a device and run `run` command to apply installation and start the updated Apps "
+                     "(unless no Apps to update or dockerless system)\n";
```
doanac, line 18: maybe we just do the wording better here instead of the v88 fix.

mike-sul: agree.

PR #205, opened 2022-10-25
Fix doockerless or appless offline update v88

mike-sul:


apps/aklite-offline/cmds.cpp
```diff
@@ -14,7 +14,8 @@ int InstallCmd::installUpdate(const Config& cfg_in, const boost::filesystem::pat
     switch (install_res) {
       case offline::PostInstallAction::NeedReboot: {
         ret_code = 100;
-        std::cout << "Please reboot a device and run `run` command to apply installation and start the updated Apps\n";
+        std::cout << "Please reboot a device and run `run` command to apply installation and start the updated Apps "
+                     "(unless no Apps to update or dockerless system)\n";
```
doanac, line 18: is it hard to know which mode the device is in here? That might help make a more clear message.

Regardless - this is a worded in a tricky way that might lead a user to believe they don't need to reboot if there are no apps configured. What you are saying is that you don't need to do the `run` command after the reboot.

mike-sul: Maybe "and start the updated Apps if any"?

doanac: lets ask @kprosise - we have a professional that can help us :)

mike-sul: @kprosise 

We have two use-cases that makes the `run` after reboot to behave differently:
1. The `run` applies/confirms the new ostree installation and starts the updated Apps.

2. The `run` only applies/confirms the new ostree installation, no any Apps starting.

The question is how to articulate it in a single log message so a user is not confused.

Currently it is: `"Please reboot a device and run `run` command to apply installation and start the updated Apps (unless no Apps to update or dockerless system)"`

PR #208, opened 2022-11-21
docker: Do auth according to the spec

mike-sul:
Implement the v2 Docker registry authentication scheme as it is described in
https://github.com/distribution/distribution/blob/main/docs/spec/auth/token.md.

As a result, the aklite's docker client requests a token from the authorization service specified by Registry in the `WWW-Authenticate` response header.


mike-sul:
It's WIP, will add unit tests.

doanac:
on the change https://github.com/foundriesio/aktualizr/compare/0e22dcfcb01c6130e404012112f9c359476f3780...dc494d1e3d42794caaff828de386c27ca22271e5#diff-c543986eff9ba61e0219749a171ae0cbf692890dd8b16107c7a349922a473483R65 - is it correct to take the value of the *first* header you parse? I'd assume in the case you see it twice, you'd take the 2nd value?

mike-sul:
> on the change [foundriesio/aktualizr@0e22dcf...dc494d1#diff-c543986eff9ba61e0219749a171ae0cbf692890dd8b16107c7a349922a473483R65](https://github.com/foundriesio/aktualizr/compare/0e22dcfcb01c6130e404012112f9c359476f3780...dc494d1e3d42794caaff828de386c27ca22271e5#diff-c543986eff9ba61e0219749a171ae0cbf692890dd8b16107c7a349922a473483R65) - is it correct to take the value of the _first_ header you parse? I'd assume in the case you see it twice, you'd take the 2nd value?

The current implementation overrides existing header with a new one, so the last received header value wins.
Not sure it's worth supporting many values for the same header.

StealthyCoder (approved):

mike-sul:
@doanac I think that's good enough. I hope this effort will be justified for the `E` case when Registry will have `E` specific domain.

doanac (approved):

PR #215, opened 2022-12-21
Extract only a compose file to verify App

mike-sul:
Extract just compose file from the App archive in order to check a
compose project configuration during the "verification" step of
an overall update cycle.


mike-scott (approved):
LGTM

PR #216, opened 2023-01-09
apps: Docker pull images just after App install

mike-sul:
Run `docker-compose pull` for each installed App before creating or starting App. The pull command updates the docker store so it becomes aware of the App images. As result, the following `compose up` command won't require network connection to pull any App image data. It helps to distinguish between networking errors and pure container creation or start errors. The first one is NOT critical hence aklite should continue trying to install App. The second error type is critical and aklite should mark corresponding Target as "failing" in such the case.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -139,7 +139,7 @@ AppEngine::Result RestorableAppEngine::install(const App& app) {
   if (!create_containers_if_install_) {
     return installContainerless(app);
   }
-  return installAndCreateContainers(app);
+  return installAndCreateOrRunContainers(app);
 }
 
 AppEngine::Result RestorableAppEngine::installContainerless(const App& app) {
@@ -153,7 +153,7 @@ AppEngine::Result RestorableAppEngine::installContainerless(const App& app) {
   return res;
 }
 
-AppEngine::Result RestorableAppEngine::installAndCreateContainers(const App& app) {
+AppEngine::Result RestorableAppEngine::installAndCreateOrRunContainers(const App& app, bool run) {
   Result res{false};
 
   try {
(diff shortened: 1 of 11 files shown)


src/docker/restorableappengine.cc
```diff
@@ -756,6 +786,12 @@ void RestorableAppEngine::verifyComposeApp(const std::string& compose_cmd, const
        boost::process::start_dir = app_dir);
 }
 
+void RestorableAppEngine::pullComposeAppImages(const std::string& compose_cmd, const boost::filesystem::path& app_dir,
+                                               const std::string& flags) {
+  exec(boost::format{"%s pull %s"} % compose_cmd % flags, "failed to pull Compose App images",
+       boost::process::start_dir = app_dir);
```
doanac, line 792: Just to make sure I'm following this: The "pullImage" method called by `pullAppImages` called by `fetch` will have loaded everything. This method is really just updating docker with tag information which requires a network hop?

Maybe a comment or something in code would help or calling the method something different?

mike-sul: >The "pullImage" method called by pullAppImages called by fetch will have loaded everything. This method is really just updating docker with tag information which requires a network hop?

Correct (hash/digest not tag to be precise).

mike-sul: >>Maybe a comment or something in code would help or calling the method something different?

I'll add comment.

mike-sul: Also, the `install` and `run` method implementation is almost the same, I think I'll add common method for them in a separate commit.

mike-sul: @doanac updated by a force push.

doanac (approved):

mike-sul:
tested on a device.

PR #221, opened 2023-03-05
offline-client: Fix app blob download

mike-sul:
istream::readsome() is not quite good way to read data from an input stream as it reads "immediately available" characters from an input stream. The "immediately available" is implementation and platform specific, so its behavior is not consistent across all platforms. Also, as the doc states it
"may always return zero and extract no characters".


doanac (approved):
c++ io streams are garbage.

PR #223, opened 2023-03-09
offline: Fix dockerless system support

mike-sul:
Don't initialize and pass around the docker client's default http client within the offline updater, instead, do nothing as the docker client initializes the default one by itself if no custom http client is specified in its ctor.


mike-sul:
@doanac Do you have a chance to review it?

doanac (approved):
sorry - i thought i hit approve yesterday.

PR #224, opened 2023-03-09
custom client: Handle rollback for failed apps

doanac:
When `pacman.create_containers_before_reboot = 0`, we can hit issues where containers fail to get created upon the reboot after an OTA. In this scenario we need to rollback to the old Target.

### DIFF
--- a/examples/custom-client-cxx/main.cc
+++ b/examples/custom-client-cxx/main.cc
@@ -67,7 +67,26 @@ int main(int argc, char **argv) {
       }
 
       auto latest = res.GetLatest();
-      LOG_INFO << "Found Latest Target: " << latest.Name();
+      if (client.IsRollback(latest)) {
+        LOG_INFO << "Latest Target is marked for causing a rollback and won't be installed: " << latest.Name();
+      } else {
+        LOG_INFO << "Found Latest Target: " << latest.Name();
+      }
+
+      if (client.IsRollback(latest) && current.Name() == latest.Name()) {
+        // Handle the case when Apps failed to start on boot just after an update.
+        // This is only possible with `pacman.create_containers_before_reboot = 0`.
+        LOG_INFO << "The currently booted Target is a failing Target, finding Target to rollback to...";
+        const TufTarget rollback_target = client.GetRollbackTarget();
+        if (rollback_target.IsUnknown()) {
+          LOG_ERROR << "Failed to find Target to rollback to after failure to start Apps at boot of a new sysroot";
+          std::this_thread::sleep_for(std::chrono::seconds(interval));
+          continue;
+        }
+        latest = rollback_target;
+        LOG_INFO << "Rollback Target is " << latest.Name();
+      }
+
       if (latest.Name() != current.Name() && !client.IsRollback(latest)) {
         std::string reason = "Updating from " + current.Name() + " to " + latest.Name();
         auto installer = client.Installer(latest, reason);
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -43,6 +43,18 @@ class TufTarget {
    */
   const Json::Value &Custom() const { return custom_; }
 
+  /**
+   * Is this a known target in the Tuf manifest? There are two common causes
+   * to this situation:
+   *  1) A device has been re-registered (sql.db got wiped out) and the
+   *     /var/sota/installed_versions file is missing. The device might
+   *     running the correct target but the system isn't sure.
+   *  2) A device might be running a Target from a different tag it's not
+   *     configured for. This means the Target isn't present in the targets.json
+   *     this device is getting from the device-gateway.
+   */
+  bool IsUnknown() const { return name_ == "unknown"; }
+
  private:
   std::string name_;
   std::string sha256_;
@@ -231,6 +243,13 @@ class AkliteClient {
    * Return the Target currently running on the system.
    */
   TufTarget GetCurrent() const;
+
+  /**
+   * Find the Target to rollback to in the event the current target wasn't
+   * able to start it's Apps after rebooting from an ostree change. This
+   * situation is only possible when `pacman.create_containers_before_reboot = 0`.
+   */
+  TufTarget GetRollbackTarget() const;
 
   /**
    * Check in with device-gateway to get server managed information about
--- a/src/api.cc
+++ b/src/api.cc
@@ -163,16 +163,17 @@ boost::property_tree::ptree AkliteClient::GetConfig() const {
   return pt;
 }
 
-TufTarget AkliteClient::GetCurrent() const {
-  auto current = client_->getCurrent();
+static TufTarget convertTarget(const Uptane::Target& target) {
   int ver = -1;
   try {
-    ver = std::stoi(current.custom_version(), nullptr, 0);
+    ver = std::stoi(target.custom_version(), nullptr, 0);
   } catch (const std::invalid_argument& exc) {
-    LOG_ERROR << "Invalid version number format: " << current.custom_version();
+    LOG_ERROR << "Invalid version number format: " << target.custom_version();
   }
-  return TufTarget(current.filename(), current.sha256Hash(), ver, current.custom_data());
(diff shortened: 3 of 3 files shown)


examples/custom-client-cxx/main.cc
```diff
@@ -67,7 +67,26 @@ int main(int argc, char **argv) {
...
+        // This is only possible with `pacman.create_containers_before_reboot = 0`.
+        LOG_INFO << "The currently booted Target is a failing Target, finding Target to rollback to...";
+        const TufTarget rollback_target = client.GetRollbackTarget();
+        if (rollback_target.IsUnknown()) {
+          LOG_ERROR << "Failed to find Target to rollback to after failure to start Apps at boot of a new sysroot";
+          std::this_thread::sleep_for(std::chrono::seconds(interval));
+          continue;
+        }
+        latest = rollback_target;
+        LOG_INFO << "Rollback Target is " << latest.Name();
```
mike-sul, line 87: Maybe write "Rolling back or will roll back..." to inform about the further action the client is going to do.

examples/custom-client-cxx/main.cc
```diff
@@ -67,7 +67,26 @@ int main(int argc, char **argv) {
...
 
       auto latest = res.GetLatest();
-      LOG_INFO << "Found Latest Target: " << latest.Name();
+      if (client.IsRollback(latest)) {
+        LOG_INFO << "Latest Target is marked for causing a rollback and won't be installed: " << latest.Name();
+      } else {
+        LOG_INFO << "Found Latest Target: " << latest.Name();
+      }
+
+      if (client.IsRollback(latest) && current.Name() == latest.Name()) {
```
mike-sul, line 76: Not really important. Maybe we should introduce `operator==` for TufTarget.

src/api.cc
```diff
@@ -163,17 +163,18 @@ boost::property_tree::ptree AkliteClient::GetConfig() const {
   return pt;
 }
 
-TufTarget AkliteClient::GetCurrent() const {
-  auto current = client_->getCurrent();
+static TufTarget convertTarget(const Uptane::Target& target) {
```
mike-sul, line 166: Not important. It could be TufTarget ctor that accepts `const Uptane::Target` as an input param.

include/aktualizr-lite/api.h
```diff
@@ -232,6 +244,13 @@ class AkliteClient {
    */
   TufTarget GetCurrent() const;
 
+  /**
+   * Find the Target to rollback to in the event the current target wasn't
```
mike-sul, line 248: I am not sure that we need to be so specific about using this API function. The rollback target it returns is valid in any other use-cases too. For example, in the case of "just App" update and rollback it should return a valid "rollback" target. I understand that in this case "current" can be used as "rollback" one.
My point is that it may simplify a client code if `GetRollbackTarget` is used in all rollback cases.

mike-sul (approved):
LGTM. I've just got minor comments.

examples/custom-client-cxx/main.cc
doanac: i was on the fence on this one as well. The very next log message is almost always "Updating from X to Y". So I thought when you read them in sequence it kind of made sense.

include/aktualizr-lite/api.h
doanac: This is somewhat true, but is there another way this could ever get used given our current user-facing API? 

This function seems so specific to this one use case that I wish we could somehow embed the `if (client.IsRollback(latest) && current.Name() == latest.Name()) {` logic in main.cc into our API. i.e. somehow `GetLatest` would return the rollback Target instead but somehow also let you know that its not really "latest".

mike-sul: I think `GetRollbackTarget()` can be used in all other "rollback" cases, specifically when install() or finalizeInstall() returns error. btw, the api client does finalization implicitly, hiding it from a client while liteclient requires its clients do it explicitly.


Regarding `GetLatest` to return either latest or rollback target, I think it really depends on the approach, explicit vs implicit. I would prefer explicit and more mechanistic API, so clients can have more flexibillity in terms of implementation of their client, e.g. maybe they would like to be aware about rollback target and gets some "approval"/confirmation from upper stack to roll back to it.

doanac: > I think GetRollbackTarget() can be used in all other "rollback" cases, specifically when install() or finalizeInstall() returns error. btw, the api client does finalization implicitly, hiding it from a client while liteclient requires its clients do it explicitly.

But since the existing API is handling that internally - its not actually possible, right? i.e. the only use case in the current api to use this function is this one app failure situation.

> I think it really depends on the approach, explicit vs implicit

I don't want it to be implicit. But this PR's approach is error-prone in that you have to understand it deeply and when/how to use. It feels a bit like the "Ah - you forget to check for NULL" approach. I'm not quite sure how to make it more clean/explicit. Maybe the `CheckInResult` struct could somehow include this additional context?

mike-sul: > > I think GetRollbackTarget() can be used in all other "rollback" cases, specifically when install() or finalizeInstall() returns error. btw, the api client does finalization implicitly, hiding it from a client while liteclient requires its clients do it explicitly.
> 
> But since the existing API is handling that internally - its not actually possible, right? i.e. the only use case in the current api to use this function is this one app failure situation.

I don't think so. The `LiteInstall::Install()` does just install, if it fails then it's up to a client to implement rollback in all three rollback cases ("just app" install failure, ostree+app install failure because of ostree rollback, ostree+app install failure because of app creation or start failure after reboot). My point is that in these three rollback cases a client could use `GetRollbackTarget()`.

doanac: k - could you help show me where else I need to be checking in the example file? If I'm missing logic - then everyone else doing this will be missing logic as well.

mike-sul: The logic I mean is missing because we haven't had GetRollbackTarget() so far, and the rollback for two cases was performed implicitly through `} else {
        auto installer = client.CheckAppsInSync();` branch by syncing of the current but not latest target.
        
Now, since we have `GetRollbackTarget()` then we can do rollback explicitly for all three cases.

In this spot

```
} else {
          LOG_ERROR << "Unable to install target: " << ires;
        }
```
We can print logs saying that we are rolling back.

As for the ostree rollback after reboot case we cannot detect it explicitly and print corresponding error message because the finalization is hidden in Init(). If a client  called it directly then it could detect an error and print message about rolling back to the previous version.

The current version of the client example does the rollback for the two cases via syncing apps, and they are not considered as rollback at all, just installation failure.
So, from this perspective we have just one rollback case, if so,  then this PR is fine.

I guess that's fine as long as there are no two failing targets in raw, then the rollback through app syncing will lead to syncing to the failing target. Calling `GetRollbackTarget` and rolling back to the target it returns will do a correct rollback.

mike-sul:
@doanac Hey Andy I don't think we need to change and/or improve anything in this PR. It does exactly what was requested, i.e. exposes `GetRollbackTarget`, and the person that requested it does understand all the "rollback" use-cases.

doanac:
Yep. I think you are right. I just stared at the example main.cc for a while now. I'm not sure we can simplify it much more. I could re-structure things little so that there was an "installTarget" type function - then you could reduce the code in the main loop and make the if/else statements a little more straightforward (we sort of abuse the loop now to make latest=rollback so that we can run through the same install condition below it.

We might also change something like `GetLatest` to return a `tuple(target, isRollback)` - that way the compiler forces you to do something with it and not accidentally forget to call `IsRollback` - but that's bordering on subjective. I think I'll just merge this how it is for now.

doanac:
rebased on master

mike-sul:
@doanac  I think the github workflow actions (unit tests) are not triggered because

```
Workflows don't run in forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.
```
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories-1

doanac:
clsing in favor of https://github.com/foundriesio/aktualizr-lite/pull/228

PR #225, opened 2023-03-10
offline: Don't pull images at App start

mike-sul:
offline: Don't pull images at App start
    
By default, aklite/restorable-engine tries to pull images at App start
to distinguish between download and install failure during target
installation. E.g. if a device goes offline just after App download and
before installation then install will fail because it cannot check App
image manifest at Registry. Having ability to distinguish between fetch
and a real install failure allows aklite to keep target install retrying
if it is just fetch failure.
But it does not work for the offline client and the early app startup
because a device is really offline and it is assumed that all App data
are already preloaded, so no need to do `compose pull` at App startup.
Otherwise, it will just fail since `compose pull` tries to fetch an
image manifest regardless its presence in a local docker store.

### DIFF
--- a/apps/aklite-apps/cmds.cpp
+++ b/apps/aklite-apps/cmds.cpp
@@ -172,8 +172,19 @@ int RunCmd::runApps(const std::vector<std::string>& shortlist, const std::string
   auto http_client = std::make_shared<HttpClient>();
   auto docker_client{std::make_shared<Docker::DockerClient>()};
   auto registry_client{std::make_shared<Docker::RegistryClient>(http_client, "")};
-  Docker::RestorableAppEngine app_engine{store_root,    compose_root, docker_root, registry_client,
-                                         docker_client, client,       docker_host, compose_client};
+  Docker::RestorableAppEngine app_engine{
+      store_root,
+      compose_root,
+      docker_root,
+      registry_client,
+      docker_client,
+      client,
+      docker_host,
+      compose_client,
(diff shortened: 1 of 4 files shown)


apps/aklite-apps/cmds.cpp
```diff
@@ -172,8 +172,19 @@ int RunCmd::runApps(const std::vector<std::string>& shortlist, const std::string
...
+      docker_root,
+      registry_client,
+      docker_client,
+      client,
+      docker_host,
+      compose_client,
+      Docker::RestorableAppEngine::GetDefStorageSpaceFunc(),
+      [](const Docker::Uri& /* app_uri */, const std::string& image_uri) { return "docker://" + image_uri; },
+      false,
+      true};
```
doanac, line 187: i could have reviewed this in half the time if C++ had named function args :)

doanac (approved):

apps/aklite-apps/cmds.cpp
mike-sul: Probably it's time for introduction of `AppEngineArgs` structure.

doanac: Its not that big a deal. I just wasn't in a good "C++ headspace" and got the feeling this was changing other stuff - then i  realized it was just there to satisfy the compiler

mike-sul: yep, the clang format.

PR #230, opened 2023-03-16
Offline update rollback handling

mike-sul:


mike-sul:
@doanac This change fixes and improves rollback handling in different scenarios including rollback when the initial/previous Apps were preloaded.

Also, it detects and informs a client about rollback to "unknown". aklite-offline does nothing in this case except returning dedicated/specific status code and logging corresponding messages.
In the next PR, I'll try to address the "unknown" case by turning "unknown" to "known" - initial device state, which is ostree hash a device is booted on and Apps it currently runs.

src/rootfstreemanager.cc
```diff
@@ -138,6 +138,19 @@ data::InstallationResult RootfsTreeManager::install(const Uptane::Target& target
   return res;
 }
 
+data::InstallationResult RootfsTreeManager::finalizeInstall(const Uptane::Target& target) {
+  // "finalize" the ostree update even if there is no ostree update (current.hash == target/pending.hash)
+  // so the reboot flag is removed if there was a reboot was some reason just before calling finalize.
```
doanac, line 143: I'm not quite sure how we can get into this situation. Also wondering if this should instead be fixed in the OstreeManager logic?

src/offline/client.cc
doanac, line 1: this is nice.

doanac (commented):
nice work. i have one question about the finalizeInstall method

src/rootfstreemanager.cc
mike-sul: It happens when `finalizeInstall` is called if there is pending Target that updates just Apps. It happens all the time for the offline aklite updating just Apps; it requires the docker engine restart, so install()---- pending target --> dockerd restart ----> finalize()(app start) ---pending target becomes installed.

I think it can also may happen for the online client, it's just hard to reproduce.

mike-sul: It can be implemented in OStreeManager, I just don't want to bother with up-streaming it, and I am not sure it's really needed for libaktualizr/aktualizr.

doanac: care to add something about the offline client in the comment? This is the kind of line of code somebody might want to delete or "refactor" some day thinking it doesn't make sense.

mike-sul: I updated the comment and added more details.

mike-sul: As a matter of fact, after reproducing it for a normal case/flow, I think the fix should be done in libaktualizr (as you suggested). I just don't wanna do it now, instead, I created a ticket for it and do it later once we collect at least >2 things to change in libaktualizr.

doanac (approved):
the comment makes total sense now. thanks

mike-sul:
[ci-retest]

PR #231, opened 2023-03-20
Turning the "unknown" Target to the initial Target based on what is currently installed

mike-sul:
If `installed_versions` is not specified or the DB is removed then aklite considers the current Target as "unknown". It causes an issue if a rollback happens since aklite does not know what Apps to (re-)install and run on rolback.

The current change aims to fix it by deducing the current Target taking into account the current ostree hash a device is booted on and currently installed Apps.

Therefore, this change significantly reduces occurrences of the rollback to "unknown". It may happen only if aklite fails to get a list of currently running Apps.


tests/aklite_offline_test.cc
```diff
@@ -425,29 +429,75 @@ TEST_F(AkliteOffline, RollbackIfAppStartFailsWithAppShortlisting) {
...
   const auto new_target{addTarget({createApp("app-01", "compose-start-failure")})};
+
+  // make the initial Target setting fail, so the current Target is "unknown"
+  const Docker::Uri uri{Docker::Uri::parseUri(app01.uri)};
+  const auto app_dir{test_dir_.Path() / "reset-apps" / "apps" / uri.app / uri.digest.hash()};
+  Utils::writeFile(app_dir / Docker::Manifest::Filename, std::string("broken json"));
   ASSERT_EQ(install(), offline::PostInstallAction::NeedReboot);
   reboot();
   ASSERT_EQ(run(), offline::PostRunAction::RollbackToUnknownIfAppFailed);
+  // Cannot perform rollback to unknown, so still booted on a new target's hash
```
mike-sul, line 500: I am sitting on a fence about this use-case. There are two options:

1. Leave it as it is. A device is booted on a new target's hash but cannot run the updated Apps. No rollback because the rollback Target is "unknown".

2. Re-deploy the rollback ostree deployment if found any (even if it's unknown), and return `RollbackNeedReboot`, so the following reboot makes a device to boot on the unknown Target's hash.

doanac: It would be pretty hard to get into this situation, right? That kind of makes me lean towards going with 1 since its not requiring a reboot. Neither situation is great, but at least with option 1 you aren't rebooting a device

doanac (commented):
nice work. i looked into this a bit over the weekend, and the way you are getting the app uri is about the best i could find also. 

somehow forcing a label into the compose services with this URI seems to be the only other decent option and that makes them change to compose and thus restart when then didn't need to. plus, we'd have to update compose to do something like that.

tests/aklite_offline_test.cc
mike-sul: I incline towards the option 1 too. The online aklite can recover from such state by applying the following update, and it's always better to stay on the newest ostree/rootfs to do so. When it comes to the offline update, I think it's up to a client to decide what to do in such the case, whether to rollback to the previous rootfs version or stay with the current one and try to start specific apps manually. I will document it.

src/target.h
```diff
@@ -38,6 +39,18 @@ class Target {
...
+    Json::Value target_json;
+    Json::Value custom;
+
+    custom["targetFormat"] = target.type();
+    custom["hardwareIds"][0] = hw_id;
+    target_json["hashes"]["sha256"] = target.sha256Hash();
+    target_json["length"] = 0;
+    target_json["custom"] = custom;
+    return {InitialTarget, target_json};
+  }
```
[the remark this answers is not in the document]
mike-sul: no reason, I'll update it.

src/rootfstreemanager.cc
```diff
@@ -87,9 +87,14 @@ void RootfsTreeManager::setInitialTargetIfNeeded(const std::string& hw_id) {
...
-  const auto init_target{Target::toInitial(current, hw_id)};
-  storage_->savePrimaryInstalledVersion(init_target, InstalledVersionUpdateMode::kCurrent);
+  try {
+    // Turning "unknown" Target to "initial" one
+    Uptane::Target init_target{Target::toInitial(current, hw_id)};
+    completeInitialTarget(init_target);
+    storage_->savePrimaryInstalledVersion(init_target, InstalledVersionUpdateMode::kCurrent);
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "Failed to set the initial Target: " << exc.what();
+  }
```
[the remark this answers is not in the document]
mike-sul: yes, OOP implies lots of boilerplate code.

apps/aklite-offline/cmds.cpp
```diff
@@ -69,19 +69,19 @@ int RunCmd::runUpdate(const Config& cfg_in) const {
         break;
       }
       case offline::PostRunAction::RollbackOk: {
-        LOG_INFO << "Installation has failed so a device was rolled back to the previous version";
+        LOG_INFO << "Installation has failed so a device rollbacked to the previous version";
```
doanac, line 72: "rolled back". we need to make the past tense of "roll" not "back".

vkhoroz: Quite expectedly, Oxford says that _rollback_ can only be used as a noun:
https://www.oxfordlearnersdictionaries.com/definition/english/rollback?q=rollback

vkhoroz: IIUC furthermore, the device is an object here, not a subject... thus a passive form used in the original _was rolled back_.

@kprosise what do you recommend.

mike-sul: I made `device` a subject for purpose,  I think it's more correct.

vkhoroz: @mike-sul the purpose matters... I like that you gave it a thought ;)

kprosise: I'm enjoying seeing these sorts of conversations!

mike-sul:
@doanac I've done with e2e testing and everything looks fine to me.

The only thing that I am not sure about is this https://github.com/foundriesio/aktualizr-lite/pull/231/commits/fcf88fe273977d680e8d4ad94d9a9e45f304e4ea. It helps to restart a container with "no restart" policy during rollbacking to the previous version in the case of offline update. Specifically, container that is actually not updated/changed in the new Target, but we stop it in the case of offline update by restarting dockerd.

On the one side, it's expected that "no restart" container is not running after dockerd restart.
On the other side, a user expects containers to be in exactly the same state as they were before an update attempt if a rollback is successful.

kprosise (approved):
For printed content, it LGTM

doanac (approved):

PR #233, opened 2023-04-17
Handle egde cases of bootloader update

mike-sul:
- Allow any format of a bootloader version if no rollback protection.
- Proceed with update if a version file is not found in a new target.
- Reject update if new bootloader version is malformed.
- Proceed with update if the current bootloader version is not set (env var) or is malformed provided that a new version is correct.


ricardosalveti (approved):
LGTM

doanac (approved):

PR #236, opened 2023-05-05
liteclient: Drain events before destruction

mike-sul:
Make sure all report events are sent to Device Gateway at the beginning of the liteclient instance destruction.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -130,7 +130,10 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, cons
   sysroot_ = ostree_sysroot;
 }
 
-LiteClient::~LiteClient() {}  // NOLINT(modernize-use-equals-default, hicpp-use-equals-default)
+LiteClient::~LiteClient() {
+  // Make sure all events drained before fully destroying the liteclient instance.
+  report_queue.reset(nullptr);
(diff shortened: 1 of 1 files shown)


src/liteclient.cc
```diff
@@ -130,7 +130,10 @@ LiteClient::LiteClient(Config& config_in, const AppEngine::Ptr& app_engine, cons
   sysroot_ = ostree_sysroot;
 }
 
-LiteClient::~LiteClient() {}  // NOLINT(modernize-use-equals-default, hicpp-use-equals-default)
+LiteClient::~LiteClient() {
+  // Make sure all events drained before fully destroying the liteclient instance.
+  report_queue.reset(nullptr);
```
doanac, line 135: This should make it much less likely to happen. I'm not sure the conditions around how this happened, but could we still get error messages if the thread is in the middle of sending an event?

mike-sul: >but could we still get error messages if the thread is in >the middle of sending an event?

No, because the reportqueue dtor is implemented correctly.

We saw error messages because the reportqueue was sending events while the keymanager has been already destroyed/null, and the event sending implies getting access to hsm...

mike-sul: >This should make it much less likely to happen. 

It won't happen at all now.

PR #239, opened 2023-05-29
api: Skeleton e2e code for the CLI API

mike-sul:
Initial version of the CLI API and its initial implementation that covers just the happy path/flow.

### DIFF
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -28,6 +28,8 @@ enum class ExitCode {
   DownloadFailure = 50,
   InstallNeedsRebootForBootFw = 90,
   InstallNeedsReboot = 100,
+  InstallRollbackOk = 110,
+  InstallRollbackFailed = 130,
 };
 
 ExitCode Install(AkliteClient &client, int version = -1);
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -29,6 +29,7 @@ enum class ExitCode {
   InstallNeedsRebootForBootFw = 90,
   InstallNeedsReboot = 100,
   InstallRollbackOk = 110,
+  InstallRollbackNeedsReboot = 120,
   InstallRollbackFailed = 130,
 };
 
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -26,6 +26,8 @@ enum class ExitCode {
   InstallationInProgress = 30,
   NoPendingInstallation = 40,
   DownloadFailure = 50,
+  DownloadFailureNoSpace = 60,
+  DownloadFailureVerificationFailed = 70,
   InstallNeedsRebootForBootFw = 90,
   InstallNeedsReboot = 100,
   InstallRollbackOk = 110,
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -28,6 +28,7 @@ enum class ExitCode {
   DownloadFailure = 50,
   DownloadFailureNoSpace = 60,
   DownloadFailureVerificationFailed = 70,
+  InstallAppPullFailure = 80,
   InstallNeedsRebootForBootFw = 90,
   InstallNeedsReboot = 100,
   InstallRollbackOk = 110,
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -23,6 +23,7 @@ enum class ExitCode {
   Ok = 0,
   InstallationInProgress = 10,
   NoPendingInstallation = 20,
+  DownloadFailure = 40,
   InstallNeedsReboot = 100
 };
 
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -21,10 +21,13 @@ namespace cli {
 enum class ExitCode {
   UnknownError = -1,
   Ok = 0,
-  InstallationInProgress = 10,
-  NoPendingInstallation = 20,
-  DownloadFailure = 40,
-  InstallNeedsReboot = 100
+  TufMetaPullFailure = 10,
+  TufTargetNotFound = 20,
+  InstallationInProgress = 30,
+  NoPendingInstallation = 40,
+  DownloadFailure = 50,
+  InstallNeedsRebootForBootFw = 90,
+  InstallNeedsReboot = 100,
 };
 
 ExitCode Install(AkliteClient &client, int version = -1);
@@ -72,6 +75,15 @@ class TufTarget {
    *     this device is getting from the device-gateway.
    */
   bool IsUnknown() const { return name_ == "unknown"; }
+
+  /**
+   * @brief TODO
+   * @param lhr
+   * @return
+   */
+  bool operator==(const TufTarget &lhr) const {
+    return lhr.name_ == name_ && lhr.sha256_ == sha256_ && lhr.version_ == version_;
+  }
 
  private:
   std::string name_;
@@ -312,7 +324,7 @@ class AkliteClient {
    * @brief CompleteInstallation TODO
    * @return
    */
-  bool CompleteInstallation();
+  InstallResult CompleteInstallation();
 
   /**
    * Default files/paths to search for sota toml when configuration client.
--- a/include/aktualizr-lite/api.h
+++ b/include/aktualizr-lite/api.h
@@ -14,6 +14,22 @@
 
 class Config;
 class LiteClient;
+class AkliteClient;
+
+namespace cli {
+
+enum class ExitCode {
+  UnknownError = -1,
+  Ok = 0,
+  InstallationInProgress = 10,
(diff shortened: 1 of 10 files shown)


mike-sul:
@doanac This is the CLI API I came up with after playing with a few other options/approaches.

The key points of these approach are:

1) Abstract the CLI API functions (install, complete install/finalize) away from AkliteClient instantiation. As result, the implementation is agnostic of the way an AkliteClient instance is created and a client is free to choose a way to create AkliteClient. It provides both immediate benefits (easier to integrate into the current main.cc as well as into the unit test framework) and long-term - I believe we will need to extend the API with another AkliteClient ctors/factory functions in the future.

2) Don't do "finalization" in AkliteClient's ctor by default. It allows to implement a more error proof and controllable cli client (e.g. typing the same command twice, determining whether finalization actually took place and its status.)

include/aktualizr-lite/api.h
```diff
@@ -14,13 +14,34 @@
 
 class Config;
 class LiteClient;
+class AkliteClient;
+
+namespace cli {
+
+enum class ExitCode {
+  UnknonwError = 1,  // TODO
```
StealthyCoder, line 22: Small typo, `UnknownError` instead of `UnknonwError` :+1:

src/cli.cc
```diff
@@ -0,0 +1,68 @@
...
+    LOG_INFO << "Cannot install because there is another installation in progress: " << target.Name();
+    return ExitCode::InstallationInProgress;
+  }
+
+  const auto targets = client.CheckIn();
+  if (version == -1) {
+    target = targets.GetLatest();
+  } else {
+    for (const auto& t : targets.Targets()) {
+      if (t.Version() == version) {
```
vkhoroz, line 26: I guess we should also check for the hardware ID?

src/api.cc
```diff
@@ -362,4 +362,15 @@ TufTarget AkliteClient::GetPendingTarget() const {
   return !pending ? TufTarget() : Target::toTufTarget(*pending);
 }
 
-bool AkliteClient::CompleteInstallation() { return client_->finalizeInstall(); }
+InstallResult AkliteClient::CompleteInstallation() {
+  auto install_completed{client_->finalizeInstall()};
+  InstallResult complete_install_res{InstallResult::Status::Failed, ""};
+  if (install_completed) {
+    if (!client_->isBootFwUpdateInProgress()) {
```
vkhoroz, line 369: I cannot comment on commit message.

Need to add `- Handle boot firmware update` to it.

src/cli.cc
```diff
@@ -59,19 +76,33 @@ ExitCode Install(AkliteClient& client, int version) {
...
+    }
+    case InstallResult::Status::NeedsCompletion: {
+      // Update of sotree and Apps have been completed successfully;
+      // bootloader was updated too and it requires device reboot to confirm its update.
+      exit_code = ExitCode::InstallNeedsRebootForBootFw;
+      break;
+    }
+    default:
+      // TODO
+      break;
```
vkhoroz, line 138: I don't get why is there a `// TODO` comment in all the default switch clauses.
Shouldn't we core dump or return `UnknownError` in these cases?

src/cli.cc
```diff
@@ -96,6 +109,30 @@ ExitCode CompleteInstall(AkliteClient& client) {
       exit_code = ExitCode::InstallNeedsRebootForBootFw;
       break;
     }
+    case InstallResult::Status::Failed: {
+      // do rollback
+      // check rollback type, bootloader (boot on the previous version) or App driven
```
vkhoroz, line 114: Not sure if my question makes sense:
- should we ask a user if they want to retry the installation or rollback in case of failure? Maybe, an argument for that?

src/cli.cc
```diff
@@ -96,6 +109,30 @@ ExitCode CompleteInstall(AkliteClient& client) {
...
+        // ostree rollback
+        auto rollback_installer = client.CheckAppsInSync();
+        if (rollback_installer) {
+          auto rollback_download_res = rollback_installer->Download();
+          // TODO
+          auto rollback_install_res = rollback_installer->Install();
+          if (rollback_install_res.status == InstallResult::Status::Ok) {
+            exit_code = ExitCode::InstallRollbackOk;
+          } else {
+            exit_code = ExitCode::InstallRollbackFailed;
```
vkhoroz, line 126: Yet another "hard" question:
- what is the user expected to do in case if both installation and rollback fail?

vkhoroz (commented):
Good work!

I defer this to @doanac ; just found a few spots.

src/api.cc
mike-sul: I just accidentally merged two commit during rebase. Anyway, this is draft PR atm, and there are still a lot of to do, so I will restructure commits anyway.

src/cli.cc
mike-sul: It does. I don't know. I would ask for feedback at first.

mike-sul:
> i forget the details of the sprint, but it seems like it would be a good time to split up the download and install steps. That's the main thing people are trying to manage - when to download and when to install.

Makes sense to me. However, it most likely will lead to further requests to amend/extend the api (e.g. remove specific version of the downloaded target). Having said that, it looks like the right approach to discover the api customers need.

src/api.cc
```diff
@@ -351,3 +355,11 @@ InstallResult AkliteClient::SetSecondaries(const std::vector<SecondaryEcu>& ecus
   secondary_hwids_ = std::move(hwids);
   return InstallResult{InstallResult::Status::Ok, ""};
 }
+
+TufTarget AkliteClient::GetPendingTarget() const {
+  boost::optional<Uptane::Target> pending;
+  client_->storage->loadInstalledVersions("", nullptr, &pending);
+  return !pending ? TufTarget() : Target::toTufTarget(*pending);
+}
```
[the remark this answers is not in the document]
mike-sul: I would say "broken" is too strong adjective for it. It depends on a custom client specifics. For example, a custom client can have an internal state to track an update state so it prevents from calling api::Download/Install if there is pending update. Or, a custom client simply wants enforce a new update regardless if there is pending update or not (the pending update will be overwritten).

src/cli.cc
```diff
@@ -0,0 +1,68 @@
...
+namespace cli {
+
+ExitCode Install(AkliteClient& client, int version) {
+  ExitCode exit_code{ExitCode::UnknownError};
+
+  TufTarget target{client.GetPendingTarget()};
+  if (!target.IsUnknown()) {
+    LOG_INFO << "Cannot install because there is another installation in progress: " << target.Name();
+    return ExitCode::InstallationInProgress;
+  }
```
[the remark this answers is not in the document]
mike-sul: It doesn't provides details of the pending target what can be useful for end user.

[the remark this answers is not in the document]
mike-sul: Maybe we need both, or merge them into a single call that returns something like `std::tuple<bool,Target>`, or `bool IsInstallationInProgress(Target* pending_target = nullptr)` (return pending target if there is on and the param is not null).

src/cli.cc
```diff
@@ -13,18 +13,29 @@ ExitCode Install(AkliteClient& client, int version) {
...
-    for (const auto& t : targets.Targets()) {
+    for (const auto& t : get_tuf_res.Targets()) {
       if (t.Version() == version) {
         target = t;
         break;
       }
     }
   }
 
+  if (target.IsUnknown()) {
```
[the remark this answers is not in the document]
mike-sul: Yep. The "unknown" is kind of iterator::end() is the classical  find in the stl containers.

src/cli.cc
```diff
@@ -96,6 +109,30 @@ ExitCode CompleteInstall(AkliteClient& client) {
       exit_code = ExitCode::InstallNeedsRebootForBootFw;
       break;
     }
+    case InstallResult::Status::Failed: {
+      // do rollback
+      // check rollback type, bootloader (boot on the previous version) or App driven
+      const auto current{client.GetCurrent()};  // returns Target a device is booted on
```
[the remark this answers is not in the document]
mike-sul: No. The `GetRollbackTarget` works only for the "app driven" rollback.  If this is "ostree" rollback (the bootloader did rollback and a device booted on the previous target) then we need just sync "current" since the "current" is always bound to the ostree version a device is booted  on.

What I've been considering, while implementing it, is to hide the difference between the "app driver" and "ostree" rollback behind `GetRollbackTarget`. Specifically, its implementation could return the rollback target in the first case and the current one in the second case.

doanac:
> However, it most likely will lead to further requests to amend/extend the api 

Yep. I'd also made the API work in a friendly, hard-to-mess-up way where a check-in gave you the downloader and the downloader gave you the installer. This may bring us back to the wild west of hard-to-use apis.

src/api.cc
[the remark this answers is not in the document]
doanac: maybe we should keep them consistent?

PR #240, opened 2023-06-07
Skopeo parallel pull control

mike-sul:
Override the default value of the skopeo's maximum parallel/concurrent pulls if the env. var. `SKOPEO_MAX_PARALLEL_PULLS` is set to a value in the allowed range [1, 10].

If `SKOPEO_MAX_PARALLEL_PULLS` is not defined, then `skopeo` will apply the default value which is `3` for LmP and `6` for other use-cases.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -8,6 +8,7 @@
 #include <boost/algorithm/hex.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/format.hpp>
+#include <boost/lexical_cast.hpp>
 #include <boost/process.hpp>
 
 #include "crypto/crypto.h"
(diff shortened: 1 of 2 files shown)


src/docker/restorableappengine.cc
```diff
@@ -761,11 +781,18 @@ bool RestorableAppEngine::checkAppContainers(const App& app, const std::string&
...
 void RestorableAppEngine::pullImage(const std::string& client, const std::string& src,
                                     const boost::filesystem::path& dst_dir,
-                                    const boost::filesystem::path& shared_blob_dir, const std::string& format) {
+                                    const boost::filesystem::path& shared_blob_dir, int max_parallel_pulls,
+                                    const std::string& format) {
   boost::filesystem::create_directories(dst_dir);
-  exec(boost::format{"%s copy -f %s --dest-shared-blob-dir %s %s oci:%s"} % client % format % shared_blob_dir.string() %
-           src % dst_dir.string(),
-       "failed to pull image", boost::this_process::environment());
+  if (-1 == max_parallel_pulls) {
```
doanac, line 787: should we just enforce your actual range and then not need to keep this if/else logic?

mike-sul: Only LmP has the `skopeo` version that supports `--max-parallel-pulls` param, so we cannot run `skopeo copy` unconditionally. (e.g. running unit tests in the container or running aklite on a local host).

src/docker/restorableappengine.cc
```diff
@@ -76,6 +77,25 @@ RestorableAppEngine::RestorableAppEngine(boost::filesystem::path store_root, boo
...
+  if (const char* max_par_pulls_str = std::getenv("SKOPEO_MAX_PARALLEL_PULLS")) {
+    try {
+      max_parallel_pulls_ = boost::lexical_cast<int>(max_par_pulls_str);
+      if (max_parallel_pulls_ > SkopeoMaxParallelPullsHighLimit) {
+        LOG_DEBUG << "Value of SKOPEO_MAX_PARALLEL_PULLS env variable exceeds the maximum allowed; value: "
+                  << max_par_pulls_str << "; the maximum allowed: " << SkopeoMaxParallelPullsHighLimit;
+        max_parallel_pulls_ = SkopeoMaxParallelPullsHighLimit;
+      }
+      if (max_parallel_pulls_ < SkopeoMaxParallelPullsLowLimit) {
+        LOG_DEBUG << "Value of SKOPEO_MAX_PARALLEL_PULLS env variable is lower than the minimum allowed; value: "
```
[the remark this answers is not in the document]
mike-sul: Updated with two warning and one error message.

doanac (approved):

PR #245, opened 2023-06-19
cli: Add logs for the rollback after boot cases

mike-sul:
Add meaningful logs for the rollback cases.

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -32,7 +32,7 @@ static const std::unordered_map<InstallResult::Status, StatusCode> i2s = {
 };
 
 StatusCode Install(AkliteClient &client, int version, const std::string &target_name) {
-  // Check if a device is in a correct state to start a new update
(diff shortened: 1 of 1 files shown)


src/cli/cli.cc
```diff
@@ -118,28 +118,41 @@ StatusCode CompleteInstall(AkliteClient &client) {
     const auto current{client.GetCurrent()};  // returns Target a device is booted on
     if (current.Sha256Hash() != pending.Sha256Hash()) {
       // ostree rollback, aka the bootloader driven rollback
+      LOG_INFO << "Installation has failed, a device was rolled back to " << current.Name();
```
doanac, line 121: I don't think "a device" makes sense in this context - its either just "device" or "the device".

mike-sul: fixed.

mike-sul: @doanac I think it's ready to be merged.

doanac (approved):

PR #246, opened 2023-06-20
Move to openssl v3

mike-sul:
Move to openssl 3.0.2.

### DIFF
--- a/tests/device-gateway_fake.py
+++ b/tests/device-gateway_fake.py
@@ -13,6 +13,7 @@ logger = logging.getLogger("Fake Device Gateway")
 
 
 class Handler(SimpleHTTPRequestHandler):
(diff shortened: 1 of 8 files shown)


tests/device-gateway_fake.py
```diff
@@ -115,20 +122,24 @@ def registry_handler(self):
 
     def tuf_handler(self):
         logger.info("TUF: GET request %s" % self.path)
+        print(self.path)
```
doanac, line 125: leftover debug?

doanac (approved):
just one nit

tests/device-gateway_fake.py
mike-sul: good catch.

PR #248, opened 2023-06-23
composeappmanager: Stop Apps just before updating

mike-sul:
It makes sense to stop the apps that are about to be updated just before the ostree installation. If a device unexpectedly turns off immediately after a successful ostree installation but before the apps are updated, then this change ensures that the apps of the previous version are not automatically started upon boot.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -315,6 +315,27 @@ TargetStatus ComposeAppManager::verifyTarget(const Uptane::Target& target) const
 }
 
 data::InstallationResult ComposeAppManager::install(const Uptane::Target& target) const {
+  // Stopping disabled apps before creating or starting new apps
+  // because they may interfere with each other (e.g., using the same port).
+  // It is advisable to stop the disabled apps before performing ostree installation.
+  // Otherwise, they will be automatically started by dockerd during boot if the device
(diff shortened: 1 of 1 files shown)


doanac (commented):
I think this makes sense, but does this introduce a situation where a customer has a critical compose-app (perhaps it helps drive install logic) that wouldn't start upon the reboot? I guess aklite will still start up and try to do the right thing (if they use aklite)?

doanac:
Might want @Tim-Anderson and @mike-scott to take a look also

mike-sul:
@Tim-Anderson @mike-scott Do you think stopping Apps just before their update is going to break something? 

I think this change shouldn't introduce any regression since `docker compose up` does stop App containers anyway, this change just enforces it slightly earlier, before ostree update.

mike-scott:
@mike-sul I don't see any reason we can't stop disabled compose-apps prior to ostree updates.  @Tim-Anderson what do you think?

Tim-Anderson:
as long as the files are downloaded, I am fine with docker compose down, docker compose up.

Tim-Anderson:
It can still take some time for the copy from reset-apps to docker maybe we should time that

mike-sul:
@doanac Please, approve, I can merge it without at least one approval :).

PR #251, opened 2023-07-12
apps: Calculate App update size precisely

mike-sul:
This change enables aklite to utilize App layers metadata that contains precise disk usage information for each layer. As a result, aklite can calculate the App update size with a high level of accuracy. There are additional metadata that the docker store utilizes for each layer size/usage that are not considered by aklite. The overall size of these metadata does not depend on the layer size and is not significant.

If the App layers metadata is not found, aklite will fall back to an approximate calculation of the App update size.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -456,32 +456,59 @@ void RestorableAppEngine::checkAppUpdateSize(const Uri& uri, const boost::filesy
     return;
   }
 
-  const auto layers_manifest{manifest.layersManifest(arch)};
-  if (!layers_manifest.isObject()) {
-    LOG_WARNING << "App layers' manifest is missing, skip checking an App update size";
-    return;
+  uint64_t skopeo_total_update_size;
+  uint64_t docker_total_update_size;
+  bool fallback_to_estimated_update_size_calculation{true};
+
+  const auto layers_meta_desc{manifest.layersMetaDescr()};
+  if (layers_meta_desc) {
+    try {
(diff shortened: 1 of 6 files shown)


mike-sul:
@doanac This is the aklite part that goes together with:

https://github.com/foundriesio/ci-scripts/pull/285
https://github.com/foundriesio/compose-publish/pull/12

src/docker/restorableappengine.cc
```diff
@@ -893,6 +918,65 @@ uint64_t RestorableAppEngine::getDockerStoreSizeForAppUpdate(const uint64_t& com
...
+  uint64_t docker_total_update_size{0};
+
+  for (Json::ValueConstIterator ii = app_layers.begin(); ii != app_layers.end(); ++ii) {
+    const HashedDigest digest{ii.key().asString()};
+    if (store_blobs.count(digest.hash()) == 1) {
+      LOG_INFO << "\t" << digest.hash() << " -> exists";
+      continue;
+    }
+    if (!(*ii).isMember("usage")) {
+      throw std::range_error("Inavlid layers metadat; `usage` field is missing: " + (*ii).asString());
```
doanac, line 949:
```suggestion
      throw std::range_error("Invalid layers metadata; `usage` field is missing: " + (*ii).asString());
```

mike-sul: fixed

doanac:
> @doanac Did you have a chance to review this PR?

I thought you were on holiday so i was ignoring :)

doanac (approved):

PR #252, opened 2023-07-20
Improve app storing and parsing

mike-sul:
I was informed about the following error a few times:

```
Jul 11 20:01:04 stm32mp15-disco-3d00383338510b39303435 aktualizr-lite[920]: Checking Active Target status...
Jul 11 20:01:04 stm32mp15-disco-3d00383338510b39303435 aktualizr-lite[920]: App: shellhttpd, cannot check whether App is running: * Line 1, Column 1
Jul 11 20:01:04 stm32mp15-disco-3d00383338510b39303435 aktualizr-lite[920]:   Syntax error: value, object or array expected.
Jul 11 20:01:04 stm32mp15-disco-3d00383338510b39303435 aktualizr-lite[920]: shellhttpd update will be re-installed or completed
```

After investigation, I came to conclusion that that error message can appear only if `docker-compose.yml` is empty and produced by the `yaml2json` component. The `fy-tool` doesn't produce error if an input `yaml` file is empty, but the following conversion to json fails and prints the reported error message.

This change, improves error/edge cases handling in the `yaml2json` so meaningful error messages are output, as result we can understand what's going on.
Also, the question is: how come `docker-compose.yml` is empty? The only explanation is that the extraction call is not exactly save in terms of ensuring file persistence. So, this PR changes the compose file extraction logic to make sure it's safe and really persists the compose.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -144,17 +144,10 @@ AppEngine::Result RestorableAppEngine::verify(const App& app) {
   try {
     const Uri uri{Uri::parseUri(app.uri)};
     const auto app_dir{apps_root_ / uri.app / uri.digest.hash()};
-    const Manifest manifest{Utils::parseJSONFile(app_dir / Manifest::Filename)};
-    const auto archive_full_path{app_dir / (HashedDigest(manifest.archiveDigest()).hash() + Manifest::ArchiveExt)};
-    TemporaryDirectory app_tmp_dir;
 
-    // extract the compose file (i.e. docker-compose.yml) to a temporary directory to verify it.
-    exec(boost::format{"tar -xzf %s %s"} % archive_full_path % ComposeFile, "no compose file found in archive",
-         boost::process::start_dir = app_tmp_dir.Path());
-
-    LOG_DEBUG << app.name << ": verifying App: " << app_dir << " --> " << app_tmp_dir.Path();
+    LOG_DEBUG << app.name << ": verifying App: " << app_dir << " --> " << app_dir;
     exec(boost::format("%s config %s") % compose_cmd_ % "-q", "compose file verification failed",
-         boost::process::start_dir = app_tmp_dir.Path());
+         boost::process::start_dir = app_dir);
   } catch (const std::exception& exc) {
     LOG_ERROR << "failed to verify App; app: " + app.name + "; uri: " + app.uri + "; err: " + exc.what();
     res = {false, exc.what()};
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -443,9 +443,9 @@ void RestorableAppEngine::pullApp(const Uri& uri, const boost::filesystem::path&
   registry_client_->downloadBlob(archive_uri, archive_full_path, manifest.archiveSize());
   Utils::writeFile(app_dir / Manifest::Filename, manifest_str);
   Utils::writeFile(app_dir / "uri", uri.registryHostname + "/" + uri.repo + "@" + uri.digest());
-  // extract docker-compose.yml, temporal hack, we don't need to extract it
-  exec(boost::format{"tar -xzf %s %s"} % archive_full_path % ComposeFile, "no compose file found in archive",
-       boost::process::start_dir = app_dir);
+  // Extract docker-compose.yml and safely persist it so the follow-up functionality doesn't need to do it again.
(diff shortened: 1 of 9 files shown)


vkhoroz:
LGTM. Good job!

src/docker/restorableappengine.cc
```diff
@@ -443,9 +443,9 @@ void RestorableAppEngine::pullApp(const Uri& uri, const boost::filesystem::path&
   registry_client_->downloadBlob(archive_uri, archive_full_path, manifest.archiveSize());
   Utils::writeFile(app_dir / Manifest::Filename, manifest_str);
   Utils::writeFile(app_dir / "uri", uri.registryHostname + "/" + uri.repo + "@" + uri.digest());
-  // extract docker-compose.yml, temporal hack, we don't need to extract it
-  exec(boost::format{"tar -xzf %s %s"} % archive_full_path % ComposeFile, "no compose file found in archive",
-       boost::process::start_dir = app_dir);
+  // Extract docker-compose.yml and safely persist it so the follow-up functionality doesn't need to do it again.
+  const auto compose{extractComposeFile(archive_full_path)};
+  Utils::writeFile(app_dir / ComposeFile, compose);
```
doanac, line 441: orthogonal - but i noticed a bug in `writeFile` - https://github.com/foundriesio/aktualizr/blob/2023.06%2Bfio/src/libaktualizr/utilities/utils.cc#L337-L339 there needs to be a call to `sync` between the file close (or if the ostream could call sync itself) and the rename or rename could wind up pointing at an inode with no data

doanac (approved):

src/docker/restorableappengine.cc
mike-sul: I think file.close() includes sync.

vkhoroz: That said, I'm not sure how critical it is for us.

The `fsync` call can still result into another _buffering layer_ inside a disk cache; but POSIX compliant disk manufacturers must flush that cache in case of fsync.

At the very least `fsync` call would warrant that all readers will see the "updated" version of the file.

doanac: the fsync is required for power safe operations. Its a very unlikely thing, but its the only way. `rename` is an atomic call - so it will persist across a weird power fail. but if you haven't done the fsync - it can point at garbage.

mike-sul: >The file.close includes a sync, but not fsync.

That's correct. I just don't know how really is important to do it (fsync), and what would be performance cost.

mike-sul: Also, it's not really critical as the aklite will just re-extract a compose file from an App archive.

vkhoroz: One scenario I was interested in is this:
1. Application opens a file, writes to it and calls close (which calls sync).
2. Application opens a file and reads from it before the OS actually syncs the file to disk.

I'm not sure if OS gives any warranty that at step 2 application will see the writes from step 1.
Normally, it should be the case; but here we're talking about esoteric OSes on low-profile devices.
(The file rename is irrelevant here, as it will point to the same node; I'm only talking about the write to the node itself).

The fsync would give such warranties; the close - I'm not sure.

doanac: you don't need the fsync for that scenario - the OS will handle that. `fsync` handles one specific thing:
1) write file (but content is still not flushed to disk)
2) rename file (this is okay - the os has these buffers)
3) power fails before buffers are flushed to disk

When the system starts back up step 2 will be done (there will be a file with that name) but the inodes it's pointing at won't have correct data.

`writeFile` may be "safe enough" for us - depends on what's using it - but its not safe to power failures.

mike-sul: Normally, sync/close guarantees that other processes read the same content that has been written before close/sync. It should be handled at a file system level/implementation.

It doesn't guarantee that the written content is persisted on disk if a power cut happens after sync/close. That's what `fsync` is needed for - it writes down the cached data at a file system level to a disk.

mike-sul:
Looks like I have to dig into the test failure that occurs only in the github CI.

mike-sul:
I've just added a fix/improvement to the test framework to ensure that events are reset at the correct moment at the device gateway mock. https://github.com/foundriesio/aktualizr-lite/pull/252/commits/6b184d7381f5d4355c1a399a8d545db7212f0e96.

This last commit doesn't change aklite's functionality so I am merging this PR as it has been approved.

PR #258, opened 2023-08-08
Update aktualizr and add test-uptane-vectors make target

detsch:


Makefile
```diff
@@ -11,11 +11,14 @@ PKCS11_ENGINE_PATH ?= "/usr/lib/x86_64-linux-gnu/engines-3/pkcs11.so"
 
 
 TASKS = config build format tidy garage-tools custom-client
-.PHONY: $(TASKS) test
+.PHONY: $(TASKS) test test-uptane-vectors
```
mike-sul, line 14: I wonder if `all $(TASKS)` would work for if `test-uptane-vectors` is added to the list of `TASKS`?

mike-sul (approved):

PR #259, opened 2023-08-14
WIP: Use delta stats before pull

mike-sul:
- Check if static delta statistic is available before pulling an ostree commit/update.
- If the delta stat for the given update is available then check if there is enough free storage available to accommodate it.
- The check takes into account the storage watermark set in the configuration.
- Add unit tests for the new functionality.


src/rootfstreemanager.cc
```diff
@@ -250,3 +263,134 @@ data::InstallationResult RootfsTreeManager::verifyBootloaderUpdate(const Uptane:
...
+    LOG_ERROR << "Failed to fetch static delta stats; status: " << resp.getStatusStr() << ", err: " << resp.body;
+    return Json::nullValue;
+  }
+  if (resp.body.size() != ref.size) {
+    LOG_ERROR << "Fetched invalid static delta stats, size mismatch; "
+              << " expected: " << ref.size << ", got: " << resp.body.size();
+    return Json::nullValue;
+  }
+  const auto received_data_hash{
+      boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest(resp.body)))};
```
doanac, line 347: so we require canonicaljson as our input, correct?

doanac:
i sniffed around and things seem sane

src/rootfstreemanager.cc
mike-sul: Yes.

mike-sul:
Closing this PR in favor of https://github.com/foundriesio/aktualizr-lite/pull/262

PR #261, opened 2023-08-16
Improvements and cleanup related to the static delta use-case.

mike-sul:
This is an improvement/cleanup and an extension of the static delta-related test.
This change is a prerequisite for the follow-up change related to the application of delta stats/size.


mike-sul:
@detsch Just FYI, you may wanna check it to familiarize yourself with the some other parts/aspects of aklite.

doanac (approved):

PR #262, opened 2023-08-16
Delta stat application: check update size before fetching update data

mike-sul:
- Pull delta statistics for a given update if available.
  
- If available, then check if there is enough free storage available
  taking into account the storage watermark.
    
- Don't pull update/ostree if there is not enough free storage available
  and exit with the corresponding error.
    
- Add unit tests for the given use-case.

### DIFF
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -14,10 +14,34 @@
 #include "storage/invstorage.h"
 #include "target.h"
 
+const unsigned int RootfsTreeManager::Config::DefaultSysrootStorageWatermark{90};
+const unsigned int RootfsTreeManager::Config::MinSysrootStorageWatermark{50};
+const unsigned int RootfsTreeManager::Config::MaxSysrootStorageWatermark{95};
+
 RootfsTreeManager::Config::Config(const PackageConfig& pconfig) {
   if (pconfig.extra.count(UpdateBlockParamName) == 1) {
     std::string val{pconfig.extra.at(UpdateBlockParamName)};
     UpdateBlock = val != "0" && val != "false";
+  }
+  if (pconfig.extra.count(SysrootStorageWatermarkParamName) == 1) {
+    const std::string val_str{pconfig.extra.at(SysrootStorageWatermarkParamName)};
+    try {
+      const auto val{boost::lexical_cast<unsigned int>(val_str)};
+      if (val < MinSysrootStorageWatermark) {
+        LOG_ERROR << "Value of `" << SysrootStorageWatermarkParamName << "` parameter is too low: " << val_str
+                  << "; setting it the minimum allowed: " << MinSysrootStorageWatermark;
+        SysrootStorageWatermark = MinSysrootStorageWatermark;
+      } else if (val > MaxSysrootStorageWatermark) {
+        LOG_ERROR << "Value of `" << SysrootStorageWatermarkParamName << "` parameter is too high: " << val_str
+                  << "; setting it the maximum allowed: " << MaxSysrootStorageWatermark;
+        SysrootStorageWatermark = MaxSysrootStorageWatermark;
+      } else {
+        SysrootStorageWatermark = val;
+      }
+    } catch (const std::exception& exc) {
+      LOG_ERROR << "Invalid value of `" << SysrootStorageWatermarkParamName << "` parameter: " << val_str
+                << "; setting it the default value: " << DefaultSysrootStorageWatermark;
+    }
   }
 }
 
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -14,6 +14,13 @@
 #include "storage/invstorage.h"
 #include "target.h"
 
+RootfsTreeManager::Config::Config(const PackageConfig& pconfig) {
+  if (pconfig.extra.count(UpdateBlockParamName) == 1) {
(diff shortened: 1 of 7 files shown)


mike-sul:
Happy path output
```
Found reference to a file with static delta stats, downloading it...
Fetching delta stats -> http://localhost:34095/treehub/delta-stats/aacf24d189c887f16c89f87ac87ae9311e6f898fd5c18d9c20ca1cc09c93cc55
File with static delta stats has been downloaded, parsing it...
Found and pulled delta stats, checking if update can fit on a disk...
Fetching static delta; required 42504, available 53248 out of 368640(90% of the volume capacity 409600)
ostree-pull: 1 delta parts, 3 loose fetched; 42 KiB transferred in 0 seconds; 0 bytes content written
Checking for Apps to be installed or updated...
```

Not enough space to pull delta:
```
Found reference to a file with static delta stats, downloading it...
Fetching delta stats -> http://localhost:34095/treehub/delta-stats/aacf24d189c887f16c89f87ac87ae9311e6f898fd5c18d9c20ca1cc09c93cc55
File with static delta stats has been downloaded, parsing it...
Found and pulled delta stats, checking if update can fit on a disk...
Download unsuccessful after 1 attempts; err: Insufficient storage available; err: required 42504, available 36864 out of 368640(90% of the volume capacity 409600)
```

Enough space to pull delta but the ostree commit wants more:
```
Found reference to a file with static delta stats, downloading it...
Fetching delta stats -> http://localhost:34095/treehub/delta-stats/aacf24d189c887f16c89f87ac87ae9311e6f898fd5c18d9c20ca1cc09c93cc55
File with static delta stats has been downloaded, parsing it...
Found and pulled delta stats, checking if update can fit on a disk...
Fetching static delta; required 42504, available 45056 out of 368640(90% of the volume capacity 409600)
Error while pulling image: 0 opcode open-splice-and-close: Writing content object: min-free-space-percent '3%' would be exceeded, at least 4.1 kB requested
Failed to fetch from http://localhost:34095/treehub, err: opcode open-splice-and-close: Writing content object: min-free-space-percent '3%' would be exceeded, at least 4.1 kB requested
Download unsuccessful after 1 attempts; err: Insufficient storage available; path: /tmp/aktualizr-2ded-3fbf-192d-ddd6/5883-15eb-dir/sysrepo; err: opcode open-splice-and-close: Writing content object: min-free-space-percent '3%' would be exceeded, at least 4.1 kB requested
```

mike-sul:
@doanac This is the implementation that is agnostic of the storage watermark source. Once we agree on the watermark configuration I'll add corresponding changes to this PR.

mike-sul:
That's odd that the same test passes locally, the container is the same so it's supposed to behave consistently...

doanac (commented):
the first n-1 changes look great. feel free to merge those now if it makes this easier. i'm reading the real change now.

src/rootfstreemanager.cc
```diff
@@ -250,3 +277,151 @@ data::InstallationResult RootfsTreeManager::verifyBootloaderUpdate(const Uptane:
...
+    LOG_ERROR << "Fetched invalid static delta stats, size mismatch; "
+              << " expected: " << ref.size << ", got: " << resp.body.size();
+    return Json::nullValue;
+  }
+  const auto received_data_hash{
+      boost::algorithm::to_lower_copy(boost::algorithm::hex(Crypto::sha256digest(resp.body)))};
+  if (received_data_hash != ref.sha256) {
+    LOG_ERROR << "Fetched invalid static delta stats, hash mismatch; "
+              << " expected: " << ref.sha256 << ", got: " << received_data_hash;
+    return Json::nullValue;
```
doanac, line 378: I guess the right behavior is to "go ahead and try". It almost feels like having an invalid file might ought to create an error though.

src/rootfstreemanager.cc
```diff
@@ -250,3 +277,151 @@ data::InstallationResult RootfsTreeManager::verifyBootloaderUpdate(const Uptane:
...
+    LOG_ERROR << "Incorrect metadata about static delta statistics are found in Target;"
+              << " err: missing `size` field or it's not an integer";
+    return false;
+  }
+  ref = {delta_stats_ref["sha256"].asString(), delta_stats_ref["size"].asUInt()};
+  return true;
+}
+
+Json::Value RootfsTreeManager::downloadDeltaStats(const DeltaStatsRef& ref, const Remote& remote) {
+  static const uint64_t DeltaStatsMaxSize{1024 * 1024};
```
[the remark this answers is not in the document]
mike-sul: Since, it's used solely within this method then it should be fine to define it here. That said, I don't have a strong opinion on it.

src/rootfstreemanager.cc
mike-sul: If the stat file is invalid then we assume that the delta itself is invalid?
If the delta wire size is used during delta fetch then I think this error is critical.
Since there is no any relation between the stat file content and the delta pull process I don't have a strong opinion on it.
I would rather consider this error as non-critical and continue with update to make it consistent with the current implementation.

mike-sul:
@doanac I think that's enough changes for this PR.
Application of the new param to the ostree config will be in the next PR (core.min-free-space-percent = (100 - sysroot_storage_watermark) ).

doanac (approved):

PR #266, opened 2023-09-04
Unify and improve free space verification

mike-sul:
- Add functionality to obtain storage usage that can be used across all use-cases.
- Apply the new storage usage obtaining functionality to all ostree pull use-cases, with and without delta, negative and positive cases.
- Extend and improve the unit tests to verify the new and modified functionality.

### DIFF
--- a/src/storage/stat.cc
+++ b/src/storage/stat.cc
@@ -92,6 +92,7 @@ ostream& operator<<(ostream& os, const storage::Volume::UsageInfo& t) {
     if (t.required.first > 0) {
       os << "required: " << t.required.first << "B unknown%";
     }
+    os << "; fstatvfs err: " << t.err;
     return os;
   }
 
--- a/src/storage/stat.cc
+++ b/src/storage/stat.cc
@@ -1,0 +1,106 @@
+#include "stat.h"
+
+#include <fcntl.h>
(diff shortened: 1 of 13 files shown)


src/storage/stat.cc
```diff
@@ -0,0 +1,106 @@
...
+  }
+  struct statvfs fs_stat {};
+  const auto stat_res = fstatvfs(fd, &fs_stat);
+  close(fd);
+  if (-1 == stat_res) {
+    return std::string("Failed to obtain statistic about the path volume; path: ") + path +
+           ", err: " + std::strerror(errno);
+  }
+  stat.blockNumb = fs_stat.f_blocks;
+  stat.freeBlockNumber = (getuid() != 0 ? fs_stat.f_bavail : fs_stat.f_bfree);
```
doanac, line 33: is this for unit tests? Don't think we can run as non-root? maybe this could use a comment.

doanac (approved):
Seems like a nice clean up. I didn't think to much on the free space math. You might want Andre to take a look

src/storage/stat.cc
mike-sul: Yes. This makes code user permission agnostic, actually I copy-pasted it from the libostree :).

PR #268, opened 2023-09-06
Apply usage stat to apps

mike-sul:
Apply the new storage usage functionality to the Apps pulling:
  1. Use it in the pre-pull check.
  2. Print logs before and after pull with storage usage info.
  3. Add the usage statistic to the download complete event.


src/docker/restorableappengine.cc
```diff
@@ -418,18 +414,14 @@ void RestorableAppEngine::pullApp(const Uri& uri, const boost::filesystem::path&
   Docker::Uri archive_uri{uri.createUri(HashedDigest(manifest.archiveDigest()))};
   const auto archive_full_path{app_dir / (HashedDigest(manifest.archiveDigest()).hash() + Manifest::ArchiveExt)};
 
-  boost::system::error_code ec;
-  boost::filesystem::space_info fs_storage_info{boost::filesystem::space(app_dir, ec)};
-  if (ec.failed()) {
-    LOG_WARNING << "Failed to get an available storage size: " << ec.message();
+  storage::Volume::UsageInfo usage_info{storage_space_func_(app_dir)};
+  if (!usage_info.isOk()) {
+    LOG_ERROR << "Failed to obtain storage usage statistic: " << usage_info.err;
```
doanac, line 419: Are we logging this same message twice?

src/docker/restorableappengine.cc
```diff
@@ -963,15 +955,11 @@ void RestorableAppEngine::checkAvailableStorageInStores(const std::string& app_n
...
-
-    std::tie(capacity, available) = storage_space_func_(store_path);
-    LOG_INFO << app_name << " -> " << store_name << " store total update size: " << required_storage
-             << " bytes; available: " << available << " (out of " << capacity << ") "
-             << ", path: " << store_path.string();
-    if (required_storage > available) {
-      throw InsufficientSpaceError(store_name, store_path.string(), required_storage, available);
+    storage::Volume::UsageInfo usage_info{storage_space_func_(store_path)};
+    LOG_INFO << app_name << " -> " << store_name
+             << " store total update size: " << usage_info.withRequired(required_storage);
```
doanac, line 960: don't you need a `if usage_info.isOk` before printing printing this?

doanac (commented):
Nothing major. One thing that's starting to creep up to me is the number of times we seem to be calling functions to get the usageinfo. Are there places we call it twice just because the code structure to bubble up the original info isn't possible?

src/docker/restorableappengine.cc
mike-sul: No. I've done intentionally. The `operator <<` for `UsageInfo ` has `if` if it's not ok, and will print an error. I find it helpful, since a user can see problems with obtaining usage statistics.

mike-sul:
> Nothing major. One thing that's starting to creep up to me is the number of times we seem to be calling functions to get the usageinfo. Are there places we call it twice just because the code structure to bubble up the original info isn't possible?

I am not sure I see calling it twice for the same spot/moment of the update process. We have two more calls in addition to what we had, the pre-pull and post-pull calls. Obviously that the storage usage before and after pulling data is different, so two calls.

doanac (approved):

PR #269, opened 2023-09-07
sysroot: Remove redundant conf param

mike-sul:
Remove the knob to configure the reserved amount of storage for the delta-based update. Instead, use the `min-free-space-percent` value set in ostree repo config. This param value can be overridden in the sota toml configuration -
`[pacman]:sysroot_ostree_reserved_space_percentage`.


doanac (approved):
i can tell this is a good decision just based on how much code was just removed

PR #270, opened 2023-09-08
Don't pull ostree if no available storage

mike-sul:
- Don't pull ostree if the amount of free storage is lesser than the amount of reserved storage (no delta case).

### DIFF
--- a/src/rootfstreemanager.cc
+++ b/src/rootfstreemanager.cc
@@ -56,10 +56,6 @@ DownloadResultWithStat RootfsTreeManager::Download(const TufTarget& target) {
     bool delta_stat_avail{getDeltaStatIfAvailable(target, remote, delta_stat)};
 
     storage::Volume::UsageInfo pre_pull_usage_info{getUsageInfo()};
-    if (!pre_pull_usage_info.isOk()) {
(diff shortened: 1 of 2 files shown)


src/rootfstreemanager.cc
```diff
@@ -78,6 +74,12 @@ DownloadResultWithStat RootfsTreeManager::Download(const TufTarget& target) {
     } else {
       if (pre_pull_usage_info.isOk()) {
         LOG_INFO << "Pre-pull storage usage info; " << pre_pull_usage_info;
+        if (pre_pull_usage_info.available.first == 0) {
```
doanac, line 77: This means 0 bytes free? Should we compare to our minimum reserved storage instead?

mike-sul: No, it means `avail==0` which happens if `free <= reserved`. The math happens in `getUsageInfo` https://github.com/foundriesio/aktualizr-lite/blob/54040f33e83f2ba360f66fa51d832bfaa2d7416e/src/storage/stat.cc#L50.

There is no info/stat about the update size, so we cannot compare `pre_pull_usage_info.available` to update size at this spot as we do above if the update size is known.

doanac (approved):

PR #271, opened 2023-09-08
apps: Assume stateless app is pulled when running it

mike-sul:
It might happen that the compose app state file is missing when a call to run it is invoked. This should not happen under normal circumstances since the previous `fetch` call makes sure the state file is created and set.
However, in the case of switching from the restorable app type to compose one or manual app pull (`docker-compose pull/up` the app's compose dir ends up with no state file at the moment of its running. So, this fix assumed that the app is pulled if `run` is called and enforces the app state file setting to the `pulled` state. Even if the app compose content is missing the run command will pull it.

### DIFF
--- a/src/docker/composeappengine.cc
+++ b/src/docker/composeappengine.cc
@@ -107,6 +107,19 @@ AppEngine::Result ComposeAppEngine::run(const App& app) {
     return {false, "App dir doesn't exist, cannot install App that hasn't been fetched"};
   }
 
+  if (!isFetched(app)) {
+    // This situation can occur only if the switch from restorable to compose apps
+    // happened or App was pulled manually by running `docker-compose` command.
+    // So, the app directory with the compose content is present while the `.meta` directory
+    // and the `.meta/.version` files are missing.
+    // Therefore, we assume that App has been already pulled and set its state correspondingly.
+    LOG_WARNING << "App state is unknown, assuming it's pulled and trying to run it...";
+    // make sure the app meta dir exists
+    boost::filesystem::create_directories(appRoot(app) / MetaDir);
+    // assume that the app state is "pulled"
+    AppState(app, appRoot(app), true).setState(AppState::State::kPulled);
+  }
+
   AppState state(app, appRoot(app));
 
   // do App start if not started yet


src/docker/composeappengine.cc
```diff
@@ -107,6 +107,19 @@ AppEngine::Result ComposeAppEngine::run(const App& app) {
...
+    // This situation can occur only if the switch from restorable to compose apps
+    // happened or App was pulled manually by running `docker-compose` command.
+    // So, the app directory with the compose content is present while the `.meta` directory
+    // and the `.meta/.version` files are missing.
+    // Therefore, we assume that App has been already pulled and set its state correspondingly.
+    LOG_WARNING << "Trying to run App if it's not fetched or failure to get its status...";
+    // make sure the app meta dir exists
+    boost::filesystem::create_directories(appRoot(app) / MetaDir);
+    // assume that the app state is "pulled"
+     AppState(app, appRoot(app), true).setState(AppState::State::kPulled);
```
doanac, line 120: This indentation looks off? Guessing clang-tidy will spot that. 

Nitpick - but the `if it's` part of the log_warning message confuses me.

mike-sul: I fixed the indentation.

>Nitpick - but the if it's part of the log_warning message confuses me.
English grammar or the content.

doanac: not sure - maybe both. I think you are trying to say something like "The App appears to have either been fetched manually or an unexpected error occurred reading its status"?

mike-sul: >The App appears to have either been fetched manually

This is an assumption and one of the possible reasons.
So, I am trying to convey the state/situation not what might led to it.

The state can be either:
1. No compose content at all in the app dir
2. compose content is there, just the state file is missing.
3. compose content and the state file is there, but a failure to read it occurs.

doanac: maybe you just need to re-word the first half then `Trying to run App if it is not fetched`.

mike-sul: Maybe `App state is unknown, assuming it's pulled and trying to run it...`.
The error message about failure to read it should appear before it, the `isFetched()` call should print it.

doanac: sounds good. thanks

mike-sul: Updated.

doanac (approved):

PR #272, opened 2023-09-18
Enable TLS connections sharing

detsch:
Update aktualizr reference and update aklite to use new `HttpClientWithShare` class.


mike-sul (approved):
Nice work, thanks!

PR #275, opened 2023-10-06
rootfsmanager: Don't pull delta stat if offline update

mike-sul:
The fioctl command that downloads content for offline update does not fetch the delta stat if the one is present for a given target. Therefore, the offline update client should not attempt to fetch the delta stat file.


mike-sul:
@detsch This https://github.com/foundriesio/aktualizr-lite/pull/275/commits/63f99c9d0897ba138e2de6e60ddec93bbcde5e79 should fix the test failure issue.

detsch (approved):
Confirmed that the changes fix aklite-offline tests. Code LGTM.

doanac (approved):

PR #277, opened 2023-10-09
Control offline update downgrade

mike-sul:
Prevent accidental downgrades during offline updates by disallowing them by default.
Introduce a new Command Line Interface (CLI) option `--force/-f` to enable forced downgrades, empowering users to enforce updates even if the current version is higher than the specified one.
This change ensures a more controlled update process and enhances user flexibility.

### DIFF
--- a/apps/aklite-offline/cmds.cpp
+++ b/apps/aklite-offline/cmds.cpp
@@ -29,11 +29,12 @@ int CheckCmd::checkSrcDir(const Config& cfg_in, const boost::filesystem::path& s
   return ret_code;
 }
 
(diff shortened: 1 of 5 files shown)


mike-sul:
@detsch This commit https://github.com/foundriesio/aktualizr-lite/pull/277/commits/473be0cd8c8bc17970b0f5881dfde24f8501d1ef is not exactly related to the PR. I just didn't want to create a new PR for such a tiny fix (a user may wanna specify custom path to the configuration).

detsch (approved):
Tested and got the expected results. Code LGTM

PR #279, opened 2023-10-24
WIP: TUF API

detsch:


apps/tufctl/main.cpp
```diff
@@ -0,0 +1,50 @@
+#include "aktualizr-lite/tuf/aklite-tuf.h"
```
mike-sul, line 1: It think it doesn't make sense to name the header file as `aklite-tuf.h` since it's already under "aktualizr-lite" namespace/prefix. I would call it just `include/aktualizr-lite/tuf/tuf.h`

include/aktualizr-lite/tuf/aklite-tuf.h
```diff
@@ -0,0 +1,248 @@
+#ifndef AKLITE_TUF_API_H_
```
mike-sul, line 1: I would inline it with the path, e.g. `AKLITE_TUF_TUF_H_` -> `aktualizr-lite/tuf/tuf.h` (AKLITE is a short form of `aktualizr-lite`).

include/aktualizr-lite/tuf/aklite-tuf.h
```diff
@@ -0,0 +1,248 @@
...
+#include "storage/invstorage.h"
+#include "uptane/fetcher.h"
+#include "uptane/imagerepository.h"
+
+#include "aktualizr-lite/tuf/aklite-tuf.h"
+
+namespace aklite {
+namespace tuf {
+
+// DISCUSS: multiple fetch methods, or a single fetchRole?
```
mike-sul, line 25: I don't have a strong opinion about it. I let you decide based on your experience and gut feeling implementing the API.

include/aktualizr-lite/tuf/aklite-tuf.h
```diff
@@ -0,0 +1,248 @@
...
+// DISCUSS: return string vs return Json
```
mike-sul, line 26: I incline towards returning json since I believe that all clients would parse a string to json anyway.

include/aktualizr-lite/tuf/aklite-tuf.h
```diff
@@ -0,0 +1,248 @@
...
+    }
+  }
+
+  void fetchLatestRole(std::string* result, int64_t maxsize, Uptane::RepositoryType repo,
+                       const Uptane::Role& role) const override {
+    fetchRole(result, maxsize, repo, role, Uptane::Version());
+  }
+};
+
+// DISCUSS: What fields do we need? Explicit fields or map?
```
mike-sul, line 156: I would make it `struct` and extends it (add fields) based on needs. Just add only those fields that are needed by the current client(s).

include/aktualizr-lite/tuf/aklite-tuf.h
```diff
@@ -0,0 +1,248 @@
...
+  std::map<std::string, std::shared_ptr<TufRepoSource>> repo_sources;
+  std::shared_ptr<TufRepo> repo;
+  
+  ExternalTufInterface(std::vector<boost::filesystem::path> cfg_dirs) {
+    repo = std::make_shared<AkTufRepo>(cfg_dirs);
+  }
+
+  // DISCUSS: Each RepoSrc may be configured using a separate section in .toml file, a separate .toml file or a separate dir
+  //          Using a separate dir would reduce the changes required in libaktualizr.
+  void AddRepoSrc(std::string name, std::vector<boost::filesystem::path> cfg_dirs, std::vector<std::string> tags, std::string local_path) {
```
mike-sul, line 221: I feel like this part should be designed little bit differently.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,293 @@
...
+#include "storage/invstorage.h"
+#include "uptane/fetcher.h"
+#include "uptane/imagerepository.h"
+
+#include "aktualizr-lite/tuf/tuf.h"
+
+namespace aklite {
+namespace tuf {
+
+// TODO: consider switching to a single fetchRole
```
mike-sul, line 26: Do you still plan on adding a single `fetchRole` method?

apps/tufctl/main.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+      ++pos;
+    }
+  }
+
+  // Try individual fetch operations. sota.toml is not used
+  {
+    for (auto const& source : sources) {
+      Json::StreamWriterBuilder builder;
+      builder["indentation"] = "  ";
```
vkhoroz, line 48: How is this `builder` used? I don't see any reference of it in the code.

apps/tufctl/main.cpp
```diff
@@ -0,0 +1,94 @@
...
+        std::string uri = pos->second.get<std::string>("uri");
+        std::cout << "uri " << uri << " " << uri.rfind("\"file://", 0) << std::endl;
+
+        std::shared_ptr<aklite::tuf::TufRepoSource> source;
+        if (uri.rfind("\"file://", 0) == 0) {
+          source = std::make_shared<aklite::tuf::AkLocalTufRepoSource>(pos->first, pos->second);
+        } else {
+          source = std::make_shared<aklite::tuf::AkHttpsTufRepoSource>(pos->first, pos->second);
+        }
+        sources.push_back(source);
```
vkhoroz, line 37: Do we really need to bring this Uptane's complexity of multi-source TUF repos?
Can we simplify it to require exactly one repo in the config?

apps/tufctl/main.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  // Try individual fetch operations. sota.toml is not used
+  {
+    for (auto const& source : sources) {
+      Json::StreamWriterBuilder builder;
+      builder["indentation"] = "  ";
+      try {
+        auto json = source->fetchRoot(1);
```
vkhoroz, line 50: I'd expect a loop to fetch all root roles here, starting at the version _currently stored locally plus 1_, and ending at the _version for which we received 404_.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    if (enable_hsm && pt.count("tls_clientcert_id") > 0) {
+      config.tls.cert_source = CryptoSource::kPkcs11;
+      config.p11.tls_pkey_id = Utils::stripQuotes(pt.get<std::string>("tls_clientcert_id"));
+    } else {
+      config.tls.cert_source = CryptoSource::kFile;
+      config.import.tls_clientcert_path =
+          utils::BasedPath(Utils::stripQuotes(pt.get<std::string>("tls_clientcert_path")));
+    }
```
vkhoroz, line 71: This mTLS/HSM config handling seems not directly related to TUF.
Do you plan to extract this into some lower-level library function and reuse here?

vkhoroz (commented):
This is a good start @detsch. Nice work!

apps/tufctl/main.cpp
mike-sul: This is one of the main requirements - supporting tuf repo update from both a local and remote TUF repo (offline and online update hybrid mode). Once we add support of two source repos it doesn't make much difference to support N source repos :).

apps/tufctl/main.cpp
mike-sul: This is just a test main. That logic should be in a real client that uses the API. Also, it should support importing of the initial TUF root meta - which is essentially update from a local source TUF repo.

apps/tufctl/main.cpp
vkhoroz: I thought it is _either ... or ..._ relation, but if it is _both ... and ..._ relation, then it totally makes sense.

apps/tufctl/main.cpp
detsch: That's a leftover from when the methods returned a json object instead of a string. I'll remove the lines. In order to dump the json object, I was calling `Json::writeString(builder, json)`.

include/aktualizr-lite/tuf/tuf.h
detsch: It is related in the sense that it is used to access the server hosting the TUF metadata, being part of the "RepoSource" implementation.

mike-sul: First of all, I think `AkHttpsTufRepoSource` implementation should be moved out of the interface eventually. Then, the implementation can be improved if needed.

vkhoroz: I mean, we can reuse this same mTLS setup for other things, like reporting/eventing to the device gateway.
That's why it'd look better as a lower-level function, which is then reused by several components/apps.

mike-sul: I think it makes sense. However it can be done later if needed.

The new architecture implies that the reporting/evening will be performed by another component, the one that manages an update process, not the tuf client.

detsch:
Current configuration toml file for `tufctl` looks like this:
```
[storage]
path = "/home/detsch/projects/fio/aktualizr-lite/local-device-001/"

[source "my_default_https"]
uri = "https://ota-lite.foundries.io:8443/repo"
tls_cacert_path = "/home/detsch/projects/fio/aktualizr-lite/local-device-001/root.crt"
tls_pkey_path = "/home/detsch/projects/fio/aktualizr-lite/local-device-001/pkey.pem"
tls_clientcert_path = "/home/detsch/projects/fio/aktualizr-lite/local-device-001/client.pem"
tag = "main"

[source "my_local"]
uri = "file:///home/detsch/projects/fio/aktualizr-lite/local-device-001/test_tuf/xx/tuf"
```

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+};
+
+class TufRepo {
+ public:
+  virtual std::vector<Target> GetTargets() = 0;
+  virtual void updateMeta(std::shared_ptr<TufRepoSource> repo_src) = 0;
+};
+
+// TufRepo implementation that uses libaktualizr to provide TUF metadata handling and storage
+class AkTufRepo : public TufRepo {
```
mike-sul, line 125: I think this class definition should be out of the TUF interface/API scope. I would define it in separate header file which is located under `src` dir.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+class TufRepoSource {
+ public:
+  virtual std::string fetchRoot(int version) = 0;
+  virtual std::string fetchTimestamp() = 0;
+  virtual std::string fetchSnapshot() = 0;
+  virtual std::string fetchTargets() = 0;
+};
+
+// TufRepoSource implementation for fetching remote meta-data via https using libaktualizr
+class AkHttpsTufRepoSource : public TufRepoSource {
```
mike-sul, line 43: I would move this class definition our of the tuf interface (tuf.h).

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+
+// TufRepoSource implementation for fetching local meta-dat
+class AkLocalTufRepoSource : public TufRepoSource {
+ public:
+  class NotFoundException : public std::runtime_error {
+   public:
+    NotFoundException(const std::string& path) : std::runtime_error("Metadata hasn't been found; file path: " + path) {}
+  };
+
+  AkLocalTufRepoSource(std::string name_in, boost::property_tree::ptree pt);
```
mike-sul, line 71: Why do we need to ctors? And why do we need `boost::property_tree::ptree` to initialize the local repo source?

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+ private:
+  virtual std::string fetchFile(boost::filesystem::path meta_file_path);
+
+  std::string name;
+  boost::filesystem::path src_dir_;
+};
+
+// Wrapper around any TufRepoSource implementation to make it usable directly by libaktualizr,
+// by implementing Uptane::IMetadataFetcher interface
+class FetcherWrapper : public Uptane::IMetadataFetcher {
```
mike-sul, line 88: This is not part of the interface for sure. This is internal part of the AkHttpsTufRepoSource implementation.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+
+  void fetchLatestRole(std::string* result, int64_t maxsize, Uptane::RepositoryType repo,
+                       const Uptane::Role& role) const;
+
+ private:
+  std::shared_ptr<TufRepoSource> repo_src;
+};
+
+// Discuss: Keep custom as Json, or add explicit fields for the required values
+struct Target {
```
mike-sul, line 102: I think we should try and use this "target" type https://github.com/foundriesio/aktualizr-lite/blob/2afacef575719f83da3b99bf4c3f7226bd09a89d/include/aktualizr-lite/api.h#L22 (`TufTarget`). Specifically, move `TufTarget` from `api.h` to this `tuf.h` and add "include `tuf.h` " to `api.h`

src/tuf/tuf.cc
```diff
@@ -0,0 +1,202 @@
+#include "aktualizr-lite/tuf/tuf.h"
+
+namespace aklite {
+namespace tuf {
+
+AkHttpsTufRepoSource::AkHttpsTufRepoSource(std::string name_in, boost::property_tree::ptree& pt) { init(name_in, pt); }
```
mike-sul, line 6: I would move `*TufRepoSource` declaration and definition to dedicated h/cc files.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+}
+
+namespace aklite {
+namespace tuf {
+
+// enum class RoleEnum { Root = 0, Snapshot = 1, Targets = 2, Timestamp = 3, InvalidRole = -1 };
+// virtual std::string fetchRole(RoleEnum role, int version) = 0;
+
+// TODO: consider switching to a single fetchRole
+class TufRepoSource {
```
mike-sul, line 34: I think we can remove the `Tuf` prefix from all type names since they are under/in `tuf` namespace.

include/aktualizr-lite/api.h
```diff
@@ -15,6 +15,11 @@
 class Config;
 class LiteClient;
 
+namespace Uptane {
+class Target;
```
mike-sul, line 19: We shouldn't expose the aktualizr's internal types in the interface.

src/api.cc
```diff
@@ -119,31 +121,11 @@ AkliteClient::~AkliteClient() {
...
-    LOG_WARNING << "Unable to update latest metadata, using local copy: " << std::get<1>(rc);
-    if (!client_->checkImageMetaOffline()) {
-      LOG_ERROR << "Unable to use local copy of TUF data";
-      return CheckInResult(CheckInResult::Status::Failed, "", {});
-    }
-    status = CheckInResult::Status::OkCached;
-  }
-
+// Returns a sorted list of targets matching tags and hwid (or one of secondary_hwids)
+std::vector<TufTarget> AkliteClient::filterTargets(const std::vector<Uptane::Target> allTargets,
```
mike-sul, line 125: It looks like non-class static and internal function.

src/api.cc
```diff
@@ -167,9 +149,55 @@ CheckInResult AkliteClient::CheckIn() const {
...
+  }
+  client_->reportNetworkInfo();
+  client_->reportHwInfo();
+  client_->reportAppsState();
+
+  auto status = CheckInResult::Status::Ok;
+  Uptane::HardwareIdentifier hwidToFind(client_->config.provision.primary_ecu_hardware_id);
+
+  LOG_INFO << "Refreshing Targets metadata";
+  const auto rc = client_->updateImageMeta();
```
mike-sul, line 169: I think we can move this online checkin to the new TUF api usage too.
Also, it looks like we need to add another method to the `TufRepo` interface - `checkMeta` which can be used instead of `client_->checkImageMetaOffline`.

src/api.cc
detsch: I've converted it now to a non-class static function. Force-push coming soon.

include/aktualizr-lite/tuf/tuf.h
detsch: Not needed, but it made the constructor consistent with the one for AkHttpsTufRepoSource, which requires a few more settings than just the uri. I'll keep just the one receiving the uri, and drop the one receiving the ptree.

include/aktualizr-lite/tuf/tuf.h
detsch: Done. Moved the Ak* classes to specific h and cc files inside `src` dir

include/aktualizr-lite/tuf/tuf.h
detsch: Done

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,142 @@
...
+  void fillConfig(Config& config, boost::property_tree::ptree& pt);
+  std::string fetchRole(Uptane::Role role, int64_t maxsize, Uptane::Version version);
+
+  std::string name_;
+  boost::filesystem::path tmp_dir_path_;
+  std::shared_ptr<Uptane::IMetadataFetcher> meta_fetcher_;
+};
+
+// TufRepoSource implementation for fetching local meta-dat
+class AkLocalTufRepoSource : public TufRepoSource {
```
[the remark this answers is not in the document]
detsch: Done

include/aktualizr-lite/tuf/tuf.h
detsch: Done

src/tuf/tuf.cc
detsch: Done

include/aktualizr-lite/tuf/tuf.h
detsch: Done

include/aktualizr-lite/tuf/tuf.h
detsch: Done

include/aktualizr-lite/api.h
detsch: Fixed

src/api.cc

src/api.cc
detsch: I'll leave this changes to a new commit, yet to be created

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,105 @@
+#ifndef AKLITE_TUF_TUF_H_
+#define AKLITE_TUF_TUF_H_
+
+#include <iostream>
+#include <string>
+
+#include <boost/algorithm/string.hpp>
```
mike-sul, line 7: I think many of these headers became not needed here after the last cleaning work.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,105 @@
...
+#include <boost/program_options.hpp>
+#include <boost/program_options/variables_map.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include "crypto/keymanager.h"
+#include "http/httpclient.h"
+#include "json/json.h"
+#include "libaktualizr/config.h"
+#include "storage/invstorage.h"
+#include "uptane/fetcher.h"
```
mike-sul, line 18: The aklite specific headers shouldn't be here for sure.

src/tuf/aklocalreposource.cc
```diff
@@ -0,0 +1,30 @@
+#include "aklocalreposource.h"
+
+namespace aklite {
+namespace tuf {
+
+AkLocalRepoSource::AkLocalRepoSource(std::string name_in, std::string local_path) {
```
mike-sul, line 6: It looks like the given local repo source is not "aklite" specific. If so then `ak` prefix can be removed from its naming.

src/api.cc
mike-sul: Makes sense.

mike-sul:
Looks nice so far, keep up the good work!

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,97 @@
+#ifndef AKLITE_TUF_TUF_H_
+#define AKLITE_TUF_TUF_H_
+
+#include <iostream>
+#include <string>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
```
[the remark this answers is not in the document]
detsch: Not needed. Removed them now.

include/aktualizr-lite/tuf/tuf.h
detsch: Fixed.

src/tuf/aklocalreposource.cc
detsch: Done.

include/aktualizr-lite/tuf/tuf.h
mike-sul: @detsch What's your plan on this?

include/aktualizr-lite/api.h
```diff
@@ -221,6 +221,8 @@ class AkliteClient {
    */
   CheckInResult CheckIn() const;
 
+  CheckInResult CheckInLocal(const std::string path) const;
```
mike-sul, line 224: Please add a unit test for this new check in method similar to the unit test for the remote checkin https://github.com/foundriesio/aktualizr-lite/blob/e5106f240b4e2c27414e36c855d5d54ab33ab7e0/tests/apiclient_test.cc#L102.

include/aktualizr-lite/tuf/tuf.h
```diff
@@ -0,0 +1,90 @@
+#ifndef AKLITE_TUF_TUF_H_
```
mike-sul, line 1: @doanac I think the tuf api/interface is clean enough for the initial review.

include/aktualizr-lite/tuf/tuf.h
detsch: Let's keep the current approach, with multiple methods.

include/aktualizr-lite/api.h
detsch: Done. Strongly based on the `CheckIn` test.

detsch:
Regarding the second patch, that adds `CheckInLocal` and adjusts `CheckIn`, I've avoided changes in `LiteClient` for now. For example, I've replace the use of `client_->checkImageMetaOffline()` instead of adjusting `LiteClient::checkImageMetaOffline` code to use the new API. Focusing in the `api.cc` code is probably safer at the moment, but eventually, when we decide to update the main aklite executable, we will probably change  `LiteClient::updateImageMeta`, `LiteClient::allTargets`, `LiteClient::checkImageMetaOffline` and revert part of the changes made to `api.cc` now. It is just a couple of lines, though.

src/tuf/localreposource.h
```diff
@@ -0,0 +1,27 @@
...
+  // DISCUSS: limit to max version?
+  virtual std::string fetchRoot(int version);
+  virtual std::string fetchTimestamp();
+  virtual std::string fetchSnapshot();
+  virtual std::string fetchTargets();
+
+ private:
+  virtual std::string fetchFile(boost::filesystem::path meta_file_path);
+
+  std::string name;
```
mike-sul, line 23: What's the `name` for?

src/api.cc
```diff
@@ -119,57 +126,87 @@ AkliteClient::~AkliteClient() {
...
+  if (!configUploaded_) {
+    client_->reportAktualizrConfiguration();
+    configUploaded_ = true;
+  }
+  client_->reportNetworkInfo();
+  client_->reportHwInfo();
+  client_->reportAppsState();
+
+  boost::property_tree::ptree pt;
+  pt.put<std::string>("tag", client_->tags.empty() ? "" : client_->tags.at(0));
```
mike-sul, line 201: I think we need to set three more headers here: `x-ats-dockerapps`, `x-ats-target`, `x-ats-ostreehash` to make Device Gateway happy. The `checkin` function (https://github.com/foundriesio/device-gateway/blob/f1008143f27e6dc11dbf5bcf718d759a3df343d4/server/handlers_meta.go#L105) is called by the root meta handler, and these headers' values are used to set a device state at the DB.

Obviously, in the future when the tufctl becomes a separate/standalone daemon we will need to figure out how to change/adjust overall approach since the tufctl service won't know or shouldn't know about the currently installed target, ostree and list of configured apps.

src/tuf/akhttpsreposource.cc
```diff
@@ -20,13 +29,8 @@ AkHttpsRepoSource::~AkHttpsRepoSource() {
...
-void AkHttpsRepoSource::init(std::string name_in, boost::property_tree::ptree& pt) {
+void AkHttpsRepoSource::init(std::string name_in, boost::property_tree::ptree& pt, Config& config) {
   name_ = name_in;
-
-  boost::program_options::variables_map m;
-  Config config(m);
-  fillConfig(config, pt);
-
   // Libaktualizr places certificate files inside a sqlite database
   tmp_dir_path_ = boost::filesystem::temp_directory_path() /
```
mike-sul, line 35: Why do we need this tmp dir? IIRC, there is some redundancy in aklite with it.
If key/cert are stored in  HSM then there is no any tmp dir/files are required at all since libcurl/ssl loads them directly from HSM.
In the no HSM case, a key and certs are already on a file system, why do we need to copy them to some tmp storage?

detsch: I'll try to call `http_client->setCerts` directly, that might remove the need for the tmp dir. Working on it.

detsch: I've update the code to use `http_client->setCerts`, removing the use of a tmp dir.

src/tuf/localreposource.h
mike-sul: ok

src/api.cc
detsch: I've now added the fields to `CheckIn` logic, which was missing in the previous patches.

mike-sul: ok. I tested it by using the custom sota client/example, and it works fine now, a device status is not "erased" as it is used to be.

src/tuf/akhttpsreposource.cc
mike-sul: It works in my local setup.

src/api.cc
```diff
@@ -119,57 +127,91 @@ AkliteClient::~AkliteClient() {
...
+  auto current = client_->getCurrent();
+  pt.put<std::string>("dockerapps", Target::appsStr(current, ComposeAppManager::Config(client_->config.pacman).apps));
+  pt.put<std::string>("target", current.filename());
+  pt.put<std::string>("ostreehash", current.sha256Hash());
+
+  auto repo_src = std::make_shared<aklite::tuf::AkHttpsRepoSource>("temp-remote-repo-source", pt, client_->config);
```
mike-sul, line 208: We may wanna optimize it in the future so a new instance of `AkHttpsRepoSource` is not created on each checkin, instead just once when `AkliteClient` is initiated. Actually, this is rather about the http client not `AkHttpRepoSource`, so we may have another ctor for `AkHttpRepoSource` that accepts an http client as an input. Then just the http client is created once in the `AkliteClient` context. As result we'll the TLS/TCP connection can be reused.

It's not critical to do it now, can be done in the following PRs.

mike-sul:
@detsch It LGTM.

So, now, to perform final e2e testing, please, create a new PR for a branch within the `aktualizr-lite` github project so the CI will run all tests. Also we will be able to bitbake it in a scope of factory and test it e2e. In particular, we need to make sure that there is no regression in the custom sota client.

The next step is to add download from a local path/folder to the api (another PR), so we will have an offline update support in the api.

detsch:
Added fixes for clang-tidy errors.

detsch:
Closing this PR, #282 will be used from now on.

PR #282, opened 2023-11-28
Add a TUF API to aklite

detsch:
The API allows a better isolation between aktualizr-lite and libaktualizr, and also allows the use of multiple sources for a same TUF repository.

Changes were originally submitted in #279

### DIFF
--- a/apps/tufctl/main.cpp
+++ b/apps/tufctl/main.cpp
@@ -1,0 +1,106 @@
+#include <boost/property_tree/ini_parser.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/ptree_fwd.hpp>
+#include "aktualizr-lite/tuf/tuf.h"
+#include "tuf/akhttpsreposource.h"
+#include "tuf/akrepo.h"
+#include "tuf/localreposource.h"
+
+// Strip leading and trailing quotes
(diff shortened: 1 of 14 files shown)


mike-sul:
LGTM. Let's bitbake and test it e2e (focus on the custom client, make sure no regression there).

mike-sul (approved):
I think this is ready to go.

apps/tufctl/main.cpp
```diff
@@ -0,0 +1,106 @@
+#include <boost/property_tree/ini_parser.hpp>
```
doanac, line 1: This seems like its more for debug/development? I'm not sure what the long term point of this is - is more of an "examples"?

mike-sul: Yes, this is rather a test and an example utility. That's why it's not built by default.

doanac: y'all put a lot of thought into this so i won't bikeshed. But it seems like we might be developing an "examples" directory over time.

doanac (approved):

mike-sul:
@detsch Feel free to merge it.

detsch:
I've done some tests yesterday using imx8mm with SE05X, focusing on custom-client. All seems OK.

mike-sul:
> I've done some tests yesterday using imx8mm with SE05X, focusing on custom-client. All seems OK.

Let's merge it then.

PR #283, opened 2023-11-30
Composeapp integration

mike-sul:
Switch to the new golang based utility for compose app management.
The new utility will replace:
1. `skopeo` (container image download&install, `skopeo` can be removed after the switch);
2. the C++ code in aklite: 
   a) to pull and install app (first phase, current);
   b) all other app related functionality, such as pruning, checking if fetched, installed, currently running, etc;
4. `aklite-apps` - the utility that is currently aimed for the app early startup (and as it is turned out customers started using it just for general app management).


mike-sul:
@doanac 
This WIP branch works well with https://github.com/foundriesio/composeapp, however there are a few cleanings and adjustment works are required in the `aklite-apps` and `aklite-offline` utilities.

mike-sul:
@doanac @detsch I might add some more changes or improvement to this PR, but on the whole, it works fine in the scope of the LmP that includes this change https://github.com/foundriesio/meta-lmp/pull/1369.

mike-sul:
@doanac @detsch I think this PR is good to go. Once it's merged I'll add a commit with an aklite version to the corresponding meta-lmp PR https://github.com/foundriesio/meta-lmp/pull/1369.

src/ctr/appengine.cc
```diff
@@ -0,0 +1,10 @@
+#include "appengine.h"
+
+
+namespace ctr {
+  AppEngine::Result AppEngine::fetch(const App& app) {
+    // TODO
+    return true;
+  }
+
+} // namespace ctr
```
doanac: Is the "ctr" here implying the future code with be using the containerd "ctr" stuff? You tend to err on the side of overly verbose with names - this seems a bit terse to me

CMakeLists.txt
```diff
@@ -20,7 +20,6 @@ option(BUILD_OSTREE "Set to ON to compile with ostree support" ON)
 option(ALLOW_MANUAL_ROLLBACK "Set to ON to build with support of manual rollbacks" OFF)
 option(BUILD_AKLITE_OFFLINE "Set to ON to build cli command for an offline update" OFF)
 option(BUILD_AKLITE_WITH_NERDCTL "Set to ON to build with support of nerdctl/containerd" OFF)
-option(BUILD_AKLITE_APPS "Set to ON to build cli command for Apps management" ON)
 option(BUILD_TUFCTL "Set to ON to build sample tuf control application" OFF)
```
doanac, line 23: The commit message is inaccurate. The next commit turns on composectl

CMakeLists.txt
```diff
@@ -20,7 +20,6 @@ option(BUILD_OSTREE "Set to ON to compile with ostree support" ON)
 option(ALLOW_MANUAL_ROLLBACK "Set to ON to build with support of manual rollbacks" OFF)
 option(BUILD_AKLITE_OFFLINE "Set to ON to build cli command for an offline update" OFF)
 option(BUILD_AKLITE_WITH_NERDCTL "Set to ON to build with support of nerdctl/containerd" OFF)
-option(BUILD_AKLITE_APPS "Set to ON to build cli command for Apps management" ON)
```
doanac, line 23: Seems like this is going to cause changes to https://docs.foundries.io/latest/user-guide/container-preloading/container-preloading.html ? what's the plan for that? ie can this break someone using container-preloading?

src/ctr/appengine.cc
mike-sul: >Is the "ctr" here implying the future code with be using the containerd "ctr" stuff?

No, I just meant a short name for `containers` and I've not planned to use any `ctr` stuff. Currently just some pieces of containerd's functionality is used by `composectl`.

Yes, I tend to be verbose when it comes to variables/classes/methods, not for namespaces though :).

I just couldn't come up with any reasonable and short name for this namespace, maybe just `compose`?

CMakeLists.txt
mike-sul: Why do you think it is inaccurate?

CMakeLists.txt
mike-sul: No, the given change doesn't cause changes to the preloading.

CMakeLists.txt
mike-sul: Yes, it is.

CMakeLists.txt
doanac: The commit 1e05cd3d4e5da5c9e98fe554fb48b02fc204090d only removes aklite apps. It does not turn on composectl usage.

CMakeLists.txt
doanac: The commit 1e05cd3 only removes aklite apps. It does not turn on composectl usage.

CMakeLists.txt
doanac: But that document refers to the `aklite-apps` tool which is going away. Is it simply a matter of removing the text or do we have people that use this tool? Or is this a "release notes" thing where people upgrading will have to change something?

mike-sul: The corresponding meta-lmp change creates a symbolic link `aklite-apps` that refers to `composectl` https://github.com/foundriesio/meta-lmp/pull/1369/commits/4243e880cda2c03c5b71a373527fcf450793b1c3#diff-5d7444c561f7fd4d3c4b9d87080fecb23114011e6530898be51380df7fd2b2c4R24.

CMakeLists.txt
mike-sul: I agreed on it.

src/ctr/appengine.cc
mike-sul: It's preferable to name a sub-directory with the same name as a namespace. Maybe, calling it `composeapp` is a good name for both?

doanac: lgtm

CMakeLists.txt
doanac: ah. i assumed the first release of this would have it optional. We are going to require the new utlitlity (if you use aklite-apps). I guess the aklite-apps part of this PR could really be its own unique PR and is almost orthogonal to the introduction of composeapp as an app engine.

Minor thing but maybe something to consider in future release notes. I'm guessing the size of `composectl` is a bit bigger than the old `aklite-apps`. We might want to note that or at least know it ourselves.

mike-sul: Removing `skopeo` compensate diff between `composectl` and `aklite-apps` sizes.
Sure, we can add info about approximate size of `composectl`.

doanac: it helps the *total* size. I was thinking about the cost of the OTA

mike-sul: ah, ok.

FYI (arm64, rpi4)
```
du -sh /bin/composectl
9.0M	/bin/composectl
```

src/ctr/appengine.cc
mike-sul: 86a0de0

CMakeLists.txt
mike-sul: Updated the commit message.

doanac (approved):
looks good.

PR #284, opened 2023-12-05
offline: Don't require docker restart if just app update

mike-sul:
The recent [patch](https://github.com/foundriesio/meta-lmp/blob/main/meta-lmp-base/recipes-containers/docker/files/0001-tarexport-optimize-image-loading-on-local-host.patch) to the docker engine makes its restart redundant since app images are correctly registered at the end of the image loading process.


detsch (approved):
LGTM

PR #285, opened 2023-12-06
aklite-apps: Remove the register command

mike-sul:
This change removes `register` command which becomes unneeded since the docker engine has been patched with functionality that registers images at the end of image loading process.


detsch (approved):
LGTM

PR #286, opened 2023-12-07
offline: Require all enabled apps for offline update

mike-sul:
It is unsafe to employ "content-based" shortlisting due to the potential risk of installing unwanted targets or unintentionally removing necessary apps.

The proposed alternative mandates that all enabled/configured apps listed in the target must be present in the update source bundle, specifically the intersection of pacman.compose_app and target.apps. This ensures that a device is updated with the anticipated ostree and apps as defined by the target JSON and device configuration.

It's worth noting that there might be scenarios where multiple targets share the same ostree and the app intersection. This is not problematic as the update content remains consistent across these targets. The only drawback is that the offline updater might display an unexpected target name in the logs, potentially deviating from the user's expectations. However, it's crucial to emphasize that the update content, including ostree version and app list with their respective versions, precisely aligns with the user's intended installation.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -655,21 +655,20 @@ bool ComposeAppManager::compareAppsStates(const Json::Value& left, const Json::V
   return true;
 }
 
-ComposeAppManager::AppsContainer ComposeAppManager::getAppsToFetch(const Uptane::Target& target,
-                                                                   bool check_store) const {
+ComposeAppManager::AppsContainer ComposeAppManager::getRequiredApps(const Config& cfg, const Uptane::Target& target) {
   AppsContainer apps;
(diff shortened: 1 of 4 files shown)


detsch (approved):
Code looks good. The `enforce_pacman_type` addition seems to be orthogonal to the main change, perhaps it could be a separate commit it that is not too much work. I'll be able to give the code some testing in a few hours.

mike-sul:
> The `enforce_pacman_type` addition seems to be orthogonal to the main change, perhaps it could be a separate commit it that is not too much work. I'll be able to give the code some testing in a few hours.

Your are right. The `enforce_pacman_type` is really needed for the tests in this context. I just hope that this code will be removed soon anyway once you/we implement the offline update in the api.

PR #289, opened 2023-12-14
api: Add offline target installation

detsch:
Add a new LocalLiteInstall class, that inherits from the original
LiteInstall class, overriding the Download method, making it copy target
data locally instead of fetching it remotelly.

The implementation is strongly based on the standalone offline client
code.

---
Unit tests still need to be adjusted, specially `CheckInLocal` that  now requires apps and ostree data in order to accept a target as valid.

There is an ongoing discussion about `CheckAppsInSync`. Perhaps it does not need to receive a source path in the offline case, as reset-apps would be accessed instead.

### DIFF
--- a/examples/custom-client-cxx/main.cc
+++ b/examples/custom-client-cxx/main.cc
@@ -83,8 +83,8 @@ int main(int argc, char **argv) {
 
     try {
       CheckInResult res = local_update_path.empty()?
-        client->CheckIn() :
-        client->CheckInLocal(local_update_path);
+                           client->CheckIn():
+                           client->CheckInLocal(local_update_path + "/tuf", local_update_path + "/ostree_repo", local_update_path + "/apps");
       if (res.status != CheckInResult::Status::Ok && res.status != CheckInResult::Status::OkCached) {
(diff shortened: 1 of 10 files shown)


examples/custom-client-cxx/main.cc
```diff
@@ -50,6 +51,13 @@ int main(int argc, char **argv) {
     cfg_dirs = AkliteClient::CONFIG_DIRS;
   }
 
+  if (env.end() != env.find("AKLITE_LOCAL_UPDATE_PATH")) {
```
mike-sul, line 54: I would rather add new sub-command(s) or even a new executable for offline update. The online daemon mode and the offline non-daemon mode/update differ in subtle ways, hence it's better not to mix their implementation within a single context.

mike-sul: IMO, it's worth adding a new executable that would provide customers with an example of a cli update client (non-daemon) that can perform both online and offline update.

mike-sul: There are diffs (no need in the env. var., no reboot, etc). However there is no need in a new executable indeed. Please, see my comment in slack (as well as corresponding jira tickets) and sorry for misunderstanding :).

PR #290, opened 2023-12-15
Remove nerdctl

mike-sul:
Remove nerdctl support.


doanac (approved):
deleted code is always good but it's a little sad this never took off.

mike-sul:
> deleted code is always good but it's a little sad this never took off.

Unfortunately.
We've gone too far with relying on dockerd. Also, `nerdctl` is tuned/tailored for the NTT/cloud use case too much. These two facts are key reasons for dropping this direction.

PR #291, opened 2024-01-05
tests: Support chunked streaming of image load manifest

mike-sul:
The offline test suite uses a real `composectl` utility which streams the image load manifest to a docker daemon. Therefore, the docker daemon mock should support receiving the manifest transferred via a chunked HTTP stream, otherwise the offline unit tests fail.


mike-sul:
This is just a fix of the offline unit test, i.e. adjustment to the latest version of `composectl`.

doanac (approved):

PR #292, opened 2024-01-08
Fix image load manifest tarring

mike-sul:
It turned out that the previous logic of the image load manifest tarring didn't work on all use-cases. This changes fixes and improves the tarring, specifically calculation of the buffer size required for resultant tar data.


src/docker/dockerclient.cc
```diff
@@ -216,10 +216,25 @@ Json::Value DockerClient::getEngineInfo() {
...
+  }
   struct archive* a;
   struct archive_entry* entry;
   size_t archive_size;
-  std::vector<uint8_t> tar_data(2 * data.size() + 500 + 512 + 512);
+  // align size to 512 block size, the size must be a multiple of 512, assumes `dataSize` is not zero
+  const size_t aligned_data_size =
+      (data.size() + tar_block_size - 1) & ~(tar_block_size - 1) /* data_size + 511 (0x1FF) & ~0x1FF */;
+  const size_t estimated_archive_size =
+      tar_block_size /* header */ + aligned_data_size + 512 * 2 /* end-of-file marker, two block filled wth zeros */;
```
detsch, line 236:
```suggestion
      tar_block_size /* header */ + aligned_data_size + 512 * 2 /* end-of-file marker, two blocks filled with zeros */;
```

detsch (approved):
LGTM

src/docker/dockerclient.cc
mike-sul: updated

PR #295, opened 2024-01-16
Strengthen target check before starting installation

mike-sul:
Strengthen target check before starting installation:
 - make sure the target specified to the installer matches the one that is found in the local TUF repo;
 - make sure the target specified to the local installer is available locally - its content is found in the specified local folder during the check-in phase.


detsch (approved):
LGTM

I did not have a chance to test it yet, but will give it a try next week.

PR #296, opened 2024-01-17
cli: offline: Let client to decide on install mode

mike-sul:
Don't enforce the "ostree only" installation mode for an offline update. Instead, let a user/client decide which of the installation modes to apply.


detsch (approved):
LGTM

PR #297, opened 2024-01-17
liteclient: Import the first common root meta versions

mike-sul:
If a device is not registered and/or there is no a device certificate then a liteclient does not know whether to preload production or CI root metadata preloaded on a device. Instead of failing, let's import the first two versions since they are same across prod and CI. It's even more secure compared to failing and relying on TOFU (pulling all root meta versions starting from 1.root.json).


mike-sul:
@detsch This is another fix to run `aklite-offline` on non-registered devices.

PR #298, opened 2024-01-22
Offline update corner case fixes

mike-sul:
- Fix offline update for non-registered devices
- Don't allow offline update if the provided TUF meta are invalid even if a local/cached one are valid.
- Fix error printing if the specified source directory doesn't exist.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -340,7 +340,12 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
       .AppsDir = local_update_source->app_store,
   };
 
-  if (result.status == CheckInResult::Status::Failed) {
+  if (result.status == CheckInResult::Status::OkCached) {
+    // There is no reason to allow offline update if the specified TUF metadata are invalid,
+    // even if a local copy is valid.
+    result.status = CheckInResult::Status::Failed;
+  }
+  if (!result) {
     return result;
   }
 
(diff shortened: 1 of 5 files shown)


src/api.cc
```diff
@@ -631,6 +636,12 @@ class LocalLiteInstall : public LiteInstall {
 
     ostree_sysroot_ = std::make_shared<OSTree::Sysroot>(offline_update_config_.pacman);
     storage_ = INvStorage::newStorage(offline_update_config_.storage, false, StorageClient::kTUF);
+    if (offline_update_config_.pacman.type == ComposeAppManager::Name &&
+        offline_update_config_.pacman.extra.count("reset_apps") == 0) {
+      LOG_ERROR << "Cannot perform offline update if non-restorable app engine is set; set `[pacman].reset_apps = "
+                   "\"\"` in the device config.";
+      throw std::invalid_argument("Invalid an app engine type");
```
detsch, line 643:
```suggestion
      throw std::invalid_argument("Invalid app engine type");
```

src/liteclient.cc
```diff
@@ -143,6 +143,13 @@ LiteClient::LiteClient(Config config_in, const AppEngine::Ptr& app_engine, const
     }
   }
   if (config.pacman.type == ComposeAppManager::Name) {
+    if (app_engine == nullptr && config.pacman.extra.count("reset_apps") == 0 && !isRegistered(*key_manager_) &&
+        config.pacman.extra.count("enforce_composeapp_engine") == 0) {
+      LOG_DEBUG << "Device is not registered and a type of an app engine is not configured,"
+                   " enforcing the restorable package manager";
```
detsch, line 149:
```suggestion
      LOG_DEBUG << "Device is not registered and the app engine type is not configured,"
                   " enforcing the use of restorable package manager";
```

detsch (approved):
LGTM

src/api.cc
mike-sul: fixed via force push

src/liteclient.cc
mike-sul: fixed via force push

PR #299, opened 2024-01-23
Unify CLI exit codes and improve err msgs

mike-sul:
- Improve log messages related to updating of the device/local TUF repo.
- Unify exit codes across both CLI utilities, `aklite-offline` and `aktualizr-lite`.
- Add tests to cover changes related to the exit code unification.


src/api.cc
```diff
@@ -332,33 +328,38 @@ std::vector<TufTarget> toTufTargets(const std::vector<Uptane::Target>& targets)
...
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "Failed to update the local TUF repo: " << exc.what();
+    return CheckInResult{CheckInResult::Status::Failed, "", {}};
+  }
+
+  LOG_INFO << "The local TUF repo has been successfully updated";
+  LOG_INFO << "Searching for matching TUF Targets...";
+  auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, client_->tags, secondary_hwids_);
+  if (matchingTargets.empty()) {
+    LOG_ERROR << "No matching Targets are found for a given device; check a device tag or a hardware ID";
```
detsch, line 344:
```suggestion
    LOG_ERROR << "There are no Targets matching the current device tag and hardware ID";
```

src/api.cc
```diff
@@ -214,8 +175,43 @@ CheckInResult AkliteClient::CheckIn() const {
...
+  }
+
+  if (check_status == CheckInResult::Status::Failed) {
+    return CheckInResult{check_status, "", {}};
+  }
+
+  LOG_INFO << "Searching for matching TUF Targets...";
+  auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, client_->tags, secondary_hwids_);
+  if (matchingTargets.empty()) {
+    LOG_ERROR << "No matching Targets are found for a given device; check a device tag or a hardware ID";
```
detsch, line 210:
```suggestion
    LOG_ERROR << "There are no Targets matching the current device tag and hardware ID";
```

detsch (approved):
LGTM

src/api.cc
mike-sul: Can we add the verb `found` or synonym to this error message so it goes inline with the `"Searching for matching TUF Targets..." message. 
Maybe something like "Couldn't find..." or "Failed to find"?

detsch: Sure, makes sense. Adjusted the suggestions accordingly.

mike-sul: updated via force push

PR #300, opened 2024-01-26
[WIP] Allow install to be called independently of pull

detsch:
These are the changes required to support a standalone install operation. They allow apps to be installed properly, and the installation to fail at the right time when the target was not pulled.

At the current stage, there is some code duplication that could be avoided (see `TDB` comments).

A subsequent PR will include the custom client commands allowing a standalone install operation.


detsch:
Closing in favor of PR #301

PR #301, opened 2024-01-26
Allow install to be called independently of pull - V2

detsch:
[Creating a new PR since the current solution is very different (code wise) to the original PR #300.]

These are the changes required to support a standalone install operation. They allow apps to be installed properly, and the installation to fail at the right time when the target was not pulled.

A subsequent PR will include the custom client commands allowing a standalone install operation.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -404,6 +404,14 @@ class LiteInstall : public InstallContext {
   InstallResult Install() override {
     client_->logTarget("Installing: ", *target_);
 
+    // Call appsInSync to update applications list inside the package manager
+    client_->appsInSync(*target_);
+    // setAppsNotChecked is required to force a re-load of apps list in case of a new Download operation
+    client_->setAppsNotChecked();
+    if (client_->VerifyTarget(*target_) != TargetStatus::kGood) {
+      return InstallResult{InstallResult::Status::DownloadFailed, ""};
+    }
+
     auto rc = client_->install(*target_, mode_);
     auto status = InstallResult::Status::Failed;
     if (rc == data::ResultCode::Numeric::kNeedCompletion) {
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -326,6 +326,10 @@ TargetStatus ComposeAppManager::verifyTarget(const Uptane::Target& target) const
   all_apps_to_fetch.insert(cur_apps_to_fetch_.begin(), cur_apps_to_fetch_.end());
 
   for (const auto& pair : all_apps_to_fetch) {
+    if (!app_engine_->isFetched({pair.first, pair.second})) {
+      return TargetStatus::kNotFound;
+    }
+
     if (!app_engine_->verify({pair.first, pair.second})) {
       return TargetStatus::kInvalid;
     }
--- a/tests/apiclient_test.cc
+++ b/tests/apiclient_test.cc
@@ -342,6 +342,34 @@ TEST_F(ApiClientTest, InstallModeOstreeOnlyIfJustApps) {
   }
 }
 
+TEST_F(ApiClientTest, InstallWithoutDownload) {
+  auto liteclient = createLiteClient();
+  ASSERT_TRUE(targetsMatch(liteclient->getCurrent(), getInitialTarget()));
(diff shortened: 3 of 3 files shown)


tests/apiclient_test.cc
```diff
@@ -342,6 +342,34 @@ TEST_F(ApiClientTest, InstallModeOstreeOnlyIfJustApps) {
   }
 }
 
+TEST_F(ApiClientTest, InstallWithoutDownload) {
```
mike-sul, line 345: I would add a positive test that downloads and installs a target by using another instance of liteclient and correspondingly another instance of the installer for this two actions.

src/api.cc
```diff
@@ -404,6 +404,13 @@ class LiteInstall : public InstallContext {
   InstallResult Install() override {
     client_->logTarget("Installing: ", *target_);
 
+    // Call appsInSync to update applications list inside the package manager
+    client_->appsInSync(*target_);
+    client_->setAppsNotChecked();
```
mike-sul, line 409: What is the use-case `setAppsNotChecked()` is needed for?

detsch: We need to reset `are_apps_checked_ = false` in case the same instance is used for Downloading a different target. If `are_apps_checked_ == true`, the apps list will not be updated. I'll add a comment in this location, since it is not obvious.

mike-sul: ok, maybe it makes sense to add a unit test for this subtle use-case then.

detsch: There is already a unit test that fails without it inside `test_cli`, at this point:
```
  ASSERT_EQ(cli::Install(*akclient, target01.Version()), cli::StatusCode::Ok);
  ASSERT_EQ(akclient->GetCurrent(), target01);

  auto app01 = registry.addApp(
      fixtures::ComposeApp::create("app-01", "service-01", "image-01", fixtures::ComposeApp::ServiceTemplate,
                                   Docker::ComposeAppEngine::ComposeFile, "compose-start-failure"));
  const auto target02 = createTufTarget(&app01);

  ASSERT_EQ(cli::Install(*akclient, target02.Version()), cli::StatusCode::InstallNeedsReboot);
```

tests/apiclient_test.cc
detsch: I believe a test like this would be much more useful if the actual code for the `ComposeAppEngine` is used, instead of `MockAppEngine`. Using `MockAppEngine`, the previous version would still make the test pass.
Where could I place it?

src/api.cc
mike-sul: ok

tests/apiclient_test.cc
mike-sul: The `cli_test` uses the real package manager and app engine.
However, it will require the CLI API extending - adding functions to pull and install separately.

detsch: I had a hard time trying to make a proper test inside `cli_test.c`. Not only `AkliteClient` has to be changed, but the mocks and actual objects organization inside that file require changes. For example, multiple objects created with `createAkClient` share the same `app_engine`. Just creating separate `app_engine` instances is not that hard, but they would have to share the same and path, which would require more changes. It looks this is too much work for at the moment, for little benefit, so I simply added one test to the api tests, even if it does not test the `ComposeAppEngine` behavior.

src/composeappmanager.cc
```diff
@@ -326,6 +326,10 @@ TargetStatus ComposeAppManager::verifyTarget(const Uptane::Target& target) const
   all_apps_to_fetch.insert(cur_apps_to_fetch_.begin(), cur_apps_to_fetch_.end());
 
   for (const auto& pair : all_apps_to_fetch) {
+    if (!app_engine_->isFetched({pair.first, pair.second})) {
+      return TargetStatus::kInvalid;
```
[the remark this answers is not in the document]
detsch: Fixed.

tests/apiclient_test.cc
mike-sul: I understand and that's fine.

mike-sul (approved):

PR #302, opened 2024-01-30
custom-client: Add additional running modes, with separate commands

detsch:
Original running mode now invoked with `custom-sota-daemon`.

Added commands: `check`, `pull`, `install`, `run`.

---
Items for discussion:

- I did not use AkLiteClient (cli.cc) methods, but copied over code from there. We could use `cli.cc`, but from what I understand, methods provided by `api.cc` are the interface for a customized client. If we start to officially provide `cli.cc` methods to external users, that must be a well though decision, as the behavior and signature of `cli.cc` methods should not be changed anymore.

- I avoided code reuse between different commands code. That is to simplify custom solution by users, simply taking one command code and changing it.

- I did not add `log-level` cmd option, but is not hard to add.

- See a few TBD comments bellow, if it is OK to return SUCCESS in those cases.

---
I've been using the new commands for a while, but if the overall structure is OK, I'll do a few more tests before merge.

### DIFF
--- a/examples/custom-client-cxx/cli.cpp
+++ b/examples/custom-client-cxx/cli.cpp
@@ -1,0 +1,205 @@
+#include <iostream>
+#include <thread>
+
+#include <boost/filesystem.hpp>
+#include <boost/log/core.hpp>
+#include <boost/log/expressions.hpp>
+#include <boost/log/trivial.hpp>
+#include <boost/process.hpp>
+
+#include "aktualizr-lite/api.h"
+#include "aktualizr-lite/cli/cli.h"
+#include "cli.h"
+
+#define LOG_INFO BOOST_LOG_TRIVIAL(info)
+#define LOG_WARNING BOOST_LOG_TRIVIAL(warning)
+#define LOG_ERROR BOOST_LOG_TRIVIAL(error)
+
+static void print_status(aklite::cli::StatusCode ret) {
+  switch (ret) {
+    case aklite::cli::StatusCode::Ok:
+      std::cout << "SUCCESS";
+      break;
+
+    // Possible return codes for check, pull and install commands
+    case aklite::cli::StatusCode::CheckinOkCached:
+      std::cout << "SUCCESS: Unable to fetch updated TUF metadata, but stored metadata is valid";
+      break;
+    case aklite::cli::StatusCode::CheckinFailure:
+      std::cout << "FAILURE: Unable to get TUF metadata";
+      break;
+    case aklite::cli::StatusCode::CheckinNoMatchingTargets:
+      std::cout << "FAILURE: There is no matching target for the device";
+      break;
+    case aklite::cli::StatusCode::CheckinNoTargetContent:
+      std::cout << "FAILURE: There is no target metadata in the local path";
+      break;
+    case aklite::cli::StatusCode::TufTargetNotFound:
+      std::cout << "FAILURE: Selected target not found";
+      break;
+
+    // Possible return codes for pull and install commands
+    case aklite::cli::StatusCode::InstallationInProgress:
+      std::cout << "FAILURE: Unable to pull/install: there is an installation that needs completion";
+      break;
+    case aklite::cli::StatusCode::DownloadFailure:
+      std::cout << "FAILURE: Unable to download target";
+      break;
+    case aklite::cli::StatusCode::DownloadFailureVerificationFailed:
+      std::cout << "FAILURE: Target downloaded but verification has failed";
+      break;
+    case aklite::cli::StatusCode::DownloadFailureNoSpace:
+      std::cout << "FAILURE: There is no enough free space to download the target";
+      break;
+    case aklite::cli::StatusCode::InstallAlreadyInstalled:
+      std::cout << "FAILURE: Selected target is already installed";
+      break;
+    case aklite::cli::StatusCode::InstallDowngradeAttempt:
+      // Should not hit this error, since force_downgrade is set to true
+      std::cout << "FAILURE: Attempted to install a previous version";
+      break;
+
+    // Possible return codes for install command
+    case aklite::cli::StatusCode::InstallAppsNeedFinalization:
+      std::cout << "SUCCESS: Execute `custom-sota-client run` command to finalize installation";
+      break;
+    case aklite::cli::StatusCode::InstallNeedsRebootForBootFw:
+      std::cout << "FAILURE: Reboot is required before installing the target";
+      break;
+    case aklite::cli::StatusCode::InstallAppPullFailure:
+      std::cout << "FAILURE: Unable read target data, make sure it was pulled";
+      break;
+
+    // Possible return codes for install and run command
+    case aklite::cli::StatusCode::InstallNeedsReboot:
+      std::cout << "SUCCESS: Reboot to finalize installation";
+      break;
+    case aklite::cli::StatusCode::OkNeedsRebootForBootFw:
+      std::cout << "SUCCESS: Reboot to finalize bootloader installation";
+      break;
+    case aklite::cli::StatusCode::InstallRollbackNeedsReboot:
+      std::cout << "FAILURE: Installation failed, rollback initiated but requires reboot to finalize";
+      break;
+
+    // Possible return codes for run command
+    case aklite::cli::StatusCode::NoPendingInstallation:
+      std::cout << "FAILURE: No pending installation to run";
+      break;
+    case aklite::cli::StatusCode::InstallRollbackOk:
+    case aklite::cli::StatusCode::InstallOfflineRollbackOk:
+      std::cout << "FAILURE: Installation failed, rollback performed";
+      break;
+    case aklite::cli::StatusCode::InstallRollbackFailed:
+      std::cout << "FAILURE: Installation failed and rollback operation was not successful";
+      break;
+    case aklite::cli::StatusCode::UnknownError:
+      std::cout << "FAILURE: Unknown error";
+      break;
+
+    default:
+      std::cout << "FAILURE: Unexpected return code " << static_cast<int>(ret);
+      break;
+  }
+  std::cout << std::endl;
+}
+
+static std::unique_ptr<AkliteClient> init_client(bool online_mode) {
+  boost::log::core::get()->set_filter(boost::log::trivial::severity >= boost::log::trivial::info);
+
+  std::vector<boost::filesystem::path> cfg_dirs;
+
+  auto env{boost::this_process::environment()};
+  if (env.end() != env.find("AKLITE_CONFIG_DIR")) {
+    cfg_dirs.emplace_back(env.get("AKLITE_CONFIG_DIR"));
+  } else if (online_mode) {
+    cfg_dirs = AkliteClient::CONFIG_DIRS;
+  } else {
+    // sota.toml is optional in offline mode
+    for (const auto& cfg : AkliteClient::CONFIG_DIRS) {
+      if (boost::filesystem::exists(cfg)) {
+        cfg_dirs.emplace_back(cfg);
+      }
+    }
+  }
+
+  try {
+    return std::make_unique<AkliteClient>(cfg_dirs, false, false);
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "Failed to initialize the client: " << exc.what();
+    return nullptr;
+  }
+}
+
+int cmd_check(std::string local_repo_path) {
+  auto client = init_client(local_repo_path.empty());
+  if (client == nullptr) {
+    return EXIT_FAILURE;
+  }
+
+  LocalUpdateSource local_update_source;
+  if (local_repo_path.empty()) {
+    LOG_INFO << "Online mode";
+  } else {
+    auto abs_repo_path = boost::filesystem::canonical(local_repo_path).string();
+    LOG_INFO << "Offline mode. Updates path=" << abs_repo_path;
+    local_update_source = {abs_repo_path + "/tuf", abs_repo_path + "/ostree_repo", abs_repo_path + "/apps"};
+  }
+
+  auto status = aklite::cli::CheckIn(*client, local_repo_path.empty() ? nullptr : &local_update_source);
+  print_status(status);
+  return aklite::cli::IsSuccessCode(status) ? EXIT_SUCCESS : EXIT_FAILURE;
+}
+
+int cmd_pull(std::string target_name, std::string local_repo_path) {
+  auto client = init_client(local_repo_path.empty());
+  if (client == nullptr) {
+    return EXIT_FAILURE;
+  }
+
+  LocalUpdateSource local_update_source;
+  if (local_repo_path.empty()) {
+    LOG_INFO << "Online mode";
+  } else {
+    auto abs_repo_path = boost::filesystem::canonical(local_repo_path).string();
+    LOG_INFO << "Offline mode. Updates path=" << abs_repo_path;
+    local_update_source = {abs_repo_path + "/tuf", abs_repo_path + "/ostree_repo", abs_repo_path + "/apps"};
+  }
+
+  auto status =
+      aklite::cli::Pull(*client, -1, target_name, true, local_repo_path.empty() ? nullptr : &local_update_source);
+  print_status(status);
+  return aklite::cli::IsSuccessCode(status) ? EXIT_SUCCESS : EXIT_FAILURE;
+}
+
+int cmd_install(std::string target_name, std::string local_repo_path) {
+  auto client = init_client(local_repo_path.empty());
+  if (client == nullptr) {
+    return EXIT_FAILURE;
+  }
+
+  LocalUpdateSource local_update_source;
+  if (local_repo_path.empty()) {
+    LOG_INFO << "Online mode";
+  } else {
+    auto abs_repo_path = boost::filesystem::canonical(local_repo_path).string();
+    LOG_INFO << "Offline mode. Updates path=" << abs_repo_path;
+    local_update_source = {abs_repo_path + "/tuf", abs_repo_path + "/ostree_repo", abs_repo_path + "/apps"};
+  }
+
+  auto status =
+      aklite::cli::Install(*client, -1, target_name, InstallMode::All, true,
+                           local_repo_path.empty() ? nullptr : &local_update_source, aklite::cli::PullMode::None);
+  print_status(status);
+  return aklite::cli::IsSuccessCode(status) ? EXIT_SUCCESS : EXIT_FAILURE;
+}
+
+int cmd_run() {
+  auto client = init_client(false);
+  if (client == nullptr) {
+    return EXIT_FAILURE;
+  }
+
+  auto status = aklite::cli::CompleteInstall(*client);
+  print_status(status);
+  return aklite::cli::IsSuccessCode(status) ? EXIT_SUCCESS : EXIT_FAILURE;
+}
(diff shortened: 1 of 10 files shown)


mike-sul:
Running `custom-sota-client` without any parameters outputs

```
Missing command

Usage:
	 aklite-offline <cmd> [options]
Supported commands: check install run pull daemon 
```
I suppose it should be `custom-sota-client` instead of `aklite-offline`.

mike-sul:
When I am running `custom-sota-client daemon` I am getting the following error:

```
[2024-01-31 11:03:21.768061] [0x00007f0aa8290340] [error]   Update content of the specified target is not available locally:
```
For some reason the installer treats this update as an offline one, however I see this log at the beginning:

```
[2024-01-31 11:03:09.470805] [0x00007f1edc20a340] [info]    Online mode
```

I think the issue is in this line https://github.com/foundriesio/aktualizr-lite/blob/42a18e5f148e55f43a71d7a573117bf45d9c92a6/examples/custom-client-cxx/daemon.cpp#L124

The `&local_update_source` is never `nullptr`, I think we should use `is_local_update?&local_update_source:nullptr` instead.

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,365 @@
...
+        LOG_INFO << "Found Latest Target: " << latest.Name();
+        selected_target = latest;
+      }
+    } else {
+      for (auto const &target: res.Targets()) {
+        if (target.Name() == target_name) {
+          selected_target = target;
+          break;
+        }
+        LOG_ERROR << "Target " << target_name << " not found or is incomplete";
```
mike-sul, line 127: I think it should be `continue` here so a user can install a specified target `target_name` if it is in the list of available targets (`res.Targets()`).

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,365 @@
...
+        LOG_ERROR << "Failed to find Target to rollback to after failure to start Apps at boot of a new sysroot";
+        return EXIT_FAILURE;
+      }
+      selected_target = rollback_target;
+      LOG_INFO << "Rollback Target is " << selected_target.Name();
+    }
+
+    if (selected_target.Name() != current.Name() && !client->IsRollback(selected_target)) {
+      std::string reason = "Updating from " + current.Name() + " to " + selected_target.Name();
+      auto installer = client->Installer(selected_target, reason, "", InstallMode::All, &local_update_source);
```
mike-sul, line 147: I think it should be something like `is_local_update?&local_update_source:nullptr` instead of `&local_update_source` because the code returns the "local" installer unconditionally.

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,365 @@
...
+    LOG_INFO << "Nothing to do";
+    return EXIT_SUCCESS;
+
+  } catch (const std::exception &exc) {
+    LOG_ERROR << "Failed to find or update Target: " << exc.what();
+    return EXIT_FAILURE;
+  }
+}
+
+int cmd_install(std::string target_name, std::string local_repo_path) {
```
mike-sul, line 167: The install command has the same two issues as the `pull` command.

detsch:
Working on using cli.cc functions, but, meanwhile, pushed the fixes for the identified issues.

detsch:
> Running `custom-sota-client` without any parameters outputs
> 
> ```
> Missing command
> 
> Usage:
> 	 aklite-offline <cmd> [options]
> Supported commands: check install run pull daemon 
> ```
> 
> I suppose it should be `custom-sota-client` instead of `aklite-offline`.

Fixed

examples/custom-client-cxx/cli.cpp
detsch: Fixed

examples/custom-client-cxx/cli.cpp
detsch: Fixed

detsch:
Force pushed a version using `aklite::cli`. I did not test it yet will proceed working on it tomorrow.

include/aktualizr-lite/cli/cli.h
```diff
@@ -34,12 +34,30 @@ enum class StatusCode {
...
 };
 
+/**
+ * The installation mode to be applied. Specified during InstallContext context initialization.
+ */
+enum class PullMode {
+  /**
+   * Default mode, do pull target during install operation.
+   */
+  PerformPull = 0,
```
mike-sul, line 44: IMO, `PullMode::All` and `PullMode::None` looks a bit better, the `Pull` word in the enum value is redundant since the enum type denotes it.

include/aktualizr-lite/cli/cli.h
```diff
@@ -34,12 +34,30 @@ enum class StatusCode {
...
+};
+
 StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_source);
 
+StatusCode Pull(AkliteClient &client, int version = -1, const std::string &target_name = "",
+                InstallMode install_mode = InstallMode::All, bool force_downgrade = true,
+                const LocalUpdateSource *local_update_source = nullptr);
 
 StatusCode Install(AkliteClient &client, int version = -1, const std::string &target_name = "",
                    InstallMode install_mode = InstallMode::All, bool force_downgrade = true,
```
mike-sul, line 58: Not related to this PR. This `force_downgrade` looks a bit off now. I think it can be replaced with another install mode. Let's think about it and make corresponding changes in another PR.

include/aktualizr-lite/cli/cli.h
```diff
@@ -34,12 +34,30 @@ enum class StatusCode {
...
+  /**
+   * Do no pull target during install. Target is expected to be pulled before.
+   */
+  SkipPull,
+};
+
 StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_source);
 
+StatusCode Pull(AkliteClient &client, int version = -1, const std::string &target_name = "",
+                InstallMode install_mode = InstallMode::All, bool force_downgrade = true,
```
mike-sul, line 54: Do we really need `install_mode` for `Pull`?

src/cli/cli.cc
```diff
@@ -77,8 +77,9 @@ StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_s
   return res2StatusCode<CheckInResult::Status>(c2s, cr.status);
 }
 
-StatusCode Install(AkliteClient &client, int version, const std::string &target_name, InstallMode install_mode,
-                   bool force_downgrade, const LocalUpdateSource *local_update_source) {
+static StatusCode pullAndInstall(AkliteClient &client, int version, const std::string &target_name,
```
mike-sul, line 80: Maybe it makes sense to introduce separate functions `pull` and `install`?

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,150 @@
...
+  std::unique_ptr<AkliteClient> client;
+  try {
+    client = std::make_unique<AkliteClient>(cfg_dirs);
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "Failed to initialize the client: " << exc.what();
+    return EXIT_FAILURE;
+  }
+
+  auto ret = aklite::cli::Pull(*client, -1, "", InstallMode::All, true, is_local_update ? &local_update_source : nullptr);
+  return aklite::cli::IsSuccessCode(ret)? EXIT_SUCCESS : EXIT_FAILURE;
```
mike-sul, line 87: Why not return the exit code returned from `aklite::cli::Pull`?

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,150 @@
...
+  std::unique_ptr<AkliteClient> client;
+  try {
+    client = std::make_unique<AkliteClient>(cfg_dirs);
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "Failed to initialize the client: " << exc.what();
+    return EXIT_FAILURE;
+  }
+
+  auto ret = aklite::cli::CheckIn(*client, is_local_update ? &local_update_source : nullptr);
+  return aklite::cli::IsSuccessCode(ret)? EXIT_SUCCESS : EXIT_FAILURE;
```
mike-sul, line 50: Why not return the exit code returned from `aklite::cli::Pull`?

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  std::unique_ptr<AkliteClient> client;
+  try {
+    client = std::make_unique<AkliteClient>(cfg_dirs);
+  } catch (const std::exception& exc) {
+    LOG_ERROR << "Failed to initialize the client: " << exc.what();
+    return EXIT_FAILURE;
+  }
+
+  auto ret = aklite::cli::Install(*client, -1, "", InstallMode::All, true, is_local_update ? &local_update_source : nullptr, aklite::cli::PullMode::SkipPull);
```
mike-sul, line 123: Maybe it makes sense to add `switch (ret) case <>:` statement and print message instructing user what to do next depending on the `ret` value.

include/aktualizr-lite/cli/cli.h
detsch: The value is passed during the Installer instantiation, and the internal Download has access to it. But it is not used, and sounds like there would be no reason to use it there. I'll remove the parameter.

include/aktualizr-lite/cli/cli.h
detsch: I was not happy with the current names. `All` and `None` are certainly better.

examples/custom-client-cxx/cli.cpp
detsch: Returning the exit code from `aklite::cli` functions directly, while making it clear if it is a success or a failure, and what to do next, sounds a reasonable approach.

I'm a heavy user of the `&&` operator when running multiple commands in shell, so I tend to like having 0 for success whenever possible. But the consistency with the remaining tools, and the fact that the flow should in fact depend on the returned value, are more important.

Do you think it is worth to keep the `IsSuccessCode` function?

src/cli/cli.cc
detsch: IMO, only if e break the current function in some functions that can be reused. Otherwise there is too much duplication between standalone pull and install functions.  I'm taking a look.

examples/custom-client-cxx/cli.cpp
mike-sul: This is just an example, so I think it is fine to return 0 or 1.
What is an important is to add `switch case` (or if) statement that prints a message that corresponds to the `check/pull/install/run` execution result and in some cases instructs what to do the next.

Whether to return 0 or 1 is up to a customer to decide. I think it depends on a specific use-case.  I can imagine use-case when a custom client triggers another binary/script as a reaction on certain `ret` values (e.g. reboot), so the custom client doesn't need to return specific exit codes, just 0 and 1.

include/aktualizr-lite/cli/cli.h
detsch: Updated enum entries names.

include/aktualizr-lite/cli/cli.h
detsch: Removed `install_mode` from `Pull`

src/cli/cli.cc
detsch: I spent some time trying to get to a good solution having separate functions, without duplicating much logic. But the resulting code ended up being worse than it is now. So, from my feeling, it is better to keep is as is.

examples/custom-client-cxx/cli.cpp
detsch: I've added switch case logic for every possible return code. Since there is quite a lot of duplication between possible return codes for each command, I've concentrated all of them in a single `printStatus` function. Let me know what you think about it.

detsch:
I did some testing but will still do more tests on Monday.

src/cli/cli.cc
mike-sul: That's fine.

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+static void printStatus(aklite::cli::StatusCode ret) {
+  switch (ret) {
+    case aklite::cli::StatusCode::Ok:
+      std::cout << "SUCCESS";
+      break;
+
+    // Possible return codes for check, pull and install commands
+    case aklite::cli::StatusCode::CheckinOkCached:
+      std::cout << "SUCCESS: TUF data already up-to-date";
```
mike-sul, line 25: Not sure this is correct message since `CheckinOkCached` is usually returned when a client fails to fetch meta from remote tuf repo. This code just states that a local TUF repo is valid.

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    // Possible return codes for install and run command
+    case aklite::cli::StatusCode::InstallNeedsReboot:
+      std::cout << "SUCCESS: Reboot to finalize installation";
+      break;
+    case aklite::cli::StatusCode::OkNeedsRebootForBootFw:
+      std::cout << "SUCCESS: Reboot to finalize bootloader installation";
+      break;
+    case aklite::cli::StatusCode::InstallRollbackNeedsReboot:
+      std::cout << "SUCCESS: Installation failed, rollback was successful but requires reboot to finalize";
```
mike-sul, line 80: Not sure if it is "SUCCESS".

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,232 @@
...
+      break;
+
+    // Possible return codes for run command
+    case aklite::cli::StatusCode::NoPendingInstallation:
+      std::cout << "FAILURE: No pending installation to run";
+      break;
+    case aklite::cli::StatusCode::InstallRollbackOk:
+    case aklite::cli::StatusCode::InstallOfflineRollbackOk:
+      std::cout << "SUCCESS: Installation failed but rollback was successful";
```
mike-sul, line 89: Not sure if it is "SUCCESS". Maybe something like "FAILURE: failed to install; SUCCESS: rollback is successful".

examples/custom-client-cxx/cli.cpp
mike-sul: Looks fine to me.

mike-sul:
LGTM, will do some testing a bit later.

detsch:
> Please, make a PR to meta-lmp in order to add `daemon` to the `ExecStart` value https://github.com/foundriesio/meta-lmp/blob/b696ba3b883eace52e422e9a1c806aad6d5fd367/meta-lmp-base/recipes-sota/custom-sota-client/files/systemd.service#L13, unless we make the `custom-sota-client` command run in the daemon mode by default if none of the commands is specified as a CLI parameter. I don't have a strong opinion on it.

No reason to break compatibility, IMO. I've made `daemon` command the default if  no parameters are passed.

mike-sul:
If I run `pull` and specify non-existing target name, then the error says `version: latest`.

```
$CUSTOM_EXE  pull -t non-existing-target

[2024-02-06 05:43:55.162127] [0x00007f195710e340] [error]   No Target found; version: latest, hardware ID: "intel-corei7-64", tag: "main"
FAILURE: Selected target not found
```

The issue is not in the PR's code, but it would be great to fix it in this PR.
https://github.com/foundriesio/aktualizr-lite/blob/be03cd3ba276a1dd1567f1c3f73a5126b7205711/src/cli/cli.cc#L99
It doesn't take into account the `target_name` value if `version==-1`.

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,232 @@
...
+  auto env{boost::this_process::environment()};
+  if (env.end() != env.find("AKLITE_CONFIG_DIR")) {
+    cfg_dirs.emplace_back(env.get("AKLITE_CONFIG_DIR"));
+  } else {
+    cfg_dirs = AkliteClient::CONFIG_DIRS;
+  }
+
+  std::unique_ptr<AkliteClient> client;
+  try {
+    client = std::make_unique<AkliteClient>(cfg_dirs);
```
mike-sul, line 223: The `run` command reports failure while it is actually succeeds in the "happy path" flow.
```
[+] Building 0.0s (0/0)                                                                                                                                      docker:default
[2024-02-06 06:00:23.284611] [0x00007f507d90a340] [info]    Finalization has been completed successfully; Target: intel-corei7-64-lmp-30
[2024-02-06 06:00:23.292628] [0x00007f507d90a340] [info]    Use existing SQL storage: "/work/fio/prjs/lmp-device-register/fioed-device/sql.db"
[2024-02-06 06:00:23.296813] [0x00007f507d90a340] [error]   There is no pending installation to complete
FAILURE: No pending installation to run

```
The problem is that the `AkliteClient` object is instantiated with `finalize=true` while it should be `false` so the finalization/run can be executed and processed in `aklite::cli::CompleteInstall` command.

The same is for the `AkliteClient` instantiation for the rest commands.

mike-sul:
Does it require much effort to remove `-t` parameter and instead just allow target specifying without `-t`? E.g:

```
custom-sota-client install intel-corei7-64-lmp-30
```

examples/custom-client-cxx/cli.cpp
detsch: Updated all `AkliteClient` instantiations, setting `finalize=false`.

detsch:
> If I run `pull` and specify non-existing target name, then the error says `version: latest`.

Created a new patch fixing that issue. I've structured the code so that both `version` and `name` are printed if the caller sets them both, for whatever reason. Otherwise, it will print only the parameter than was set. If neither is set, `version: latest` will be shown.

examples/custom-client-cxx/cli.cpp
```diff
@@ -0,0 +1,232 @@
...
+  }
+  std::cout << std::endl;
+}
+
+int cmd_check(std::string local_repo_path) {
+  boost::log::core::get()->set_filter(boost::log::trivial::severity >= boost::log::trivial::info);
+  bool is_local_update{false};
+  LocalUpdateSource local_update_source;
+
+  std::vector<boost::filesystem::path> cfg_dirs;
```
mike-sul, line 110: Do you think it is possible and makes sense to extract the `AkliteClient` instantiation into a separate function and invoke the client initialization function from each `cmd_*` functions. (I mean lines 106 - 132).

src/cli/cli.cc
```diff
@@ -96,8 +106,20 @@ StatusCode Install(AkliteClient &client, int version, const std::string &target_
...
-    LOG_ERROR << "No Target found; version: " << (version == -1 ? "latest" : std::to_string(version))
-              << ", hardware ID: " << client.GetConfig().get("provision.primary_ecu_hardware_id", "")
+    std::string target_string;
+    if (version == -1 && target_name.empty()) {
+      target_string = " version: latest,";
+    } else {
+      if (!target_name.empty()) {
+        target_string = " name: " + target_name + ",";
+      }
+      if (version != -1) {
```
mike-sul, line 116: This `if` is redundant. Maybe the ternary operator `<>?<>:<>` will work a bit better?

mike-sul:
LGTM, I just have a few minor comments. Let's finalize it and do e2e testing.

examples/custom-client-cxx/cli.cpp
detsch: At first, my idea was to avoid calling additional functions, to make each `cmd_` more self contained. But once the code has been changed to use the cli api, the boilerplate code ended up being much bigger than the actual relevant code. So, yes, I believe it is now better to have a initialization function that is shared among `cmd_*`. Already prepared it, will force push in a bit.

src/cli/cli.cc
mike-sul: please, ignore, I misunderstood the code in the first place.

examples/custom-client-cxx/cli.cpp
detsch: Changes pushed

mike-sul (approved):

mike-sul (approved):

PR #303, opened 2024-02-08
main: Unify locking mechanism, using the same mechanism as the API

detsch:
The aktualizr-lite executable was using a different locking mechanism if compared to the one used in the aklite API, allowing concurrent execution between the two methods. This commit changes the aktualizr-lite executable so that it uses the mechanism originally used in the API, which relies on flock(LOCK_EX).

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1,3 +1,4 @@
+#include <sys/file.h>
 #include <iostream>
 #include <thread>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -418,11 +426,7 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
 }
 
 static int cli_install(LiteClient& client, const bpo::variables_map& params) {
-  // Make sure no any other update instances are running, i.e. neither the daemon or the other CLI update/finalize is
-  // running
-  // The API's AkliteClient uses different type of file locking, so it cannot be used for syncing of the daemon and the
-  // CLI command running. Specifically, the API's lock relies on "#define LOCK_EX 2 /* Exclusive lock.  */" while the
-  // daemon's lock is based on "# define F_WRLCK 1 /* Write lock.  */"
+  // Make sure no other update instances are running, i.e. neither the daemon or other CLI update/finalize
   FileLock lock;
```
mike-sul, line 430: I think this becomes redundant now as the `AkliteClient` ctor at #445 uses the same locking mechanism as the daemon.

mike-sul: However, let's leave it, it won't harm and prevent the `LiteClient` instance creation, it may do some "bad" stuff in its ctor.

mike-sul (approved):
LGTM, will test it later.

mike-sul:
For some reason `test_aktualizr-lite` fails with

```
Failed to convert the input target version to integer, consider it as a name of Target: stoi
Failed to obtain the aklite lock, another instance must be running !!!
+ cleanup
```

detsch:
> We need to switch `apply_lock` to `false` instead of `true` when instantiating AkliteClient:

Or we can drop `FileLock lock` in these two functions, like you mentioned before.

mike-sul:
> > We need to switch `apply_lock` to `false` instead of `true` when instantiating AkliteClient:
> 
> Or we can drop `FileLock lock` in these two functions, like you mentioned before.

Let's keep it there, so the `LiteClient` ctor is not invoked.

mike-sul (approved):

PR #306, opened 2024-02-19
Differentiate error codes in CheckIn status, and allow OkCached in offline mode

detsch:


src/api.cc
```diff
@@ -353,7 +354,20 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
   } catch (const Uptane::SecurityException& exc) {
     const std::string err{exc.what()};
     LOG_ERROR << "Failed to update the local TUF repo SecurityException: " << err;
-    return CheckInResult{CheckInResult::Status::SecurityError, "", {}};
+    if (err.find("Rollback attempt") != std::string::npos) {
+      LOG_INFO << "Checking the local TUF repo...";
+      try {
+        tuf_repo_->checkMeta();
```
mike-sul, line 360: If we wanna add this code to v93 then I would remove it (the code is fine). I think we should do `checkMeta` only after verification of the additional file `install.targets.json` expiration time. It will be added in v94.

But if we are not going to add this code to v93 then this code lgtm.

include/aktualizr-lite/cli/cli.h
```diff
@@ -32,6 +32,9 @@ enum class StatusCode {
   InstallRollbackOk = 110,
   InstallRollbackNeedsReboot = 120,
   InstallRollbackFailed = 130,
+  CheckinSecurityError = 140,
```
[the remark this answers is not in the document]
detsch: Done. Updated to 11, 12, 13.

src/api.cc
```diff
@@ -333,8 +347,25 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
   LOG_INFO << "Updating the local TUF repo with metadata located in " << local_update_source->tuf_repo << "...";
   try {
     tuf_repo_->updateMeta(repo_src);
+    LOG_INFO << "The local TUF repo has been successfully updated";
+  } catch (const Uptane::SecurityException& exc) {
+    const std::string err{exc.what()};
+    LOG_ERROR << "Failed to update the local TUF repo SecurityException: " << err;
```
mike-sul, line 354: Maybe it makes sense to use something more user-friendly instead of `SecurityException`, e.g. `Failed to update the local TUF repo; TUF meta check failure: << err`?

src/api.cc
```diff
@@ -333,8 +347,25 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
...
+    return CheckInResult{CheckInResult::Status::SecurityError, "", {}};
+  } catch (const Uptane::ExpiredMetadata& exc) {
+    const std::string err{exc.what()};
+    LOG_ERROR << "Failed to update the local TUF repo ExpiredMetadata: " << err;
```
mike-sul, line 371: `LOG_ERROR << "Failed to update the local TUF repo; TUF meta are expired: " << err;`

src/api.cc
mike-sul: I tend towards breaking this PR into two:

1. The PR that has the TUF error specialization -  it should go to v93.
2. The PR based on the PR #1 that includes the local repo check if the offline bundle has lower version meta. This one will go to v94.

WDYT?

detsch: Sure. If you believe it is worth to add the additional return codes now, that is fine with me.

Link on the comment is wrong, but I got the idea. Will create a separate PR for the second patch.

mike-sul: sorry, I meant the PR described in the first item.

mike-sul: I think it's worth adding additional error codes now because I believe the specific error messages will decrease amount of support work. Quite often customers send a request to fix an issue and the error message just says `Failed to update Image repo metadata`. Having more specific error message can help to identify the issue faster.

detsch:
> @detsch I think it would be great to catch those specific exceptions and print corresponding error messages in the liteclient implementation:
> 
> https://github.com/foundriesio/aktualizr-lite/blob/b2140a50e48537680f5832141141d6db579c6e00/src/liteclient.cc#L546
> 
> https://github.com/foundriesio/aktualizr-lite/blob/b2140a50e48537680f5832141141d6db579c6e00/src/liteclient.cc#L566
> 
> (This one for v93)

Done.

src/api.cc
```diff
@@ -333,8 +347,25 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
...
+    return CheckInResult{CheckInResult::Status::ExpiredMetadata, "", {}};
+  } catch (const Uptane::MetadataFetchFailure& exc) {
+    const std::string err{exc.what()};
+    LOG_ERROR << "Failed to update the local TUF repo MetadataFetchFailure: " << err;
```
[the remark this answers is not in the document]
detsch: Done.

src/api.cc
detsch: Done.

src/api.cc
detsch: Done.

src/api.cc
detsch: Moved the second commit to a new branch, on top of the one from this PR

detsch:
Switching back to draft as I add a few unit tests.

examples/custom-client-cxx/cli.cpp
```diff
@@ -26,14 +26,23 @@ static void print_status(aklite::cli::StatusCode ret) {
       std::cout << "SUCCESS: Unable to fetch updated TUF metadata, but stored metadata is valid";
       break;
     case aklite::cli::StatusCode::CheckinFailure:
-      std::cout << "FAILURE: Unable to get TUF metadata";
+      std::cout << "FAILURE: Error handling TUF metadata";
```
mike-sul, line 29: Maybe `Failed to update` instead of `Error handling`, "handling" sounds a bit vague ?

mike-sul (approved):

examples/custom-client-cxx/cli.cpp
detsch: Adjusted.

PR #307, opened 2024-02-23
offline: Add --target parameter to aklite-offline install

detsch:
This allow a specific target in a offline installation source dir to be selected for installation. The one with a higher version is still selected by default if the parameter is omitted.


mike-sul (approved):
LGTM. Please, test it manually.

PR #308, opened 2024-02-26
api: Use db metadata when offline pack has outdated TUF metadata

detsch:
Offline checkin was never falling back to using the current stored metadata. Now, if the offline checkin error is specifically related to try to process outdated metadata, i.e., there is already a more recent version in the local storage, we return an OK status (OkCached), allowing an offline update attempt to proceed.


mike-sul:
I closed this PR in favor of this commit https://github.com/foundriesio/aktualizr-lite/pull/316/commits/f68a52b10641168c48dad5646c90ba46fcc9c586 in the other PR.

PR #309, opened 2024-03-04
appengine: Don't require connectivity when starting app

mike-sul:
The goal of the change is to address a need for connection with Registry during app startup.
To achieve this, the following steps are taken:

- Images are loaded into the Docker store immediately after being pulled onto a device (into the Skopeo's image store).

- `docker compose pull` is invoked after the images have been loaded, it updates Docker engine's DB with digest references to the images.

As result, the subsequent call `docker compose up -d` does NOT require a connection with Registry, and apps can be started when a device is offline.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -124,6 +124,21 @@ AppEngine::Result RestorableAppEngine::fetch(const App& app) {
     const auto images_dir{app_dir / "images"};
     LOG_DEBUG << app.name << ": downloading App images from Registry(ies): " << app.uri << " --> " << images_dir;
     pullAppImages(uri, app_compose_file, images_dir);
+
+    // Install compose app to the `install_root` (`/var/sota/compose-apps`) and
+    // load app images to the docker engine store.
+    installAppAndImages(app);
+    // The docker engine's API doesn't allow to load digest referenced images (an uri with sha256 hash).
+    // Only tag referenced images can be loaded into the docker store.
+    // It leads to a need for connectivity to Registry when `docker compose up` is invoked
+    // and the app's `docker-compose.yml` contains digest references to images.
+    // To overcome this issue we need to invoke `docker compose pull` after the app image loading,
+    // so the docker engine pulls the image index/manifest and updates its "image DB" with
+    // digest references to the app's images.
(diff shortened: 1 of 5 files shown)


src/docker/restorableappengine.cc
```diff
@@ -124,6 +124,21 @@ AppEngine::Result RestorableAppEngine::fetch(const App& app) {
...
+    installAppAndImages(app);
+    // The docker engine's API does't allow to load digest referenced images (an uri with sha256 hash).
+    // Only tag referenced images can be loaded into the docker store.
+    // It leads to a need in connectivity to Registry when `docker compose up` is invoked
+    // and the app's `docker-compose.yml` contains digest references to images.
+    // To overcome this issue we need to invoke `docker compose pull` after the app image loading,
+    // so the docker engine pulls the image index/manifest and updates its "image DB" with
+    // digest references to the app's images.
+    // After that, the `install` or `run` phase (`docker compose up`) does not require connectivity/networking to start
+    // the app.
```
detsch, line 139:
```suggestion
    // The docker engine's API doesn't allow to load digest referenced images (an uri with sha256 hash).
    // Only tag referenced images can be loaded into the docker store.
    // It leads to a need for connectivity to Registry when `docker compose up` is invoked
    // and the app's `docker-compose.yml` contains digest references to images.
    // To overcome this issue we need to invoke `docker compose pull` after the app image loading,
    // so the docker engine pulls the image index/manifest and updates its "image DB" with
    // digest references to the app's images.
    // After that, the `install` or `run` phase (`docker compose up`) does not require connectivity/networking to start
    // the app.
```

src/docker/restorableappengine.cc
```diff
@@ -124,6 +124,21 @@ AppEngine::Result RestorableAppEngine::fetch(const App& app) {
...
+    pullComposeAppImages(compose_cmd_, install_root_ / app.name);
```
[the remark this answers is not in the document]
mike-sul: No, this is the `fetch()` method which implies a need in connectivity.

detsch (approved):
LGTM

src/docker/restorableappengine.cc
mike-sul: fixed.

PR #310, opened 2024-03-07
Enable events and callbacks for offline updates

detsch:


mike-sul (approved):

src/api.cc
```diff
@@ -698,10 +696,7 @@ class LocalLiteInstall : public LiteInstall {
 
   // TODO: implement `Install()` in such the way that `LiteClient` is not required as it is done for `Download()`
   InstallResult Install() override {
-    const auto tls_server{client_->config.tls.server};
-    client_->config.tls.server = "";
```
mike-sul, line 702: I think with this change it doesn't make sense to override the `Install()` method at all, so the `LocalLiteInstall::Install()` can be removed at all.

detsch: Removed.

detsch:
There is one conceptual issue when we look at when does `aktualizr-lite` daemon invokes the `check-for-update-post` callback. It is not right after the TUF check in operation, but after one specific target is selected, and additional checks (`is not rollback` and `is not the current target`) are performed.

In the API, as implemented now, the callback is invoked earlier, right at the end of the TUF check-in operation, which returns a list of targets. For now, in offline mode and when using the API, no target will be passed to the `callback` method and the `INSTALL_TARGET_NAME` env variable will be empty.

We might want to review that logic, to decide if we keep it that way, or if we change the implementation to increase the consistency between aklite daemon and API behavior.

src/liteclient.h
```diff
@@ -88,7 +88,7 @@ class LiteClient {
   FRIEND_TEST(AkliteTest, RollbackIfAppsInstallFails);
   FRIEND_TEST(AkliteTest, RollbackIfAppsInstallFailsAndPowerCut);
 
-  void callback(const char* msg, const Uptane::Target& install_target, const std::string& result = "");
+  virtual void callback(const char* msg, const Uptane::Target& install_target, const std::string& result);
```
mike-sul, line 91: A bit odd that we have to make it "virtual" just for sake of testing. I wonder if it is possible to test it without making this method virtual?

detsch: It makes testing much harder. The production code would have to be changed even more if we don't make the method virtual and still want to mock it (a template would have to be used). Since the performance penalty is negligible, I believe it is ok to make this concession in production code to favor testing.

mike-sul: ok.

src/api.cc
```diff
@@ -690,7 +690,10 @@ class LocalLiteInstall : public LiteInstall {
     client_->logTarget("Downloading: ", *target_);
 
     auto downloader = createOfflineDownloader();
+    client_->notifyDownloadStarted(*target_, reason);
```
mike-sul, line 717: It looks weird to me each time I look into it. What's the point to send a callback to a caller from the method/function invoked by the caller. It may lead to lots of issues in the future...

I think that's fine to have callbacks for the CLI tools and the daemon but not for the API. Unfortunately, it is not easy to implement, but we have to think about it.

mike-sul:
LGTM, will test it e2e in my factory.

PR #312, opened 2024-03-18
tuf: Add GetRoot method to TUF API

detsch:
Also add an implementation to AkRepo class.

---
@mike-sul not sure if that is what you need.

### DIFF
--- a/include/aktualizr-lite/tuf/tuf.h
+++ b/include/aktualizr-lite/tuf/tuf.h
@@ -29,10 +29,10 @@ class RepoSource {
   RepoSource& operator=(const RepoSource&) = delete;
   RepoSource& operator=(const RepoSource&&) = delete;
 
-  virtual std::string fetchRoot(int version) = 0;
-  virtual std::string fetchTimestamp() = 0;
-  virtual std::string fetchSnapshot() = 0;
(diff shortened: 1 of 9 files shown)


include/aktualizr-lite/tuf/tuf.h
```diff
@@ -197,6 +197,7 @@ class Repo {
   Repo& operator=(const Repo&&) = delete;
 
   virtual std::vector<TufTarget> GetTargets() = 0;
+  virtual std::string GetLatestRoot() = 0;
```
mike-sul, line 200: Can we name it as `getRoot(int version)`? If `version` equals to `-1` then return the latest version.

mike-sul:
> not sure if that is what you need.

That's what I need, specifically I need targets role public keys.

include/aktualizr-lite/tuf/tuf.h
mike-sul: Have you considered making it `GetRoot(int version)` so it cover all use-cases?

mike-sul: Yes, it LGTM, please go ahead, test, and merge it.

mike-sul (approved):
Just rebase and test it.

PR #313, opened 2024-03-22
Fix/send empty app header

mike-sul:
It turned out that libcurl requires setting a header value as <header-name>; to send the header with an empty value in a http request as <header-value>:\n.
This change makes sure that a header is set in libcurl correctly if the header value is empty.
The previous version set it as <header-name>: which led to not sending the header in a request at all - libcurl simply removes such header from a list of request headers.


vkhoroz (approved):
LGTM

detsch (approved):
LGTM

PR #316, opened 2024-04-15
Offline update: checking bundle metadata

mike-sul:
Verify a bundle metadata before updating a local/device TUF repo:
- check its signatures;
- check a bundle type (ci, prod), and whether it matches the device's one;
- check a bundle tag, and whether it matches the device's one;
- check whether the bundle targets are listed in the TUF targets of the device TUF repo.

Update and extend the unit tests to support the new use-cases.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -141,8 +141,7 @@ static std::vector<TufTarget> filterTargets(const std::vector<TufTarget>& allTar
                                             const std::vector<std::string>& secondary_hwids) {
   std::vector<TufTarget> targets;
   for (const auto& t : allTargets) {
-    if (!tags.empty() /* Should we really allow a device configuration without a tag? */
-        && !t.HasOneOfTags(tags)) {
+    if (!tags.empty() && !t.HasOneOfTags(tags)) {
       continue;
     }
     if (t.HardwareId() == hwidToFind) {
@@ -419,10 +418,11 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
 
   LOG_INFO << "The local TUF repo has been successfully updated";
   LOG_INFO << "Searching for matching TUF Targets...";
-  auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, client_->tags, secondary_hwids_);
+  auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, {}, secondary_hwids_);
   if (matchingTargets.empty()) {
     err_msg =
-        "Couldn't find Targets matching the current device tag and hardware ID; check a device tag or a hardware ID";
+        "Couldn't find Targets matching the device's hardware ID; check a tag or a hardware ID of the device and the "
+        "bundle's tag";
     LOG_ERROR << err_msg;
     return CheckInResult{CheckInResult::Status::NoMatchingTargets, "", {}};
   }
--- a/src/api.cc
+++ b/src/api.cc
@@ -22,6 +22,26 @@
 #include "tuf/localreposource.h"
 #include "uptane/exceptions.h"
 
+class BundleMetaError : public std::logic_error {
+ public:
+  enum class Type {
+    NotFound,
+    RootMetaNotFound,
+    InvalidMetadata,
+    BadSignature,
+    UnmetSignatureThreshold,
+    Expired,
+  };
+  BundleMetaError(Type type_in, const std::string& err_msg) : std::logic_error(err_msg), type{type_in} {}
+  Type type;
+};
+
+static Json::Value checkAndGetBundleMeta(const std::shared_ptr<aklite::tuf::Repo>& device_tuf_repo,
+                                         const boost::filesystem::path& bundle_tuf_repo_path);
+static void printBundleMeta(const Json::Value& bundle_meta);
+static std::vector<TufTarget> getTrustedBundleTargets(const std::shared_ptr<aklite::tuf::Repo>& tuf_repo,
+                                                      const Json::Value& bundle_meta);
+
 const std::vector<boost::filesystem::path> AkliteClient::CONFIG_DIRS = {"/usr/lib/sota/conf.d", "/var/sota/sota.toml",
                                                                         "/etc/sota/conf.d/"};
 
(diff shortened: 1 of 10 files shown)


src/api.cc
```diff
@@ -963,3 +1015,138 @@ InstallResult AkliteClient::SetSecondaries(const std::vector<SecondaryEcu>& ecus
...
+    }
+  }
+
+  const int targets_sign_threshold{root_meta["signed"]["roles"]["targets"]["threshold"].asInt()};
+
+  int found_valid_sig_number{0};
+  const auto bundle_meta_path{bundle_tuf_repo_path / "bundle-targets.json"};
+  if (!boost::filesystem::exists(bundle_meta_path)) {
+    throw BundleMetaError(BundleMetaError::Type::NotFound,
+                          "Failed to find the bundle metadata: " + bundle_meta_path.string());
```
vkhoroz, line 1094: Looking at this change I am under the impression that it _requires_ a `bundle-targets.json` to be present in a bundle.

But, there are already bundles used generated and used by our users without this file.
Don't you think we need to continue supporting those bundles?

src/liteclient.cc
```diff
@@ -836,3 +827,22 @@ bool LiteClient::isRegistered(const KeyManager& key_manager) {
...
+    const auto bc{key_manager.getBC()};
+    if (bc == prod_bc_value) {
+      res = Type::Prod;
+      LOG_DEBUG << "Found production business category in a device certificate: " << bc;
+    } else {
+      res = Type::Dev;
+      LOG_DEBUG << "Missing or non-production business category found in a device certificate: " << bc;
+    }
+  } catch (const std::exception& exc) {
+    LOG_WARNING << "A device certificate is not found or failed to parse it: " << exc.what();
```
vkhoroz, line 845: If there is no client certificate or we failed to parse it - how can the device connect to our APIs?
Shouldn't we just fail hard in this case?

src/api.cc
```diff
@@ -419,10 +418,11 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
 
   LOG_INFO << "The local TUF repo has been successfully updated";
   LOG_INFO << "Searching for matching TUF Targets...";
-  auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, client_->tags, secondary_hwids_);
+  auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, {}, secondary_hwids_);
```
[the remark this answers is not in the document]
mike-sul: >Having a tag configured on the device tells TUF to filter for specific targets.

It is not quite correct statement. The device tag tells to filter for specific set of TUF metadata - <tag> -> [timestamp, snapshot, targets].

And, the latter commit does verify if the bundle's tag matches the device's one. So, this check is kind of redundant.

src/api.cc
mike-sul: We do support them by the previous LmP/aklite versions until v94. If a device is already running the code with this check it means that a device is on v94. So, if the bundle was generated earlier it means that it has outdated TUF meta and target(s), and the update will fail anyway.

src/liteclient.cc
mike-sul: In the case of an offline update connection to the API is not required.

src/api.cc
```diff
@@ -320,6 +331,15 @@ static std::vector<Uptane::Target> getAvailableTargets(const PackageConfig& pcon
...
     auto custom_data{t.custom_data()};
+    if (!offline_app_shortlist.empty()) {
+      // If the app shortlist is ON then update the found target app list so the downloader
+      // and the installer will update only the apps specified in the shortlist.
+      Json::Value app_shortlist;
+      for (const auto& app : offline_app_shortlist) {
+        app_shortlist[app] = custom_data["docker_compose_apps"][app];
+      }
+      custom_data["docker_compose_apps"] = app_shortlist;
+    }
```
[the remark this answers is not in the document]
doanac: calling the function something like `getAvailableTargetsLocal` might reduce the cognitive load of reading this function.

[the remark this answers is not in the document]
mike-sul: The plan is to move all the offline/local specific implementation from the api.cc to `src/offline/*` directory and put it under `offline` namespace. Therefore, all those online/local prefixes/suffixes will become redundant.

detsch:
Code looks good IMO. I'll test it to see if I can find any issue.

src/api.cc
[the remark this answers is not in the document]
vkhoroz: Makes sense.

src/api.cc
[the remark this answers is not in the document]
vkhoroz: I've noticed that second check.
Given that, a check in this function makes no harm... it is in fact a kind of a sanity check, which is rather logical.

I usually prefer to have too many checks than too little checks, simply because it is more robust.

src/api.cc
vkhoroz: Hmm... interesting logic.
I think I agree to it; and my hope is that there is no leak in it.

src/liteclient.cc
vkhoroz: Just double checking.
Do you mean that having a device certificate is not required for offline device too?

That makes sense to me.

src/api.cc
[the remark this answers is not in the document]
mike-sul: I am actually kind on the fence about this check, that's why I implemented in the way allowing its easy removal. I think we need to discuss it offline, including customer support team (this repo is public).

src/api.cc
mike-sul: I don't have really strong opinion about it, just don't see real-life use-case for it. Also, it is kind of mechanism to enforce customers to update fioctl to the version that supports uptane-line bundles.

src/liteclient.cc
mike-sul: >Do you mean that having a device certificate is not required for offline device too?

Yes.

vkhoroz (commented):
LGTM

Once @doanac or @detsch approves, I'm fine with it.

detsch (approved):
I've done some testing, and the changes seems to be working as expected.

mike-sul:
> I've done some testing, and the changes seem to be working as expected.

Unfortunately, we have to re-test it since I amended the behavior slightly according to what we agreed offline/slack.

src/api.cc
```diff
@@ -922,3 +1034,165 @@ InstallResult AkliteClient::SetSecondaries(const std::vector<SecondaryEcu>& ecus
...
+  }
+  if (!bundle_targets_json["signatures"].isArray()) {
+    throw BundleMetaError(BundleMetaError::Type::InvalidMetadata,
+                          "The bundle metadata is invalid;"
+                          " the `signatures` field is not an array; got: \n" +
+                              bundle_targets_str);
+  }
+
+  const std::string signed_body{Utils::jsonToCanonicalStr(bundle_targets_json["signed"])};
+  LOG_INFO << "Checking the bundle metadata singatures (required: " << targets_sign_threshold << ")...";
```
detsch, line 1114:
```suggestion
  LOG_INFO << "Checking the bundle metadata signatures (required: " << targets_sign_threshold << ")...";
```

mike-sul: Fixed in the force push.

(I turned off the spell checker in my IDE because it eats too much CPU, it looks like I have to find another spell checker :)).

mike-sul:
@detsch I am going to merge it, I think we have tested it quite well.

detsch:
> @detsch I am going to merge it, I think we have tested it quite well.

Yep, should be good to go.

PR #317, opened 2024-04-24
fixup!: Don't pull app images if offline update

mike-sul:
Do not pull compose app images through the `docker compose pull` for sake of their registration in the docker store in case of the offline update. The offline update client does image registration through writing image digest URIs directly to the docker daemon's registration file. As result, no internet connectivity is required for this process.


detsch (approved):
LGTM

PR #319, opened 2024-05-09
Use composectl to check app state

mike-sul:
Use composectl to check app state:
1. whether app is fetched,
2. whether app is installed,
3. whether app is running.

Adjust unit tests to removing C++ code to check app state and using `composectl` instead.


detsch (approved):
LGTM

PR #321, opened 2024-05-24
fixup!: Check shortilst value before using it

mike-sul:
Make sure that the `shortlist` field is present, not `null`, and not empty string before using it - adding to the install target the apps that are in intersection of the tuf target app list and the `fetched-apps.shortlist` field.

This helps to overcome a problem if `fetched-apps.shortlist` is set to `""` by the ci-scripts if the `shortlist` field is not present in `preloaded_images` nor `app_shortlist` in `offline` sections of the factory config.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -46,6 +46,11 @@ static void checkBundleTag(const Json::Value& bundle_meta, const std::vector<std
 static std::vector<TufTarget> getTrustedBundleTargets(const std::vector<TufTarget>& tuf_targets,
                                                       const Json::Value& bundle_meta);
 
+enum class OfflineUpdateAppsShortlistType { AllTargetApps = 0, NoneOfTargetApps, OnlyShortlistedApps };
+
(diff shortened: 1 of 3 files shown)


src/api.cc
```diff
@@ -1196,3 +1219,41 @@ static void checkBundleTag(const Json::Value& bundle_meta, const std::vector<std
     }
   }
 }
+
+std::tuple<OfflineUpdateAppsShortlistType, std::set<std::string>> getAppsFetchedForOfflineUpdate(
```
detsch, line 1223: Since there is some non-trivial logic in here, adding one LOG entry (at least LOG_DEBUG) on each return statement, in line with the text in the comments, might help users to identify issues with their bundles.

detsch (approved):
LGTM

src/api.cc
mike-sul: I agree, will add logs to it, maybe even `info` logs and/or `warn`.

This is still WIP, I am going to add unit tests.

mike-sul: @detsch I added logs and unit tests.

detsch (approved):

PR #322, opened 2024-05-27
Fix app pruning and status checking if one or more of its elements are invalid

mike-sul:
- Make sure none of app manifest blobs are pruned during the app store pruning if app is in use.
- Handle exceptions of the app image manifest processing during app store pruning.
- Re-fetch app if app image index is invalid.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -652,73 +652,87 @@ bool RestorableAppEngine::areAppImagesFetched(const App& app) const {
                 << "; index: " << index_manifest;
       return false;
     }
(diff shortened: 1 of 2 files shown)


detsch (approved):
LGTM.

I suggest a few minor changes on commit https://github.com/foundriesio/aktualizr-lite/pull/322/commits/b0504b6fca81af6e89249d511d83f613cddb6736 description text. The last paragraph could be replaced with:
```
In addition to that, the app manifest and the app image manifest can be
stored in the blob directory. If so, make sure they are not removed 
during app store pruning.
```

PR #323, opened 2024-06-03
Make aklite daemon use API methods, unifying code among tools

detsch:
A new AkLiteClientExt class is introduced, centralizing some logic that was duplicated in aktualizr-lite daemon and the cli API. Behavior consistency between the aklite daemon and tools using the API was also improved, favoring the behavior of the daemon.

---

There is still a lot of testing to be done, specially in rollback situations.


src/api.cc
```diff
@@ -164,12 +164,16 @@ AkliteClient::~AkliteClient() {
...
 static bool compareTargets(const TufTarget& a, const TufTarget& b) { return a.Version() < b.Version(); }
 
-// Returns a sorted list of targets matching tags if configured and hwid (or one of secondary_hwids)
+// Returns a sorted list of OSTREE targets matching tags if configured and hwid (or one of secondary_hwids)
 static std::vector<TufTarget> filterTargets(const std::vector<TufTarget>& allTargets, const std::string& hwidToFind,
                                             const std::vector<std::string>& tags,
                                             const std::vector<std::string>& secondary_hwids) {
   std::vector<TufTarget> targets;
   for (const auto& t : allTargets) {
+    if (t.Custom()["targetFormat"] != "OSTREE") {
```
mike-sul, line 188: It's almost impossible for a non-OSTree target to slip through the overall target flow up to this point.
If it occurs then it means that something is really off, so I would print an error message here and printing target name along with the target format. Yet, still continue as you've done here.

src/api.cc
```diff
@@ -58,17 +58,29 @@ const std::vector<boost::filesystem::path> AkliteClient::CONFIG_DIRS = {"/usr/li
...
-      return *it;
+  if (version == -1 && target_name.empty()) {
+    for (auto it = targets_.crbegin(); it != targets_.crend(); ++it) {
+      if ((*it).Custom()["hardwareIds"][0] == hwid) {
+        return *it;
+      }
+    }
+  } else {
+    for (const auto& t : targets_) {
+      if (t.Version() == version || t.Name() == target_name) {
```
mike-sul, line 79: I think we need to match hwid too. There can be multiple targets of the same version but for different hardware ID.

src/api.cc
```diff
@@ -58,17 +58,29 @@ const std::vector<boost::filesystem::path> AkliteClient::CONFIG_DIRS = {"/usr/li
...
+        return t;
+      }
     }
   }
-  throw std::runtime_error("no target for this hwid");
+
+  LOG_INFO << "no target for hwid " << hwid;
+  return TufTarget();
```
mike-sul, line 86: Replacing the exception with returning of empty target may cause regression.
Is there particular need to return an empty target if not found?

src/api.cc
mike-sul: Actually, in the current implementation the `CheckInResult::targets_` are already filtered by hardware ID. So, we need either remove hwid checking for both cases, target name/version is specified and not or do checking for both cases.

tests/apiclient_test.cc
```diff
@@ -399,7 +399,7 @@ TEST_F(ApiClientTest, InstallDownloadInSeparateInstances) {
     auto result = client.CheckIn();
     ASSERT_EQ(CheckInResult::Status::Ok, result.status);
 
-    auto latest = result.GetLatest();
+    auto latest = result.SelectTarget();
```
mike-sul, line 403: If we replace all `GetLatest` calls with `SelectTarget` one then there is no test that verifies `GetLatest`. We need either remove `GetLatest` or have some of test invoking it.

include/aktualizr-lite/aklite_client_ext.h
```diff
@@ -0,0 +1,30 @@
...
+ * they are likely to have they signature and behavior changed.
+ */
+class AkliteClientExt : public AkliteClient {
+  // Inherit constructors
+  using AkliteClient::AkliteClient;
+
+ public:
+  CheckInResult GetTargetToInstall(const LocalUpdateSource* local_update_source, int version,
+                                   const std::string& target_name, const TufTarget& current, bool allow_bad_target,
+                                   bool force_apps_sync);
```
mike-sul, line 33: I would call it something like `allow_current_if_not_sync`.
I would probably introduce a bitwise `enum` to define type of targets to include into the target search and return.

include/aktualizr-lite/aklite_client_ext.h
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Those methods may eventually be part of the supported API but, for now,
+ * they are likely to have they signature and behavior changed.
+ */
+class AkliteClientExt : public AkliteClient {
+  // Inherit constructors
+  using AkliteClient::AkliteClient;
+
+ public:
+  CheckInResult GetTargetToInstall(const LocalUpdateSource* local_update_source, int version,
```
mike-sul: maybe it makes sense to make the `local_update_source` as the last parameter with a default  `nullptr` value?

include/aktualizr-lite/api.h
```diff
@@ -51,6 +52,9 @@ class CheckInResult {
   // NOLINTNEXTLINE(hicpp-explicit-conversions,google-explicit-constructor)
   operator bool() const { return status == CheckInResult::Status::Ok || status == CheckInResult::Status::OkCached; }
 
+  TufTarget selected_target = TufTarget();
```
mike-sul, line 56: I would introduce a new type and use it as a return type of `GetTargetToInstall`.
Extending the `CheckInResult` with these two fields looks like mixing fields/data of two different entities in one class.

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,146 @@
...
+                                                  const std::string& target_name, const TufTarget& current,
+                                                  bool allow_bad_target, bool force_apps_sync) {
+  // Tell CheckIn and CheckInLocal methods to not call notifyTufUpdateFinished in case of success
+  usingUpdateClientApi = true;
+  client_->setAppsNotChecked();
+  bool rollback_operation = false;
+
+  CheckInResult res = local_update_source == nullptr ? CheckIn() : CheckInLocal(local_update_source);
+
+  if (res.status != CheckInResult::Status::Ok && res.status != CheckInResult::Status::OkCached) {
```
mike-sul, line 48: It can just `if (!res)`.

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,146 @@
...
+    LOG_WARNING << "Unable to update latest metadata";
+    return res;
+  }
+
+  auto candidate_target = res.SelectTarget(version, target_name);
```
mike-sul, line 53: Just an idea. Maybe instead of `SelectTarget(version, target_name)` we can have two methods `cr.GetTarget(int version)` and `cr.GetTarget(string name)`. Passing these two params looks a bit awkward. And we have these two params in  `GetTargetToInstall` too...

Maybe we can encapsulate these two params in something like `struct TargetId { int Version; string Name}`?

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,146 @@
...
+  target_json["length"] = 0;
+  Uptane::Target target(t.Name(), target_json);
+  return target;
+}
+
+// Returns the target that should be installed, if any.
+// It might be an updated version, a rollback target, or even the currently installed target, in case we need to sync
+// apps
+CheckInResult AkliteClientExt::GetTargetToInstall(const LocalUpdateSource* local_update_source, int version,
+                                                  const std::string& target_name, const TufTarget& current,
```
mike-sul, line 39: Do we really need to pass `current`? The `AkliteClientEx` is supposed to know its state, and get the current target out of its context, I suppose.

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,146 @@
...
+    LOG_INFO << "The currently booted Target is a failing Target, finding Target to rollback to...";
+    const TufTarget rollback_target = Target::toTufTarget(client_->getRollbackTarget());
+    if (rollback_target.IsUnknown()) {
+      const auto err{boost::str(boost::format("Failed to find Target to rollback to after a failure to start "
+                                              "Apps at boot on a new version of sysroot;"
+                                              " failing current Target: %s, hash: %s") %
+                                current.Name() % current.Sha256Hash())};
+
+      LOG_ERROR << err;
+      res.status = CheckInResult::Status::Failed;  // TODO: Specify proper error status
```
mike-sul, line 74: It looks like we need to extend the checkinresult type, and use another return type for `GetTargetToInstall`.

src/api.cc
```diff
@@ -272,11 +272,13 @@ CheckInResult AkliteClient::CheckIn() const {
   LOG_INFO << "Searching for matching TUF Targets...";
   auto matchingTargets = filterTargets(tuf_repo_->GetTargets(), hw_id_, client_->tags, secondary_hwids_);
   if (matchingTargets.empty()) {
-    err_msg =
-        "Couldn't find Targets matching the current device tag and hardware ID; check a device tag or a hardware ID";
+    // TODO: consider reporting about it to the backend to make it easier to figure out
```
mike-sul, line 277: That's a good idea. However, if a device tag is invalid then DG returns 404, so we'll get the checkin error much earlier than this spot.

The idea is still interesting, it will require changing both the client and the backend sides, so I would create a ticket to consider it.

src/cli/cli.cc
```diff
@@ -45,7 +45,7 @@ static const std::unordered_map<InstallResult::Status, StatusCode> i2s = {
     {InstallResult::Status::NeedsCompletion, StatusCode::InstallNeedsReboot},
     {InstallResult::Status::AppsNeedCompletion, StatusCode::InstallAppsNeedFinalization},
     {InstallResult::Status::BootFwNeedsCompletion, StatusCode::InstallNeedsRebootForBootFw},
-    {InstallResult::Status::DownloadFailed, StatusCode::InstallAppPullFailure},
+    {InstallResult::Status::DownloadFailed, StatusCode::DownloadFailure},
```
mike-sul, line 48: It doesn't indicate whether its ostree or app pull failure.

mike-sul:
@detsch The direction LGTM to me. I am going to test it and then finalize my review.

mike-sul:
It seems like `test_aktualizr-lite` fails, the same true for my local env.

examples/custom-client-cxx/daemon.cpp
```diff
@@ -98,7 +98,7 @@ int cmd_daemon(std::string local_repo_path) {
         continue;  // There's no point trying to look for an update
       }
 
-      auto latest = res.GetLatest();
+      auto latest = res.SelectTarget();
```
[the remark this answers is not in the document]
detsch: yep, I though about that too. Originally, I've removed the GetLatest method, but then re-added it to keep backward compatibility. Since we are going to keep it, there is no need to do change GetLatest -> SelectTarget globally.

include/aktualizr-lite/aklite_client_ext.h
detsch: The method always returns the current target if the apps are not in sync. That's the behavior of the original daemon, it checks if apps are in sync on every iteration, doing an install if they are not.
`force_apps_sync=true` means that we return the current one even if the apps are in sync. That's the behavior for offline updates.

include/aktualizr-lite/api.h
detsch: Yep, makes sense. I was avoiding the creation of new types, by I agree that having these new fields might bring confusion to users of the original types.

src/aklite_client_ext.cc
mike-sul: I think we should think about optimizing/caching of the current target in the context of the lite client (out of scope of this PR).

doanac:
image

src/api.cc
detsch: I'm re-adding the `throw` statement to the `GetLatest` method, so that users of the original method will still have the original behavior.
I prefer to keep `SelectTarget` returning empty instead of throwing an exception if no matching target is found. It may be just a matter of a given target name not being in the list, which is a regular situation, not an exception. And the caller code gets also a little cleaner.
But is easy to change that if we believe it is better to have an exception. I've reverted the original `GetLatest` calls back to the original name, so there is only two places where `SelectTarget` is currently used: inside `GetLatest` and `GetTargetToInstall`.

examples/custom-client-cxx/daemon.cpp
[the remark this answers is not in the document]
detsch: Done, I've reverted original `GetLatest` calls to the original method.

src/api.cc
detsch: Done.

detsch:
> It seems like `test_aktualizr-lite` fails, the same true for my local env.



> It seems like `test_aktualizr-lite` fails, the same true for my local env.

Fixed.

src/api.cc
mike-sul: I don't have strong opinion on it. The current version LGTM.

src/api.cc
mike-sul: I understand, still idea is interesting, I am going to create ticket for it.

src/helpers.h
```diff
@@ -6,6 +6,7 @@
 #include "liteclient.h"
 
 void generate_correlation_id(Uptane::Target& t);
```
mike-sul, line 8: I guess if we add return type to the existing function then it won't break existing clients because they don't expect any return value anyway.

include/aktualizr-lite/api.h
detsch: I've created a new GetTargetToInstallResult with the new fields, and its own Status enum.

src/aklite_client_ext.cc
detsch: Done

src/aklite_client_ext.cc
detsch: I've added the `GetCurrent` call inside `GetTargetToInstall` and removed the `current` parameter

src/aklite_client_ext.cc
detsch: This specific situation was only being handled in the aklite daemon. Should we add a new status now (and documentation) for it to be identifiable for cli users? Something like `NoTargetToRollbackTo`?

src/cli/cli.cc
detsch: I've added a new Status for the ostree download failure, reverting the change on `{InstallResult::Status::DownloadFailed, StatusCode::InstallAppPullFailure}` mapping.

src/api.cc
detsch: I'm now checking for hwid when version or name is specified too. So, it is now always checked.

src/helpers.h
detsch: Key change was the input parameter, changed from `Uptane::Target` to `TufTarget`. The method is only called inside aklite daemon main loop, and the original version is not used after the changes in this PR. I've now removed it, keeping only the new method.

include/aktualizr-lite/aklite_client_ext.h
detsch: All parameters now have default values, and on the typical case, no parameters would be passed when calling the method (the `current` parameter was removed).
I tend to keep the order we have now, leaving `local_update_source` as the first one. In the scenario where the parameters are explicitly set (inside `cli.cc`), `local_update_source` seems to be the most relevant.

tests/apiclient_test.cc
detsch: `GetLatest` is now tested again, as I reverted most of the `GetLatest` -> `SelectTarget` changes.

src/api.cc
```diff
@@ -59,16 +59,34 @@ const std::vector<boost::filesystem::path> AkliteClient::CONFIG_DIRS = {"/usr/li
...
   if (hwid.empty()) {
     hwid = primary_hwid_;
   }
 
-  for (auto it = targets_.crbegin(); it != targets_.crend(); ++it) {
-    if ((*it).Custom()["hardwareIds"][0] == hwid) {
-      return *it;
+  if (version == -1 && target_name.empty()) {
+    for (auto it = targets_.crbegin(); it != targets_.crend(); ++it) {
+      if ((*it).Custom()["hardwareIds"][0] == hwid) {
```
mike-sul, line 79: Let's use the `HardwareId()` here too.

include/aktualizr-lite/aklite_client_ext.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include "aktualizr-lite/tuf/tuf.h"
+
+#include "aktualizr-lite/api.h"
+
+class GetTargetToInstallResult {
+ public:
+  //
+  enum class Status {
+    // First block must match CheckInResult::Status
```
mike-sul, line 17: It's a pain that we have to define all `CheckInResult::Status` enum values again.

include/aktualizr-lite/aklite_client_ext.h
```diff
@@ -0,0 +1,64 @@
...
+    Ok = 0,    // check-in was good
```
mike-sul, line 18: Maybe, doing `Ok = static_cast<int>(CheckInResult::Status::Ok)` will be more reliable.

include/aktualizr-lite/aklite_client_ext.h
```diff
@@ -0,0 +1,64 @@
...
+    // Additional values, specific for GetTargetToInstallResult
+    TufTargetNotFound = 100,
+  };
+
+  explicit GetTargetToInstallResult(const CheckInResult &check_in_result)
+      : status(static_cast<Status>(check_in_result.status)) {}
+
+  GetTargetToInstallResult(Status status, TufTarget selected_target, std::string reason)
+      : status(status), selected_target(std::move(selected_target)), reason(std::move(reason)) {}
+  Status status;
```
mike-sul, line 38: Just idea. Maybe having two status fields in the `GetTargetToInstallResult` context would work better? Something like:


```
CheckInResult::Status checkin_status;
Status                find_target_status;
```

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,143 @@
...
+  target_json["custom"]["targetFormat"] = "OSTREE";
+  target_json["length"] = 0;
+  Uptane::Target target(t.Name(), target_json);
+  return target;
+}
+
+// Returns the target that should be installed, if any.
+// It might be an updated version, a rollback target, or even the currently installed target, in case we need to sync
+// apps
+GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const LocalUpdateSource* local_update_source, int version,
```
mike-sul, line 38: I would consider:

1. Making `local_update_source` the last parameter.
2. Defining bitwise enumeration for  `allow_bad_target` and `force_apps_sync`, so we have a single input parameter and the target select modes can be easily extended and combined through the bitwise OR.

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,143 @@
...
+    if (rollback_target.IsUnknown()) {
+      const auto err{boost::str(boost::format("Failed to find Target to rollback to after a failure to start "
+                                              "Apps at boot on a new version of sysroot;"
+                                              " failing current Target: %s, hash: %s") %
+                                current.Name() % current.Sha256Hash())};
+
+      LOG_ERROR << err;
+      client_->notifyTufUpdateFinished(err);
+      // TODO: Specify proper error status
+      return {GetTargetToInstallResult::Status::Failed, TufTarget(), err};
```
mike-sul, line 76: RollbackTargetNotFound?

src/aklite_client_ext.cc
```diff
@@ -0,0 +1,143 @@
...
+    res.reason = std::string(rollback_operation ? "Rolling back" : "Updating") + " from " + current.Name() + " to " +
+                 res.selected_target.Name();
+
+  } else {
+    if (is_bad_target) {
+      LOG_INFO << "Target: " << candidate_target.Name() << " is a failing Target (aka known locally)."
+               << " Skipping its installation.";
+    }
+
+    auto current_target = std::make_unique<Uptane::Target>(client_->getCurrent());
```
mike-sul, line 123: Why do we need to get the current at this point if we got one at line 61?

src/cli/cli.cc
```diff
@@ -88,64 +103,45 @@ static StatusCode pullAndInstall(AkliteClient &client, int version, const std::s
...
       LOG_ERROR << "Downgrade is not allowed by default, re-run the command with `--force` option to force downgrade";
       return StatusCode::InstallDowngradeAttempt;
     }
-    LOG_WARNING << "Downgrading from " << current.Version() << " to  " << target.Version() << "...";
-  }
-
-  // Check whether the given target is already installed and synced/running
-  if (current == target && client.CheckAppsInSync() == nullptr) {
-    if (local_update_source != nullptr) {
-      return StatusCode::InstallAlreadyInstalled;
```
mike-sul, line 146: It looks like we lost `InstallAlreadyInstalled` case?

src/helpers.h
mike-sul: OK. I would just make the param const: `const TufTarget&`.

src/helpers.h
```diff
@@ -6,6 +6,7 @@
...
+std::string generate_correlation_id(TufTarget& t);
```
mike-sul, line 8: Another question/topic, not related to this PR:
1. Can we make cor ID optional paramater and auto-generate inside `PullAndInstall`?
2. How a user can "know" what is cor ID if he/she/it calls pull and install separately?

src/aklite_client_ext.cc
```diff
@@ -239,7 +239,9 @@ bool AkliteClientExt::RebootIfRequired() {
   auto is_reboot_required = client_->isRebootRequired();
 
   if (is_reboot_required.first) {
-    if (is_reboot_required.second.empty()) {
+    if (is_booted_env) {
```
mike-sul, line 242: I think it should be `if (!is_booted_env)`.

src/api.cc

src/aklite_client_ext.cc
detsch: Done

src/cli/cli.cc
detsch: I've re-added it now

src/aklite_client_ext.cc
detsch: Should not be needed. I've just fixed it.

src/aklite_client_ext.cc
detsch: Added `RollbackTargetNotFound` return code.

mike-sul: Not sure I see it.

src/main.cc
```diff
@@ -242,185 +152,41 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
     return EXIT_FAILURE;
   }
 
-  client.importRootMetaIfNeededAndPresent();
-  client.finalizeInstall();
+  std::shared_ptr<LiteClient> client_ptr{&client, [](LiteClient* /*unused*/) {}};
+  // Setting apply_lock parameter to false, since we already took the FileLock above
```
mike-sul: Maybe we should the AkliteClient's lock for sake of consistency and remove the one at line 145.

mike-sul: It was done so in the next commit.

src/aklite_client_ext.cc
detsch: Is at this same location.

https://github.com/foundriesio/aktualizr-lite/blob/f2e2221f071125e2f32de4e1784aca86f863a36a/src/aklite_client_ext.cc#L66

src/helpers.h
detsch: > * Can we make cor ID optional paramater and auto-generate inside `PullAndInstall`?
Yes, that makes sense. Right now an empty `correlation_id` leads to it being auto-generated inside the `Installer` constructor. I think it is worth to set some default values in `PullAndInstall` in the current PR, I'll do it.

> * How a user can "know" what is cor ID if he/she/it calls pull and install separately?
Currently I believe there is no way, the user would have to generate and pass the same ID to both calls. Install finalization falls in a similar situation but, in that case, there is an entry in `installed_versions` table where the cor ID is fetched from.

src/aklite_client_ext.cc
detsch: Code was adjusted to return `RollbackTargetNotFound` in that case.

mike-sul (approved):
LGTM

include/aktualizr-lite/aklite_client_ext.h
detsch: Done.

mike-sul (approved):

PR #325, opened 2024-06-06
[RFC] bootloader: add support for env variables

igoropaniuk:
Add support for providing environment variables when calling rollback-related utils like fw_setenv/fw_printenv/ fiovb_setenv/fiovb_printenv.

Provide by default FIO_OSTREE_TARGET_SYSROOT, which contains a path to target ostree sysroot.

### DIFF
--- a/src/bootloader/bootloaderlite.cc
+++ b/src/bootloader/bootloaderlite.cc
@@ -43,6 +43,8 @@ void BootloaderLite::installNotify(const Uptane::Target& target) const {
       }
       const std::string& target_version{std::get<0>(target_version_val)};
 
+      env_cmd_vars_ =  boost::str(boost::format("%s=%s") % OstreeTargetPathVar %
+                                  getTargetDir(sysroot_->deployment_path(), target.sha256Hash()));
       std::string cur_version;
       bool is_current_ver_valid;
       std::tie(cur_version, is_current_ver_valid) = getCurrentVersion();
@@ -93,17 +95,8 @@ BootloaderLite::VersionStrRes BootloaderLite::getDeploymentVersion(const std::st
 std::string BootloaderLite::getVersion(const std::string& deployment_dir, const std::string& hash,
                                        const std::string& ver_file) {
   try {
-    std::string file;
-    for (auto& p : boost::filesystem::directory_iterator(deployment_dir)) {
-      std::string dir = p.path().string();
-      if (!boost::filesystem::is_directory(dir)) {
-        continue;
-      }
-      if (boost::algorithm::contains(dir, hash)) {
-        file = dir + ver_file;
-        break;
-      }
-    }
+    std::string file = getTargetDir(deployment_dir, hash) + ver_file;
+
     if (file.empty()) {
       LOG_WARNING << "Target deployment hash was not found";
       return std::string();
@@ -121,6 +114,21 @@ std::string BootloaderLite::getVersion(const std::string& deployment_dir, const 
     LOG_ERROR << "Failed to get Target firmware version:  " << exc.what();
     return "";
   }
+}
+
+std::string BootloaderLite::getTargetDir(const std::string& deployment_dir,
+                                         const std::string& target_hash) {
+  for (auto& p : boost::filesystem::directory_iterator(deployment_dir)) {
+    std::string dir = p.path().string();
+    if (!boost::filesystem::is_directory(dir)) {
+      continue;
+    }
+    if (boost::algorithm::contains(dir, target_hash)) {
+      return dir;
+    }
+  }
+
+  return "";
 }
 
 bool BootloaderLite::isUpdateInProgress() const {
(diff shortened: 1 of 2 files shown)


src/bootloader/bootloaderlite.cc
```diff
@@ -93,17 +95,8 @@ BootloaderLite::VersionStrRes BootloaderLite::getDeploymentVersion(const std::st
...
-      std::string dir = p.path().string();
-      if (!boost::filesystem::is_directory(dir)) {
-        continue;
-      }
-      if (boost::algorithm::contains(dir, hash)) {
-        file = dir + ver_file;
-        break;
-      }
-    }
+    std::string file = getTargetDir(deployment_dir, hash) + ver_file;
```
mike-sul, line 98: `file` won't be empty even if `getTargetDir` returns empty string, so the `if` at line 100 won't work.
I suppose we need to check return value of the `getTargetDir()` call at first, and then append `ver_file`.

src/bootloader/bootloaderlite.h
```diff
@@ -35,13 +35,16 @@ class BootFwUpdateStatus {
...
   static constexpr const char* const VersionFile{"/usr/lib/firmware/version.txt"};
+  static constexpr const char* const OstreeTargetPathVar{"FIO_OSTREE_TARGET_SYSROOT"};
 
   explicit BootloaderLite(BootloaderConfig config, INvStorage& storage, OSTree::Sysroot::Ptr sysroot);
 
   VersionStrRes getDeploymentVersion(const std::string& hash) const;
 
   static std::string getVersion(const std::string& deployment_dir, const std::string& hash,
                                 const std::string& ver_file = VersionFile);
+  static std::string getTargetDir(const std::string& deployment_dir,
```
mike-sul, line 46: I suppose this method can be moved to the `private` section since it's used only within class methods.

mike-sul:
LGTM, just minor comments.

detsch (commented):
I agree with mike-sul's comments. Specially the one about `std::string file = getTargetDir(deployment_dir, hash) + ver_file;` needs to be adjusted. Other then that, looks fine.

src/bootloader/bootloaderlite.h
igoropaniuk: fixed, thanks

igoropaniuk:
Moved kFioEFI commit to this PR, also addressed comments.

@mike-sul @detsch please take a look

src/bootloader/bootloaderlite.cc
```diff
@@ -93,21 +95,14 @@ BootloaderLite::VersionStrRes BootloaderLite::getDeploymentVersion(const std::st
...
-      if (boost::algorithm::contains(dir, hash)) {
-        file = dir + ver_file;
-        break;
-      }
-    }
-    if (file.empty()) {
-      LOG_WARNING << "Target deployment hash was not found";
+    std::string target_dir = getTargetDir(deployment_dir, hash);
+
+    if (target_dir.empty() || !boost::filesystem::is_directory(target_dir)) {
```
mike-sul, line 102: I think `getTargetDir` guarantees that the return path is a dir if it is not empty. So, I would just check if the `target_dir` is empty and print the Warning message saying that the target deployment dir is not found for `hash` in `deployment_dir` (deployments top level root dir).

igoropaniuk: fixed, thanks

src/bootloader/bootloaderlite.cc
```diff
@@ -43,6 +43,8 @@ void BootloaderLite::installNotify(const Uptane::Target& target) const {
       }
       const std::string& target_version{std::get<0>(target_version_val)};
 
+      env_cmd_vars_ =  boost::str(boost::format("%s=%s") % OstreeTargetPathVar %
```
mike-sul, line 48: I think the `case RollbackMode::kFioEFI` is missing at line #37.
Also, it can be something like:

```
case RollbackMode::kFioEFI:
  env_cmd_vars_ =  boost::str(boost::format("%s=%s") % OstreeTargetPathVar...
case RollbackMode::kUbootMasked:
case RollbackMode::kFioVB: {
  ...
}
```

igoropaniuk: correct, fixed

src/bootloader/bootloaderlite.cc
```diff
@@ -95,21 +98,15 @@ BootloaderLite::VersionStrRes BootloaderLite::getDeploymentVersion(const std::st
...
-        file = dir + ver_file;
-        break;
-      }
-    }
-    if (file.empty()) {
-      LOG_WARNING << "Target deployment hash was not found";
+    std::string target_dir = getTargetDir(deployment_dir, hash);
+
+    if (target_dir.empty()) {
+      LOG_DEBUG << "Target's system root directory is not found for hash: " << hash <<
```
mike-sul, line 104: Please, make it `LOG_WARNING`.

src/bootloader/bootloaderlite.cc
```diff
@@ -93,23 +98,17 @@ BootloaderLite::VersionStrRes BootloaderLite::getDeploymentVersion(const std::st
...
+    if (target_dir.empty()) {
+      LOG_WARNING << "Target's system root directory is not found for hash: " << hash <<
+                   " in: " << deployment_dir;
       return std::string();
     }
+
+    std::string file = target_dir + ver_file;
     if (!boost::filesystem::exists(file)) {
-      LOG_DEBUG << "Target's bootloader version file is not found, expected: " << file;
+      LOG_WARNING << "Target's bootloader version file is not found, expected: " << file;
```
igoropaniuk, line 111: @mike-sul I also changed to warning here

mike-sul (approved):

detsch (approved):
LGTM

mike-sul (approved):

PR #327, opened 2024-06-11
makefile: add code coverage report for unit tests

detsch:
Coverage HTML report can be generated by calling the following Makefile targets: config-coverage build test-coverage-html

gcovr is required, and it was added to the build image Dockerfile.


mike-sul (approved):

mike-sul:
@detsch FYI, I've triggered a build of the  `foundries/aklite-dev` image with our changes, so the image should contain `gcovr` now.

PR #328, opened 2024-06-17
tests: Improve reliability of events verification

detsch:
Add a fsync operation inside the fake device gateway used for testing, and also add a delay before reading the generated events file. The delay is required because the sending of events inside aklite is not done synchronously. A ReportQueue flush is forced on LiteClient destruction, but we keep an instance active during the test procedure.

---

@mike-sul original results looked promising, but now I still get false negatives when running on Github. I'll take a better look tomorrow.

An alternative implementation would be to add some method to force the ReportQueue flush and make sure the execution only proceeds after the reply from the device gateway, but it is probably not worth it.

### DIFF
--- a/tests/device-gateway_fake.py
+++ b/tests/device-gateway_fake.py
@@ -187,6 +187,7 @@ class Handler(SimpleHTTPRequestHandler):
                     cur_events.append(e)
                     event_ids.add(e["id"])
             json.dump(cur_events, f)
+            os.fsync(f)
 
     def _reset_events(self):
         if os.path.exists(self.server.events_file):
--- a/tests/fixtures/liteclient/devicegatewaymock.cc
+++ b/tests/fixtures/liteclient/devicegatewaymock.cc
@@ -50,8 +50,15 @@ class DeviceGatewayMock {
   std::string getTlsUri() const { return url_; }
   const std::string& getPort() const { return port_; }
   Json::Value getReqHeaders() const { return Utils::parseJSONFile(req_headers_file_); }
-  Json::Value getEvents() const { return Utils::parseJSONFile(events_file_); }
+  Json::Value getEvents() const {
+    // Allow some time for the ReportQueue to be flushed and the events to arrive at the Device Gateway mock.
+    // The flush operation is forced in the LiteClient destructor, but we inspect the sent events while keeping an
+    // LiteClient instance active.
+    std::this_thread::sleep_for(std::chrono::milliseconds(500));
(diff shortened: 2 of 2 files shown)


mike-sul:
> An alternative implementation would be to add some method to force the ReportQueue flush and make sure the execution only proceeds after the reply from the device gateway, but it is probably not worth it.

I thought about it too, and I came to the same conclusion - it is not worth it.

mike-sul:
> I still get false negatives when running on Github. I'll take a better look tomorrow.

It is related to the "no space" test. It might be tricky, I can take care of it. It has to do with with different output of the "random" function on different env.

You may wanna play with this

```
   // Enough free space to pull an App bundle/archive and the App image layers/blobs.
    // But, it's not enough available free space to accommodate the App in the docker store (extracted image layers).
    SetFreeBlockNumb(37, 100);
```

and increase `37` to `38`. The problem is that it will fix the problem got the github env but it might cause failure for the local or jobserv env.

mike-sul (approved):

detsch:
> > I still get false negatives when running on Github. I'll take a better look tomorrow.
> 
> It is related to the "no space" test. It might be tricky, I can take care of it. It has to do with with different output of the "random" function on different env.

I believe it is still the same problem. `getEvents` returning an empty Json, which leads to the `store: docker` string not being found in the event (`event_err_msg` is "" in that case): `Value of: std::string::npos != event_err_msg.find("store: docker")`

For some reason, this is the most common location of the false negative.

I'll do some additional testing, probably force pushing some changes here.

mike-sul:
> > > I still get false negatives when running on Github. I'll take a better look tomorrow.
> > 
> > 
> > It is related to the "no space" test. It might be tricky, I can take care of it. It has to do with with different output of the "random" function on different env.
> 
> I believe it is still the same problem. `getEvents` returning an empty Json, which leads to the `store: docker` string not being found in the event (`event_err_msg` is "" in that case): `Value of: std::string::npos != event_err_msg.find("store: docker")`
> 
> For some reason, this is the most common location of the false negative.
> 
> I'll do some additional testing, probably force pushing some changes here.

I think it can be both.
The error message says `failed to fetch App; app: app-01; uri: hub.foundries.io/factory/app-01@sha256:8f611141fba39ed2d6185683f48f31053c4e59ed2346af465c95888df2e0bbe4; Insufficient storage available; store: skopeo;` while it should be "... store: docker". So, even if the test gets a proper event with error message the test will fail because it expects "docker" in the error message.

PR #330, opened 2024-06-19
bootloader: fix format of recently added code

detsch:
Commit: fa9e0e5 (bootloader: add support for env variables, 2024-06-06) introduced a few code formatting errors, now corrected by running `make format`.

---
@igoropaniuk `make format` generates a few changes over your original code. I took the liberty of creating a new commit for it.

### DIFF
--- a/src/bootloader/bootloaderlite.cc
+++ b/src/bootloader/bootloaderlite.cc
@@ -37,8 +37,8 @@ void BootloaderLite::installNotify(const Uptane::Target& target) const {
     case RollbackMode::kUbootGeneric:
       break;
     case RollbackMode::kFioEFI:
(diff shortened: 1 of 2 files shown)


mike-sul (approved):
I wonder why the CI tests didn't detected it.

detsch:
> I wonder why the CI tests didn't detected it.

Not the first time I see some format error slipping in. `make format` does not fail is such cases. It is not like `make check-format` from aktualizr. I could not get yet how `make check-format` works, but I can take a look at that once I have some time.

mike-sul:
> `make format` does not fail is such cases.

`make tidy` is supposed to fail in this case.

I think this CC/CXX env vars can have an effect. Specifically, we use clang and clang++ in the github flow, while you may use gcc/g++. I've seen different behavior for this two.

mike-sul (approved):

PR #331, opened 2024-06-24
main: Exit with error when there is a pending reboot on daemon start

detsch:


src/main.cc
```diff
@@ -117,7 +117,11 @@ int daemon_main_(LiteClient& client, uint64_t interval, bool return_on_sleep) {
 
   std::shared_ptr<LiteClient> client_ptr{&client, [](LiteClient* /*unused*/) {}};
   AkliteClientExt akclient{client_ptr, false, true};
-  akclient.CompleteInstallation();
+  auto finalize_result = akclient.CompleteInstallation();
```
mike-sul, line 120: I would check whether there is pending installation before running `CompleteInstallation`:

```
if (client.IsInstallationInProgress) {
   auto finalize_result = akclient.CompleteInstallation();
  ...
}
```

src/main.cc
```diff
@@ -117,7 +117,11 @@ int daemon_main_(LiteClient& client, uint64_t interval, bool return_on_sleep) {
...
+  if (finalize_result.status == InstallResult::Status::NeedsCompletion) {
+    LOG_ERROR << "A system reboot is required to complete a previous installation";
```
mike-sul, line 122: I would use the terms users have already got used, i.e. "... required to finalize the pending installation."

src/main.cc
detsch: Done.

src/main.cc
detsch: Done.

PR #334, opened 2024-06-27
Extend cli commands

detsch:


detsch:
One limitation of the current implementation is that calling the `install` command does not use the same correlation id of the previous `pull` operation on the same target.

mike-sul:
> One limitation of the current implementation is that calling the `install` command does not use the same correlation id of the previous `pull` operation on the same target.

Yes, I knew about :), let's discuss it offline.

mike-sul:
Apart the thing we discussed offline it LGTM.

src/main.cc
```diff
@@ -67,43 +67,11 @@ static int status_main(LiteClient& client, const bpo::variables_map& unused) {
...
-    }
-    for (const auto& it : t.hardwareIds()) {
-      if (it == hwid) {
-        sorted_targets.emplace(ver, t);
-        break;
-      }
-    }
-  }
+  std::shared_ptr<LiteClient> client_ptr{&client, [](LiteClient* /*unused*/) {}};
+  AkliteClient akclient(client_ptr, false, true);
```
mike-sul, line 71: Do we really want/need to finalize the pending update if a user wants just list available targets?

src/main.cc
```diff
@@ -67,43 +67,11 @@ static int status_main(LiteClient& client, const bpo::variables_map& unused) {
...
 
-  LOG_INFO << "Updates available to " << hwid << ":";
-  for (auto& pair : sorted_targets) {
-    client.logTarget("", pair.second);
-  }
-  return 0;
+  auto status = aklite::cli::CheckIn(akclient, nullptr);
```
mike-sul, line 73: I wonder why we didn't declare the second parameter of the `cli::CheckIn` with the default `nullptr` value.

src/main.cc
```diff
@@ -85,20 +85,26 @@ static int daemon_main(LiteClient& client, const bpo::variables_map& variables_m
...
-  int version = -1;
+static void get_target_id(const bpo::variables_map& params, int& version, std::string& name) {
+  name = "";
+  version = -1;
   if (params.count("update-name") > 0) {
     const auto version_str = params["update-name"].as<std::string>();
     try {
       version = std::stoi(version_str);
     } catch (const std::invalid_argument& exc) {
       LOG_INFO << "Failed to convert the input target version to integer, consider it as a name of Target: "
```
mike-sul, line 96: I don't know why I made the log as "info" log in the first place. Now, it looks like it makes sense to turn it into a debug or even trace error.

mike-sul (approved):

src/main.cc
detsch: It caught my attention too. Specially because the parameter has "name" in it. I've added a new commit changing the log to LOG_DEBUG.

src/main.cc
detsch: I'll consider this change in a next PR. It makes sense to have the default `nullptr` there

src/main.cc
mike-sul: Thanks.

PR #336, opened 2024-07-02
api, cli: Do not perform TUF checkin on every Pull or Install 

detsch:
@mike-sul please take a look at the overall set of changes. I will probably still improve the code, and also, it was not tested yet.

Main concerns:
- Make sure we do want to change the cli interface at that level. See how in the unit tests we now need to call `cli::CheckIn` explicitly most of the times before we call `cli::Install`;

- The `check-for-update-post` success callback is a bit of a problem. The original daemon behavior is to call it at the end of `GetTargetsToInstall` (like it was before the recent changes), and IMO we should keep that same behavior where possible. However, if we call the `CheckIn` method exclusively, `check-for-update-post` success should be called at the end of `CheckIn`, and, when calling the standalone pull or install operations, `GetTargetToInstall` should never call `check-for-update-post` (not even on errors). The expected behavior is defined by setting a new boolean parameter in the `AkLiteClientExt` constructor.

### DIFF
--- a/apps/aklite-offline/cmds.cpp
+++ b/apps/aklite-offline/cmds.cpp
@@ -19,7 +19,7 @@ int CheckCmd::checkSrcDir(const po::variables_map& vm, const boost::filesystem::
 
 int InstallCmd::installUpdate(const po::variables_map& vm, const boost::filesystem::path& src_dir,
                               const std::string& target_name, bool force_downgrade) const {
-  AkliteClientExt client(vm, false, false);
+  AkliteClientExt client(vm, false, false, false);
   const LocalUpdateSource local_update_source{.tuf_repo = (src_dir / "tuf").string(),
                                               .ostree_repo = (src_dir / "ostree_repo").string(),
                                               .app_store = (src_dir / "apps").string()};
--- a/include/aktualizr-lite/cli/cli.h
+++ b/include/aktualizr-lite/cli/cli.h
@@ -55,14 +55,30 @@ enum class PullMode {
   None,
 };
 
+/**
+ * The TUF check mode to be used.
+ */
+enum class CheckMode {
+  /**
+   * Default mode, do update TUF metadata from remote of local source.
+   */
+  Update = 0,
+  /**
+   * Do not update TUF roles: use currently stored metadata.
+   */
+  Current,
+};
+
 StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_source = nullptr);
 
 StatusCode Pull(AkliteClientExt &client, int version = -1, const std::string &target_name = "",
-                bool force_downgrade = true, const LocalUpdateSource *local_update_source = nullptr);
+                bool force_downgrade = true, const LocalUpdateSource *local_update_source = nullptr,
+                CheckMode check_mode = CheckMode::Update);
 
 StatusCode Install(AkliteClientExt &client, int version = -1, const std::string &target_name = "",
                    InstallMode install_mode = InstallMode::All, bool force_downgrade = true,
-                   const LocalUpdateSource *local_update_source = nullptr, PullMode pull_mode = PullMode::All);
+                   const LocalUpdateSource *local_update_source = nullptr, PullMode pull_mode = PullMode::All,
+                   CheckMode check_mode = CheckMode::Update);
 
 StatusCode CompleteInstall(AkliteClient &client);
 
--- a/include/aktualizr-lite/cli/cli.h
+++ b/include/aktualizr-lite/cli/cli.h
@@ -55,7 +55,7 @@ enum class PullMode {
   None,
 };
 
-StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_source);
+StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_source = nullptr);
 
 StatusCode Pull(AkliteClientExt &client, int version = -1, const std::string &target_name = "",
                 bool force_downgrade = true, const LocalUpdateSource *local_update_source = nullptr);
--- a/src/aklite_client_ext.cc
+++ b/src/aklite_client_ext.cc
@@ -11,6 +11,7 @@
 #include "http/httpclient.h"
 #include "libaktualizr/config.h"
 #include "liteclient.h"
+#include "logging/logging.h"
 #include "primary/reportqueue.h"
 
 #include "aktualizr-lite/tuf/tuf.h"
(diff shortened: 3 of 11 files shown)


apps/aklite-offline/cmds.cpp
```diff
@@ -19,10 +19,15 @@ int CheckCmd::checkSrcDir(const po::variables_map& vm, const boost::filesystem::
 
 int InstallCmd::installUpdate(const po::variables_map& vm, const boost::filesystem::path& src_dir,
                               const std::string& target_name, bool force_downgrade) const {
-  AkliteClientExt client(vm, false, false);
+  AkliteClientExt client(vm, false, false, false);
```
mike-sul, line 22: Not related to this PR. I think we should replace these boolean params with bitwise flags defined as an enum values, hence only one parameter is passed to the ctor (enum) and a caller defines required params through bitwise AND, e.g. (READ_ONLY | CHECK_POST_CB).

src/aklite_client_ext.cc
```diff
@@ -26,25 +27,25 @@
...
 // apps
-GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const LocalUpdateSource* local_update_source, int version,
+GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult& check_in_result, int version,
                                                              const std::string& target_name, bool allow_bad_target,
-                                                             bool force_apps_sync) {
-  // Tell CheckIn and CheckInLocal methods to not call notifyTufUpdateFinished in case of success
-  usingUpdateClientApi = true;
+                                                             bool force_apps_sync, bool is_offline_mode) {
+  std::string err;
+  if (!check_in_result) {
```
mike-sul, line 34: It looks like redundant check since it will fail at line 44-48 anyway. Also, it seems like ` client_->notifyTufUpdateFinished(err);` won't be invoked in this case, while we want  to invoke this callback even if CheckIn fails.

src/aklite_client_ext.cc
```diff
@@ -26,25 +27,25 @@
 // Returns the target that should be installed, if any.
 // It might be an updated version, a rollback target, or even the currently installed target, in case we need to sync
 // apps
-GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const LocalUpdateSource* local_update_source, int version,
+GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult& check_in_result, int version,
```
mike-sul, line 30: I would call this parameter as `checkin_result` or `checkin_res` since this is a result of the `CheckIn` call, we kind of use "check" and "in" together.

src/api.cc
```diff
@@ -573,12 +573,61 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
...
+  try {
+    tuf_repo_->CheckMeta();
+    LOG_INFO << "The stored TUF metadata is valid";
+  } catch (const std::exception& exc) {
+    err_msg = std::string("Stored TUF metadata is invalid: ") + exc.what();
+    LOG_WARNING << err_msg;
+    return CheckInResult{CheckInResult::Status::SecurityError, hw_id_, {}};
+  }
+
+  LOG_INFO << "Searching for matching TUF Targets...";
```
mike-sul, line 594: Optional, not necessary at this stage/PR. The code after CheckMeta() seems similar or even identical to the one we have in CheckIn. So, we may wanna consider unifying it, e.g. moving to the common/shared private method, etc.

src/cli/cli.cc
```diff
@@ -121,11 +121,12 @@ static StatusCode pullAndInstall(AkliteClientExt &client, int version, const std
     return StatusCode::InstallationInProgress;
   }
 
-  auto cr = client.GetTargetToInstall(local_update_source, version, target_name, true, true);
+  auto csm_res = client.CheckStoredMetadata(local_update_source);
```
mike-sul, line 124: `const auto` ?

src/api.cc
detsch: I agree. I was going to do it right at the beginning, but the amount of changes required in the CheckIn / CheckInLocal methods made me postpone it. Perhaps I'll do it as a separate patch.

src/api.cc
```diff
@@ -346,19 +346,18 @@ static std::vector<Uptane::Target> getAvailableTargets(const PackageConfig& pcon
...
-      LOG_INFO << "\tmissing ostree commit: " << t.sha256Hash();
+      LOG_DEBUG << "\t" << t.filename() << " - missing ostree commit: " << t.sha256Hash();
       continue;
     }
     if (pconfig.type != ComposeAppManager::Name) {
       auto custom_data{t.custom_data()};
       custom_data[LocalSrcDirKey]["ostree"] = src.OstreeRepoDir.string();
       found_targets.emplace_back(Target::updateCustom(t, custom_data));
-      LOG_INFO << "\tall target content have been found";
+      LOG_INFO << "\t" << t.filename() << " - all target content has been found";
```
mike-sul, line 360: I probably should use something like `target elements` or `target components` instead of `content` since English grammar consider it as a singular noun :).

src/aklite_client_ext.cc
```diff
@@ -26,25 +27,28 @@
...
-    return GetTargetToInstallResult(checkin_res);
+    err = "Can't select target to install using a failed check-in result";
+    LOG_WARNING << err << " " << static_cast<int>(checkin_res.status);
+    if (!invoke_post_cb_at_checkin_) {
+      client_->notifyTufUpdateFinished(err);
+    }
+    return {GetTargetToInstallResult::Status::Failed, TufTarget(), err};
   }
 
+  client_->setAppsNotChecked();
```
mike-sul, line 43: My understanding is that `setAppsNotChecked` is not needed if the condition at line 34 is true (negative checkin result)?

src/api.cc
mike-sul: No worries, it is not critical at all.

include/aktualizr-lite/cli/cli.h
```diff
@@ -55,14 +55,30 @@ enum class PullMode {
...
+   */
+  Current,
+};
+
 StatusCode CheckIn(AkliteClient &client, const LocalUpdateSource *local_update_source = nullptr);
 
 StatusCode Pull(AkliteClientExt &client, int version = -1, const std::string &target_name = "",
-                bool force_downgrade = true, const LocalUpdateSource *local_update_source = nullptr);
+                bool force_downgrade = true, const LocalUpdateSource *local_update_source = nullptr,
+                CheckMode check_mode = CheckMode::Update);
```
mike-sul, line 76: Maybe it makes sense to set it to `CheckMode::Current` by default?

include/aktualizr-lite/cli/cli.h
```diff
@@ -55,14 +55,30 @@ enum class PullMode {
...
 
 StatusCode Install(AkliteClientExt &client, int version = -1, const std::string &target_name = "",
                    InstallMode install_mode = InstallMode::All, bool force_downgrade = true,
-                   const LocalUpdateSource *local_update_source = nullptr, PullMode pull_mode = PullMode::All);
+                   const LocalUpdateSource *local_update_source = nullptr, PullMode pull_mode = PullMode::All,
+                   CheckMode check_mode = CheckMode::Update);
```
mike-sul, line 81: If it is set to `CheckMode::Update` by default then I think `cli_install` and `install` functions in main.cc require adjustment so `CheckMode::Current` mode is applied in the `cli_install` case?

mike-sul: It's been addressed.

src/aklite_client_ext.cc
mike-sul: @detsch Have you checked this? It looks like `setAppsNotChecked` should be invoked if if (!checkin_res) {` is true too?

src/api.cc
```diff
@@ -357,7 +357,7 @@ static std::vector<Uptane::Target> getAvailableTargets(const PackageConfig& pcon
       auto custom_data{t.custom_data()};
       custom_data[LocalSrcDirKey]["ostree"] = src.OstreeRepoDir.string();
       found_targets.emplace_back(Target::updateCustom(t, custom_data));
-      LOG_INFO << "\t" << t.filename() << " - all target content has been found";
+      LOG_INFO << "\t" << t.filename() << " - all target components have been found";
```
mike-sul, line 360: Thanks for using "components".

mike-sul (approved):

PR #337, opened 2024-07-10
examples: Remove custom sota client example

mike-sul:
Since the custom sota client is moved to [the dedicated git repo](https://github.com/foundriesio/sotactl) the example code is removed from the aklite repo/project.


detsch:
There is one `custom-client` reference left, that should be removed, inside the Makefile:
https://github.com/foundriesio/aktualizr-lite/blob/b716ae380fd9a8d64ba1651bef01efb313d1331d/Makefile#L13

mike-sul:
> There is one `custom-client` reference left, that should be removed, inside the Makefile:
> 
> https://github.com/foundriesio/aktualizr-lite/blob/b716ae380fd9a8d64ba1651bef01efb313d1331d/Makefile#L13

Fixed in the forced push, thanks.

detsch (approved):
LGTM

PR #338, opened 2024-07-15
appengine: Remove app only from runtime part

mike-sul:
Remove app components only from the runtime part, i.e., the Docker store and Compose project in the appengine->remove() call, and leave their content in the OCI/blob store, as it may be needed during rollback or if a given app is listed in reset_apps. After an update completion, unused blobs are removed from the blob stores in the prune call, ensuring proper app removal if it is not listed in reset_apps.


detsch (approved):
LGTM

PR #339, opened 2024-07-15
Retain all target apps

mike-sul:
Added the ability to persist/retain all Target apps regardless of the `compose_apps` value by introducing the pseudo value "*" for `reset_apps`.


detsch (approved):
LGTM

mike-scott:
Thanks @mike-sul LGTM

PR #340, opened 2024-07-16
appengine: Remove app only from runtime part

mike-sul:
Remove app components only from the runtime part, i.e. docker store and compose project in the appengine->remove() call, and leave its content in the OCI/blob store as it can be needed during rollback or if a given app is listed in the `reset_apps`.
After successful update unused blobs are removed from the blob stores, which guarantees proper app removal if it is not listed in `reset_apps`.


mike-sul:
@detsch This is the same commit we merged into the v94 branch yesterday.

detsch (approved):
LGTM

PR #344, opened 2024-07-26
tests: Replace LiteCLient with LiteClientMock, and add improve CheckIn tests

detsch:
@mike-sul I'm ok with dropping the changes if you don't think it is worth to replace `LiteCLient` with `LiteClientMock` globally in the tests. But I believe it will bring some additional tests possibilities that can justify the change.

### DIFF
--- a/tests/apiclient_test.cc
+++ b/tests/apiclient_test.cc
@@ -562,6 +562,30 @@ TEST_F(ApiClientTest, InstallTargetWithHackedApps) {
   ASSERT_EQ(nullptr, installer);
 }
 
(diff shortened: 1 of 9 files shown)


mike-sul (approved):

mike-sul:
> @mike-sul I'm ok with dropping the changes if you don't think it is worth to replace `LiteCLient` with `LiteClientMock` globally in the tests. But I believe it will bring some additional tests possibilities that can justify the change.

It is fine with me, just the name `LiteClientMock` looks a bit odd taking into account that it inherits the real `LiteClient`.

PR #346, opened 2024-07-31
appengine: Add required header to use std::find

mike-sul:
It turned out that g++ v14.1.1 requires explicit inclusion of the `algorithm` header to use `std::find`.


detsch (approved):
LGTM

PR #347, opened 2024-08-01
cli: Add function returning string description of StatusCode values 

detsch:
Single commit PR, on top of PR #345

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -72,6 +72,71 @@ static const std::unordered_map<InstallResult::Status, StatusCode> i2s = {
     {InstallResult::Status::UnknownError, SC::UnknownError},
 };
 
+using SC = StatusCode;
+static const std::unordered_map<SC, std::string> status2string = {
+    {SC::Ok, "Operation succeeded"},
+
+    // Return codes for CheckIn, Pull and Install
+    {SC::CheckinOkCached, "Unable to fetch updated TUF metadata, but stored metadata is valid"},
+    {SC::CheckinFailure, "Failed to update TUF metadata"},
+    {SC::CheckinNoMatchingTargets,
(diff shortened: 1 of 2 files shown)


src/cli/cli.cc
```diff
@@ -70,6 +70,66 @@ static const std::unordered_map<InstallResult::Status, StatusCode> i2s = {
     {InstallResult::Status::UnknownError, StatusCode::UnknownError},
 };
 
+using SC = StatusCode;
```
mike-sul, line 73: I would rather use this alias for the overall file (define on top and apply to all cli.cpp) or drop it and just use `StatusCode` as it is defined everywhere.

src/cli/cli.cc
```diff
@@ -70,6 +70,66 @@ static const std::unordered_map<InstallResult::Status, StatusCode> i2s = {
...
+
+    // Return codes for CompleteInstall
+    {SC::NoPendingInstallation, "No pending installation to run"},
+    {SC::InstallOfflineRollbackOk, "Offline installation failed, rollback performed"},
+    {SC::InstallRollbackOk, "Online installation failed, rollback performed"},
+    {SC::InstallRollbackFailed, "Installation failed and rollback operation was not successful"},
+    {SC::UnknownError, "Unknown error"},
+};
+
+std::string StatusCodeDescription(StatusCode status) { return status2string.at(status); }
```
mike-sul, line 131: I think we need to check whether the key (`status`) exists in the map, otherwise the `.at` call will issue an exception. Please, check the code at line 15, I think it is worth doing similar here.

mike-sul (approved):
LGTM, just minor comments.

src/cli/cli.cc
detsch: Done.

src/cli/cli.cc
detsch: Done

PR #350, opened 2024-08-12
daemon: Do not skip sleep operation after successful install

detsch:
Avoid making a situation where we repeatedly apply an update worse by not sleeping between install operations.


src/daemon.cc
```diff
@@ -48,8 +48,9 @@ int run_daemon(LiteClient& client, uint64_t interval, bool return_on_sleep, bool
           // If a reboot command is set in configuration, and is executed successfully, we will not get to this point
           break;
         }
-        if (install_result) {
-          // After a successful install, do not sleep, go directly to the next iteration
+        if (install_result && gti_res.status != GetTargetToInstallResult::Status::UpdateSyncApps) {
```
mike-sul, line 51: I think we don't the `if` condition at all now because we improved both aklite and the DG so there is no need to do checkin just after installation to update device's target name on the server.
Let's double check my statement and remove the `if` at all if I am correct.

mike-sul: I did a quick test and confirm that a device state is properly updated on the backend and displayed on the webui without immediate "checkin" after successful update.

Therefore, I think we should remove this "if" statement at all.

detsch: OK, that is even better since it keeps the loop simpler, and avoids an infinite loop without sleep in case of some false install success.

The first version of the PR was actually removing the if, but then I went back to our current source and observed that the current behavior (v94) is to skip the sleep.

mike-sul (approved):

PR #351, opened 2024-08-13
cli: Add json output mode for 'check' and 'list' commands

detsch:
Implementation was straightforward, so I just made a first version while we get seek for some feedback

### DIFF
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -1,11 +1,14 @@
 #include "aktualizr-lite/cli/cli.h"
 
+#include <algorithm>
 #include <iostream>
 #include <unordered_map>
 
 #include "aktualizr-lite/api.h"
 
 #include "aktualizr-lite/aklite_client_ext.h"
+#include "json/reader.h"
+#include "json/value.h"
 #include "logging/logging.h"
 
 namespace aklite::cli {
@@ -157,9 +160,59 @@ static CheckInResult checkIn(AkliteClientExt &client, CheckMode check_mode,
   }
 }
 
-StatusCode CheckIn(AkliteClientExt &client, const LocalUpdateSource *local_update_source, CheckMode check_mode) {
+static Json::Value getCheckInResultJson(AkliteClientExt &client, const CheckInResult &ci_res,
+                                        const GetTargetToInstallResult &gti_res) {
+  const auto current = client.GetCurrent();
+
+  Json::Value json_root;
+
+  auto app_shortlist = client.GetAppShortlist();
+  for (const auto &target : ci_res.Targets()) {
+    Json::Value json_target;
+    json_target["name"] = target.Name();
+    json_target["version"] = target.Version();
+    if (!gti_res.selected_target.IsUnknown() && gti_res.selected_target.Name() == target.Name()) {
+      json_target["selected"] = true;
+      json_target["reason"] = gti_res.reason;
+    }
+    if (client.IsRollback(target)) {
+      json_target["failed"] = true;
+    }
+    if (current.Version() == target.Version()) {
+      json_target["current"] = true;
+    } else if (current.Version() < target.Version()) {
+      json_target["newer"] = true;
+    }
+
+    Json::Value json_apps;
+    for (const auto &app : TufTarget::Apps(target)) {
+      Json::Value json_app;
+      json_app["name"] = app.name;
+      json_app["uri"] = app.uri;
+      json_app["on"] =
+          (!app_shortlist || std::find(app_shortlist->begin(), app_shortlist->end(), app.name) != app_shortlist->end());
+      json_apps.append(json_app);
+    }
+    json_target["apps"] = json_apps;
+
+    json_root.append(json_target);
+  }
+
+  return json_root;
+}
+
+StatusCode CheckIn(AkliteClientExt &client, const LocalUpdateSource *local_update_source, CheckMode check_mode,
+                   bool json_output) {
   auto cr = checkIn(client, check_mode, local_update_source);
-  if (cr) {
+  if (!cr) {
+    return res2StatusCode<CheckInResult::Status>(c2s, cr.status);
+  }
+
+  auto gti_res = client.GetTargetToInstall(cr);
+
+  if (json_output) {
+    std::cout << getCheckInResultJson(client, cr, gti_res) << "\n";
+  } else {
     if (cr.Targets().empty()) {
       std::cout << "\nNo Targets found"
                 << "\n";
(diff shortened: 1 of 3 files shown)


src/cli/cli.cc
```diff
@@ -157,9 +160,59 @@ static CheckInResult checkIn(AkliteClientExt &client, CheckMode check_mode,
...
+  auto app_shortlist = client.GetAppShortlist();
+  for (const auto &target : ci_res.Targets()) {
+    Json::Value json_target;
+    json_target["name"] = target.Name();
+    json_target["version"] = target.Version();
+    if (!gti_res.selected_target.IsUnknown() && gti_res.selected_target.Name() == target.Name()) {
+      json_target["selected"] = true;
+      json_target["reason"] = gti_res.reason;
+    }
+    if (client.IsRollback(target)) {
```
mike-sul, line 178: This is not quite efficient performance wise since each `.IsRollback()` call does exactly the same action - get installed versions from the DB. I wonder if it is possible to optimize it somehow since a user device can have thousands targets?

src/cli/cli.cc
```diff
@@ -157,9 +160,59 @@ static CheckInResult checkIn(AkliteClientExt &client, CheckMode check_mode,
...
+      json_target["bad"] = true;
```
mike-sul, line 179: Maybe we should call it `rollback`, so it looks like `"rollback": true`, which is more clear way to indicate that a rollback occurred when trying to install a given target? Just IMO.

src/cli/cli.cc
```diff
@@ -157,9 +160,59 @@ static CheckInResult checkIn(AkliteClientExt &client, CheckMode check_mode,
...
+
+StatusCode CheckIn(AkliteClientExt &client, const LocalUpdateSource *local_update_source, CheckMode check_mode,
+                   bool json_output) {
   auto cr = checkIn(client, check_mode, local_update_source);
-  if (cr) {
+  if (!cr) {
+    return res2StatusCode<CheckInResult::Status>(c2s, cr.status);
+  }
+
+  auto gti_res = client.GetTargetToInstall(cr);
```
mike-sul, line 211: I think we need to do this call only for the json output. If so, then it could be used only within "if true" branch.

src/cli/cli.cc
```diff
@@ -184,11 +237,7 @@ StatusCode CheckIn(AkliteClientExt &client, const LocalUpdateSource *local_updat
       std::cout << "\n";
     }
   }
-  if (!cr) {
-    return res2StatusCode<CheckInResult::Status>(c2s, cr.status);
-  }
 
-  auto gti_res = client.GetTargetToInstall(cr);
   if (gti_res && gti_res.selected_target.IsUnknown()) {
```
mike-sul, line 241: It looks like this is needed only for the json type of output?

src/main.cc
```diff
@@ -67,22 +67,25 @@ static int status_main(LiteClient& client, const bpo::variables_map& unused) {
...
-  auto status = aklite::cli::CheckIn(akclient, nullptr, check_mode);
+  auto status = aklite::cli::CheckIn(akclient, nullptr, check_mode, json_output);
   return static_cast<int>(status);
 }
 
-static int cli_list(LiteClient& client, const bpo::variables_map& unused) {
-  (void)unused;
-  return checkin(client, aklite::cli::CheckMode::Current);
+static int cli_list(LiteClient& client, const bpo::variables_map& params) {
+  return checkin(client, aklite::cli::CheckMode::Current, params);
```
mike-sul, line 84: I think we need to parameterize the `checkin` to indicate whether to do `GetTargetToInstall` or not because the `list` command should just list available targets. There is no need in `GetTargetToInstall` to determine whether a given target is "bad" or "current", only to determine whether it is selected or not. The latter is needed only for the `check` command.

src/cli/cli.cc
detsch: It is not related to the json output. This was added here: https://github.com/foundriesio/aktualizr-lite/pull/345/commits/487ef43eead80f51855691aae59705192431659f with the objective of allowing the CLI check command result to be used directly to decide if a pull/install is expected to be called. I.e., if there is a target that need to be installed.

src/cli/cli.cc
detsch: I find the "rollback" term a bit ambiguous. I even changed some variable names because of that during the recent refactoring. "rollback" may cause confusion as there are 2 targets involved during a rollback operation. "caused_rollback" could be more appropriate to indicate the bad one. "bad" should be clear enough, but I'm open to change it.

src/cli/cli.cc
detsch: We could add a method to `LiteClient` returning a `set` with the names of all targets that caused a rollback. It would be just a  matter of using functionality from `get_known_but_not_installed_versions` and a little bit from `known_local_target`. What do you think?

mike-sul: I thought about avoiding this iteration over targets here at all. Instead, we can extend the `filterTargets` functionality in the `AkliteClient::CheckIn()` to add the "bad" mark (another attribute of TufTarget object, or additional field in the custom json) (1).

The other two marks "current" and "selected" seems reasonable to set directly in the json output `getCheckInResultJson`. However, if we get more than one "client" for it, we may wanna add another two attributes to TufTarget in `AkliteClientExt::GetTargetToInstall`, but it means the input checkin result modification...(2)

I think it is more or less reasonable to do the first item now, unless it requires too much effort.

src/cli/cli.cc
mike-sul: Maybe "failing" then, or just "failed".

src/cli/cli.cc
mike-sul: OK. Do we need it for the `list` too?

detsch: I would keep it for the `list` operation as well. There is no change of state, and the adequate return code might be useful if someone is trying to use the currently stored metadata to decide on doing or not an update.

src/cli/cli.cc
detsch: `failed` sounds good.

src/cli/cli.cc
mike-sul: ok.

src/cli/cli.cc
detsch: I'm not sure about that, will give it a little more thought. Adding an attribute to `TufTarget`, or event to the custom json, would create an expectation on that attribute to always be correctly set, which has its price, while not being used that much.

mike-sul: OK, let's not think about performance optimization at this stage.

src/cli/cli.cc
detsch: Changed to `failed`

src/main.cc
detsch: As discussed offline, I would rather keep calling `GetTargetToInstall` even for the `list` command, to leave the interface more consistent, and to allow the related functionality to be used even when not updating the TUF metadata.

detsch:
@mike-sul I've added two patches to better handle the log messages and `--json` option: Fixed the LOG_WARNING for non-root execution, and made loglevel "fatal" by default when using "--json 1".

src/cli/cli.cc
detsch: OK. This is something we can handle in a separate PR, as the potential performance hit is only related to the new feature.

mike-sul:
> @mike-sul I've added two patches to better handle the log messages and `--json` option: Fixed the LOG_WARNING for non-root execution, and made loglevel "fatal" by default when using "--json 1".

It works better now, thanks.

mike-sul (approved):

PR #352, opened 2024-08-16
main: Add offline support for check, pull, install and update commands

detsch:


mike-sul (approved):
LGTM, will do e2e testing tomorrow.

src/main.cc
```diff
@@ -125,6 +143,15 @@ static int install(LiteClient& client, const bpo::variables_map& params, aklite:
     install_mode = params.at("install-mode").as<std::string>();
   }
 
+  LocalUpdateSource local_update_source;
```
mike-sul, line 146: It seems like there is missing change for the pull command. As result the pull and install doesn't work (the last one is supposed to work but it depends on pull which doesn't work).

detsch: You are right. I've fixed it now.

mike-sul:
@detsch We need to make clang tidy happy https://github.com/foundriesio/aktualizr-lite/actions/runs/10565675394/job/29270730335?pr=352

PR #354, opened 2024-08-26
Revert `composectl` usage for checking whether app is installed

mike-sul:
This reverts commit 37c0249d94ae1d399e170549acdb33bb5194a704.

The given functionality does not work if images hosted in docker.io are specified in a docker compose project file in a full format. For example, "docker.io/library/nginx:stable-alpine" or "docker.io/homeassistant/home-assistant:stable".

The dockerd removes the "docker.io" or "docker.io/library" prefix when stores such images in its store and as result returns the image URIs in the shorten format in a response to "http://localhost/images/json?all=1" request. For example, "nginx@sha256:<hash>" or "homeassistant/home-assistant@sha256:<hash>".

Therefore, it is not possible to map images listed in the compose project to images reported by dockerd unless the dockerd's URI shortening rules are taken in the "isAppInstalled" logic.

### DIFF
--- a/src/composeapp/appengine.cc
+++ b/src/composeapp/appengine.cc
@@ -131,30 +131,6 @@ bool AppEngine::isAppFetched(const App& app) const {
   return res;
 }
 
-bool AppEngine::isAppInstalled(const App& app) const {
-  bool res{false};
-  try {
-    std::future<std::string> output;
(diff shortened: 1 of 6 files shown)


detsch (commented):
Revert looks good.

PR #355, opened 2024-08-26
Improve API code organization

detsch:
Add some local changes in order to better organize the api.cc code. Key objective is to simplify CheckIn* methods, moving code to separate methods when appropriate, and also reducing code duplication. There is still work to be done in that direction, but the current set of changes already represent an improvement.

---

@mike-sul I still need to give it some additional testing, but it should be ready for review.

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -506,6 +506,28 @@ static Json::Value getBundleMeta(const std::shared_ptr<LiteClient>& client_,
   return bundle_meta;
 }
 
+static std::vector<TufTarget> getTrustedTargets(const std::shared_ptr<aklite::tuf::Repo>& tuf_repo_, Json::Value bundle_meta) {
+  auto trusted_targets{tuf_repo_->GetTargets()};
+  if (!bundle_meta.empty()) {
+    LOG_INFO << "Getting and checking the bundle metadata...";
+    std::vector<TufTarget> bundle_targets{getTrustedBundleTargets(trusted_targets, bundle_meta)};
+    if (bundle_targets.empty()) {
+      return {};
+    }
+
+    if (bundle_targets.size() == bundle_meta["signed"]["x-fio-offline-bundle"]["targets"].size()) {
+      LOG_INFO << "Any of the bundle targets is allowed to be installed";
+    } else {
+      LOG_INFO << "The following bundle targets are allowed to be installed: ";
+      for (const auto& t : bundle_targets) {
+        LOG_INFO << "\t" << t.Name();
+      }
+    }
+    trusted_targets = bundle_targets;
+  }
+  return trusted_targets;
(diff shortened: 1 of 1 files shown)


mike-sul:
This change requires lots of testing, especially for the offline update.

include/aktualizr-lite/api.h
```diff
@@ -371,6 +371,10 @@ class AkliteClient {
 
  private:
   void Init(Config &config, bool finalize = true, bool apply_lock = true);
+  CheckInResult checkInFailure(CheckInResult::Status check_status, std::string err_msg) const;
```
mike-sul, line 374: I would try and make these utility functions as a static non-class methods, if possible, and defined in api.cc, so the header is not changed and the api is not polluted with the helper functions.

src/api.cc
```diff
@@ -443,90 +483,56 @@ std::vector<TufTarget> toTufTargets(const std::vector<Uptane::Target>& targets)
   return ret;
 }
 
+// Throws BundleMetaError
+Json::Value AkliteClient::getBundleMeta(const LocalUpdateSource* local_update_source) const {
```
mike-sul, line 477: For example, if we pass the `client_` and `tuf_repo_` as input params then this can become non-class static method.

include/aktualizr-lite/api.h
detsch: Changed all new private methods to static functions. `api.h` has no changes in the current PR version.

src/api.cc

src/api.cc
```diff
@@ -547,8 +559,7 @@ CheckInResult AkliteClient::CheckInLocal(const LocalUpdateSource* local_update_s
     err_msg =
         "Couldn't find Targets matching the device's hardware ID; check a tag or a hardware ID of the device and the "
         "bundle's tag";
-    LOG_ERROR << err_msg;
-    return CheckInResult{CheckInResult::Status::NoMatchingTargets, "", {}};
+    return checkInFailure(client_, hw_id_, CheckInResult::Status::NoMatchingTargets, err_msg);
```
mike-sul, line 561: The previous version didn't invoke `notifyTufUpdateFinished()` while `checkInFailure` does it unconditionally.

mike-sul (approved):

src/api.cc
detsch: It was missing before. We should call `notifyTufUpdateFinished()` at this point.

mike-sul: Ah, ok.

mike-sul (approved):
LGTM. Let's do extensive e2e testing and merge it then.

detsch:
Did additional e2e testing, no problems found. Merging.

PR #356, opened 2024-08-28
liteclient: Truncate log of update complete event

mike-sul:
Truncate a description of the update complete event if the completion is not successful and it exceeds the maximum allowed size (2048). The truncation is done by erasing the description middle part, so a user can see the beginning of the app starting log and the end of it which is supposed to contain a reason of failure.


detsch (approved):
LGTM

PR #357, opened 2024-08-28
Report which apps require update in DownloadStarted event

detsch:
This is what the DonwloadStarted event details look like for an "apps sync" update operation:
```
Syncing Active Target Apps
- shellhttpd2: missing installation, to be re-installed
```
*(updated the above example after the latest force push)*

I've started by only adding the app name, but decided to go a little further by adding the specific reason as well, so I created a new patch on top of the original implementation.

### DIFF
--- a/src/aklite_client_ext.cc
+++ b/src/aklite_client_ext.cc
@@ -120,7 +120,8 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
                << " Skipping its installation.";
     }
 
-    if (force_apps_sync || !client_->appsInSync(Target::fromTufTarget(current))) {
+    auto apps_to_update = client_->appsToUpdate(Target::fromTufTarget(current));
+    if (force_apps_sync || !apps_to_update.empty()) {
       // Force installation of apps
       res.selected_target = current;
       LOG_INFO
@@ -128,7 +129,10 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
           << res.selected_target.Name();
 
       res.status = GetTargetToInstallResult::Status::UpdateSyncApps;
-      res.reason = "Syncing Active Target Apps";
+      res.reason = "Syncing Active Target Apps\n";
+      for (const auto& app_to_update : apps_to_update) {
+        res.reason += "- " + app_to_update.first + ": " + app_to_update.second + "\n";
+      }
     } else {
       // No targets to install
       res.selected_target = TufTarget();
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -193,7 +193,8 @@ ComposeAppManager::AppsContainer ComposeAppManager::getApps(const Uptane::Target
   return apps;
 }
 
-ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane::Target& t) const {
+ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane::Target& t,
+                                                                    AppsSyncReason& apps_and_reasons) const {
   AppsContainer apps_to_update;
 
   auto currently_installed_target_apps = Target::appsJson(OstreeManager::getCurrent());
@@ -206,6 +207,7 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
     if (app_data.empty()) {
       // new app in Target
       apps_to_update.insert(app_pair);
+      apps_and_reasons[app_pair.first] = "new app in target";
       LOG_INFO << app_name << " will be installed";
       continue;
     }
@@ -213,6 +215,7 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
     if (app_pair.second != app_data["uri"].asString()) {
       // an existing App update
       apps_to_update.insert(app_pair);
+      apps_and_reasons[app_pair.first] = "new version in target";
       LOG_INFO << app_name << " will be updated";
       continue;
     }
@@ -221,15 +224,24 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
         !boost::filesystem::exists(cfg_.apps_root / app_name / Docker::ComposeAppEngine::ComposeFile)) {
       // an App that is supposed to be installed has been removed somehow, let's install it again
       apps_to_update.insert(app_pair);
+      apps_and_reasons[app_pair.first] = "missing installation, to be re-installed";
       LOG_INFO << app_name << " will be re-installed";
       continue;
     }
 
     LOG_DEBUG << app_name << " performing full status check";
-    if (!app_engine_->isFetched({app_name, app_pair.second}) || !app_engine_->isRunning({app_name, app_pair.second})) {
-      // an App that is supposed to be installed and running is not fully installed or running
+    if (!app_engine_->isFetched({app_name, app_pair.second})) {
+      // an App that is supposed to be installed is not fully installed
       apps_to_update.insert(app_pair);
-      LOG_INFO << app_name << " update will be re-installed or completed";
+      apps_and_reasons[app_pair.first] = "not fetched";
+      LOG_INFO << app_name << " is not fully fetched; missing blobs will be fetched";
+      continue;
+    }
+    if (!app_engine_->isRunning({app_name, app_pair.second})) {
+      // an App that is supposed to running is not running
+      apps_to_update.insert(app_pair);
+      apps_and_reasons[app_pair.first] = "not running";
+      LOG_INFO << app_name << " is not installed or not running; will be installed and started";
       continue;
     }
   }
@@ -237,13 +249,19 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
   return apps_to_update;
 }
 
-bool ComposeAppManager::checkForAppsToUpdate(const Uptane::Target& target) {
-  cur_apps_to_fetch_and_update_ = getAppsToUpdate(target);
+ComposeAppManager::AppsSyncReason ComposeAppManager::checkForAppsToUpdate(const Uptane::Target& target) {
+  AppsSyncReason apps_and_reasons;
+  cur_apps_to_fetch_and_update_ = getAppsToUpdate(target, apps_and_reasons);
   if (!!cfg_.reset_apps) {
     cur_apps_to_fetch_ = getAppsToFetch(target);
   }
   are_apps_checked_ = true;
-  return cur_apps_to_fetch_and_update_.empty() && cur_apps_to_fetch_.empty();
+  for (const auto& app : cur_apps_to_fetch_) {
+    if (apps_and_reasons.count(app.first) == 0) {
+      apps_and_reasons[app.first] = "not fetched (reset apps)";
+    }
+  }
+  return apps_and_reasons;
 }
 
 DownloadResult ComposeAppManager::Download(const TufTarget& target) {
--- a/src/cli/cli.cc
+++ b/src/cli/cli.cc
@@ -300,7 +300,7 @@ static StatusCode pullAndInstall(AkliteClientExt &client, int version, const std
     LOG_WARNING << "Downgrading from " << current.Version() << " to  " << gti_res.selected_target.Version() << "...";
   }
 
-  auto pi_res = client.PullAndInstall(gti_res.selected_target, "", "", install_mode, local_update_source,
+  auto pi_res = client.PullAndInstall(gti_res.selected_target, gti_res.reason, "", install_mode, local_update_source,
                                       pull_mode == PullMode::All, do_install);
   return res2StatusCode<InstallResult::Status>(i2s, pi_res.status);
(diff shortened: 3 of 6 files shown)


src/aklite_client_ext.cc
```diff
@@ -120,15 +120,17 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
...
       // Force installation of apps
       res.selected_target = current;
       LOG_INFO
           << "The specified Target is already installed, enforcing installation to make sure it's synced and running:"
           << res.selected_target.Name();
 
       res.status = GetTargetToInstallResult::Status::UpdateSyncApps;
-      res.reason = "Syncing Active Target Apps";
+      res.reason =
+          "Syncing Active Target Apps\nApps requiring update:\n" + boost::algorithm::join(apps_to_update, "\n");
```
mike-sul, line 133: IMO, "Apps requiring update" is redundant after this "Syncing Active Target Apps".
So, I would compose slightly different log:
Instead of
```
 Details:
 | Syncing Active Target Apps
 | Apps requiring update:
 | ml-app (not fetched / not running)
```
produce
```
 Details:
 | Syncing Active Target Apps
 | - ml-app: not fetched / not running
 | - app01: <reason>
 | ...
```

src/composeappmanager.cc
```diff
@@ -237,13 +237,22 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
...
 
-bool ComposeAppManager::checkForAppsToUpdate(const Uptane::Target& target) {
+std::vector<std::string> ComposeAppManager::checkForAppsToUpdate(const Uptane::Target& target) {
   cur_apps_to_fetch_and_update_ = getAppsToUpdate(target);
   if (!!cfg_.reset_apps) {
     cur_apps_to_fetch_ = getAppsToFetch(target);
   }
   are_apps_checked_ = true;
-  return cur_apps_to_fetch_and_update_.empty() && cur_apps_to_fetch_.empty();
+  std::vector<std::string> ret;
```
mike-sul: Why do we need these two loops?

src/composeappmanager.cc
```diff
@@ -206,13 +208,15 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
     if (app_data.empty()) {
       // new app in Target
       apps_to_update.insert(app_pair);
+      apps_and_reasons.push_back(app_pair.first + " (new app in target)");
```
mike-sul, line 211: I don't think this case can happen during the target "sync". IIRC, it can happen only during new target installation. If so, then this is redundant, no point in adding it.

src/composeappmanager.cc
```diff
@@ -206,13 +208,15 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
...
       LOG_INFO << app_name << " will be installed";
       continue;
     }
 
     if (app_pair.second != app_data["uri"].asString()) {
       // an existing App update
       apps_to_update.insert(app_pair);
+      apps_and_reasons.push_back(app_pair.first + " (new version)");
```
mike-sul, line 219: The same as above, this is the new target case - in this case a reason of update is a new Target update.

src/composeappmanager.cc
```diff
@@ -221,6 +225,7 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
         !boost::filesystem::exists(cfg_.apps_root / app_name / Docker::ComposeAppEngine::ComposeFile)) {
       // an App that is supposed to be installed has been removed somehow, let's install it again
       apps_to_update.insert(app_pair);
+      apps_and_reasons.push_back(app_pair.first + " (removed)");
```
mike-sul, line 228: Maybe something like: "app: missing installation, to be re-installed".

src/composeappmanager.cc
```diff
@@ -229,6 +234,7 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
     if (!app_engine_->isFetched({app_name, app_pair.second}) || !app_engine_->isRunning({app_name, app_pair.second})) {
       // an App that is supposed to be installed and running is not fully installed or running
       apps_to_update.insert(app_pair);
+      apps_and_reasons.push_back(app_pair.first + " (not fetched / not running)");
```
mike-sul, line 237: I would break this "if" into two parts to detect the exact reason, whether app is not fetched or not running. Then the log message will be more precise, e.g.

```
 - app01: not fetched
 - app02: not running
```

mike-sul:
@detsch This is the nice idea, thanks for doing it.

src/composeappmanager.cc
detsch: To merge the app names from both cur_apps_to_fetch_and_update_ and cur_apps_to_fetch_. Next commit changes it anyway, perhaps I should squash the 2 last commits, if we decide that we want reason to be reported as well.

src/aklite_client_ext.cc
detsch: Updated.

src/composeappmanager.cc
detsch: Done, now we have 2 ifs

detsch:
I've updated the internal structure used to store the apps and reasons from vector to map, so the format is fully applied when building the reason string, not concatenating app name and reason before that.

src/composeappmanager.cc
detsch: I would keep adding a reason for all situations, even when they are not supposed to happen for the "app sync" case. This way we could easily spot an unexpected situation from the event alone. Also, if we eventually implement additional details for a "target update to N+1" event, we already have this information at hand.

src/composeappmanager.cc
```diff
@@ -238,21 +244,19 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
...
   }
   are_apps_checked_ = true;
-  std::vector<std::string> ret;
-  for (const auto& app : cur_apps_to_fetch_and_update_) {
-    ret.push_back(app.first);
-  }
   for (const auto& app : cur_apps_to_fetch_) {
-    if (std::find(ret.begin(), ret.end(), app.first) == ret.end()) {
-      ret.push_back(app.first);
+    auto app_and_reason = app.first + "(not fetched)";
```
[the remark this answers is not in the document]
detsch: Like for https://github.com/foundriesio/aktualizr-lite/pull/357#discussion_r1737102405, I would rather keep it. I'm using a different reason string for that specific case, so we can spot an unexpected situation immediately by looking at the event.

detsch:
@mike-sul I've taken a look into adding apps to update + reasons to the "update to target N+1" event, but it is not as trivial as this PR. We currently send the `DownloadStarted` event before knowing which apps need to be updated, and the code structure there is more complex, a lot more changes would be needed. I'll keep an eye on it and if I visualize some change that makes sense in that direction, I may try something. But it is not worth to deal with it in this PR.

src/aklite_client_ext.cc
```diff
@@ -129,8 +129,10 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
           << res.selected_target.Name();
 
       res.status = GetTargetToInstallResult::Status::UpdateSyncApps;
-      res.reason =
-          "Syncing Active Target Apps\nApps requiring update:\n" + boost::algorithm::join(apps_to_update, "\n");
+      res.reason = "Syncing Active Target Apps\n";
```
mike-sul, line 132: I think we can combine the last two commits into a single commit as they change the same code lines twice.

src/composeappmanager.cc
```diff
@@ -221,38 +225,45 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
...
 
     LOG_DEBUG << app_name << " performing full status check";
-    if (!app_engine_->isFetched({app_name, app_pair.second}) || !app_engine_->isRunning({app_name, app_pair.second})) {
-      // an App that is supposed to be installed and running is not fully installed or running
+    if (!app_engine_->isFetched({app_name, app_pair.second})) {
+      // an App that is supposed to be installed is not fully installed
       apps_to_update.insert(app_pair);
-      LOG_INFO << app_name << " update will be re-installed or completed";
+      apps_and_reasons[app_pair.first] = "not fetched";
+      LOG_INFO << app_name << " is not fully installed. Will be re-installed or have update completed";
```
mike-sul, line 238: Maybe `is not fully fetched`?

src/composeappmanager.cc
```diff
@@ -221,38 +225,45 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
...
+      continue;
+    }
+    if (!app_engine_->isRunning({app_name, app_pair.second})) {
+      // an App that is supposed to running is not running
+      apps_to_update.insert(app_pair);
+      apps_and_reasons[app_pair.first] = "not running";
+      LOG_INFO << app_name << " is not running. Will be re-installed or have update completed";
```
mike-sul, line 245: Maybe `is not installed or not running; will be installed and started`.

src/composeappmanager.cc
```diff
@@ -193,7 +194,8 @@ ComposeAppManager::AppsContainer ComposeAppManager::getApps(const Uptane::Target
   return apps;
 }
 
-ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane::Target& t) const {
+ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(
+    const Uptane::Target& t, std::unordered_map<std::string, std::string>& apps_and_reasons) const {
```
mike-sul, line 198: Maybe using `AppsContainer` type for `apps_and_reasons` instead of `std::unordered_map<std::string, std::string>` would make the change/code a bit cleaner?

src/aklite_client_ext.cc
mike-sul: Let's combine commits into a single one.

src/composeappmanager.cc
mike-sul: Let's squash the commits.

mike-sul:
> I'll keep an eye on it and if I visualize some change that makes sense in that direction, I may try something. But it is not worth to deal with it in this PR.

No need to worry about it. It is not so important for the regular update case (N to N+1 update).

src/composeappmanager.cc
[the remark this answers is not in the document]
mike-sul: ok

src/composeappmanager.cc
mike-sul: ok

src/composeappmanager.cc
detsch: I though about it. Problem is that currently `AppsContainer` contains `name` + `uri`. We could re-structure `AppsContainer` to make it contain an additional field, but it would be unset most of the time, and additional code changes would be needed. Just replacing the use of `uri` with `reason` in that specific case would be too confusing, IMO.

src/composeappmanager.cc
detsch: Updated. You are much more familiar with the details of each app installation situation, so I'm just following your lead on these two suggestions.

src/composeappmanager.cc
detsch: Updated.

src/aklite_client_ext.cc
detsch: Done.

src/composeappmanager.cc
detsch: Updated.

src/composeappmanager.cc
detsch: Commits squashed.

src/composeappmanager.cc
mike-sul: > We could re-structure `AppsContainer` to make it contain an additional field, but it would be unset most of the time

Why do we need any additional field in `AppsContainer`?

Effectively, `AppsContainer == std::unordered_map<std::string, std::string>`, this is exactly the type you use for the second param `apps_and_reasons`.
Therefore, `getAppsToUpdate(
    const Uptane::Target& t, std::unordered_map<std::string, std::string>& apps_and_reasons)` can be declared as `getAppsToUpdate(
    const Uptane::Target& t, AppsContainer& apps_and_reasons)`

detsch: I think reusing the `AppsContainer` type replacing the `uri` with `reason` as the `.second` field might led to confusion. I would expect `AppsContainer` to be used consistently in all locations.

mike-sul: Well, I don't see much problem in it as long as variable is named correspondingly. We use the same type, e.g. `std::unordered_map<std::string, std::string>` for many different things, and what exactly the container carries is determined rather by the usage context and variable name.

As an option, we can introduce a new type, e.g. `AppsSyncReason == std::unordered_map<std::string, std::string>` or something like that.

detsch: I'm OK with having  `AppsSyncReason`.
I've updated the commit accordingly.

mike-sul: Thanks.

mike-sul (approved):

PR #360, opened 2024-10-14
api: extended: Select target with all apps when running an apps sync

detsch:
Previous code was using the current target object, which is based on the current running system, and may not contain all available apps. This could lead to an issue where the apps sync update fails later on in the online case, since equality between the latest available and selected target is checked, and all apps must be included.

### DIFF
--- a/src/aklite_client_ext.cc
+++ b/src/aklite_client_ext.cc
@@ -123,7 +123,7 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
     auto apps_to_update = client_->appsToUpdate(Target::fromTufTarget(current));
     if (force_apps_sync || !apps_to_update.empty()) {
       // Force installation of apps
-      res.selected_target = current;
+      res.selected_target = checkin_res.SelectTarget(current.Version());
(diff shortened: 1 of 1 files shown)


src/aklite_client_ext.cc
```diff
@@ -123,7 +123,7 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
     auto apps_to_update = client_->appsToUpdate(Target::fromTufTarget(current));
     if (force_apps_sync || !apps_to_update.empty()) {
       // Force installation of apps
-      res.selected_target = current;
+      res.selected_target = checkin_res.SelectTarget(current.Version());
```
mike-sul, line 126: Will it work in the offline case? The checkin/TUF DB can have a target json with a list of all apps, while the target json stored for current target just some of the apps, and the offline update bundle includes only the shortlisted apps?

IIUC, we need addition `if` statement, `if (is_offline_more)`?

mike-sul: Ok, makes sense.

mike-sul (approved):

PR #361, opened 2024-11-06
End-to-end test improvements

detsch:
Adds a script to generate targets used by e2e test script. Allows the targets to be created on an arbitrary factory and tag.


detsch:
> @detsch It works fine for me, I would just update README.mk with instruction how to run the tests.

Added the instructions there now.

PR #362, opened 2024-11-15
composeappmanager: Stop apps only if rootfs update is expected

mike-sul:
Stop apps being updated only if ostree update and reboot is expected. Stopping apps in the case of an "app-only" update is not optimal, as it stops all app services/containers. In contrast, `docker compose up` can handle this more efficiently by restarting only those service containers whose images have been updated.


detsch (approved):
LGTM.
E2e test failed because of a change introduced in PR #361. Once we merge it, tests will succeed.

PR #364, opened 2024-11-21
readme: Add section guiding testing on devices

detsch:
@mike-sul I'm basically linking existing documentation. I could add more content to the README.md file itself, based on the linked documentation, if you think it is worth spending the time doing that.


mike-sul:
> I could add more content to the README.md file itself, based on the linked documentation, if you think it is worth spending the time doing that.

I think that's suffice for now, thanks!

mike-sul (approved):

PR #366, opened 2024-11-29
apps: Log execution of external commands during apps management

detsch:


src/exec.h
```diff
@@ -23,6 +24,7 @@ static void exec(const std::string& cmd, const std::string& err_msg_prefix, Args
   boost::asio::io_service io_service;
 
   try {
+    LOG_DEBUG << "Running: `" << cmd << "`";
```
mike-sul, line 27: I would make it `LOG_TRACE`.

mike-sul (approved):

PR #367, opened 2024-12-02
appengine: Check only if app installed and running

mike-sul:
Check only if app is installed and running at the "is app running" check at the beginning of each update cycle.
Whether app blobs are present&OK or not does NOT impact a properly installed and running app. Moreover, after "is app running" check, the aklite daemon checks whether all enabled apps (including the "reset" apps) are fully fetched. We need this check to make sure a user can switch (turn on/off) apps even in offline mode.
Therefore, this change removes the redundant check whether all app blobs are fetched&OK called at the "is app running" check.

### DIFF
--- a/src/docker/restorableappengine.cc
+++ b/src/docker/restorableappengine.cc
@@ -242,8 +242,7 @@ bool RestorableAppEngine::isRunning(const App& app) const {
   bool res{false};
 
   try {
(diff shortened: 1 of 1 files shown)


detsch:
What are the plans to add this to `master` branch as well?

detsch (approved):

PR #368, opened 2024-12-02
Optimize app fetch check

mike-sul:
Make sure that the check whether app is fetched is performed only once
during a single update cycle.

- If app is not running then don't check if it is fetched since it going
  to be re-updated anyway which implies checking for missing app blobs
  and pulling them if any.

- If app is enabled for running and is running then whether it is fetched
  or not is done before checking the "reset" apps fetch, so the later is skipped.

### DIFF
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -230,18 +230,18 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToUpdate(const Uptane
     }
 
     LOG_DEBUG << app_name << " performing full status check";
-    if (!app_engine_->isFetched({app_name, app_pair.second})) {
-      // an App that is supposed to be installed is not fully installed
-      apps_to_update.insert(app_pair);
-      apps_and_reasons[app_pair.first] = "not fetched";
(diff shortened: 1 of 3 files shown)


src/composeappmanager.cc
```diff
@@ -718,15 +721,26 @@ ComposeAppManager::AppsContainer ComposeAppManager::getRequiredApps(const Config
...
 
   AppsContainer apps_to_be_fetched;
   for (const auto& app : enabled_apps) {
+    if (checked_apps != nullptr && checked_apps->count(app.first) > 0) {
+      // no reason to check whether app is fetched since app is checked and marked for update because
+      // it is either not installed or not running;
+      continue;
+    }
+    if (fetched_apps != nullptr && fetched_apps->count(app.first) > 0) {
+      // no reason to check whether app is fecthed since this check was already done before because this
```
detsch, line 740:
```suggestion
      // no reason to check whether app is fetched since this check was already done before because this
```

detsch (approved):

src/composeappmanager.cc
mike-sul: Done

PR #369, opened 2024-12-06
apps: Avoid pruning unused apps on every check operation

detsch:
Only prune on startup, or when an installation is being finalized.


mike-sul (approved):

mike-sul:
@detsch I suppose we need to cherry-pick it into the v94 branch.

Also, another option would be doing similar in `AkliteClient::CheckAppsInSync()` (add else to if).

mike-sul (approved):

PR #370, opened 2024-12-06
appengine: Optimize app fetch check

mike-sul:
If an app is successfully fetched and checked then mark it as "fetched" so the next time the app fetch check is called no need to invoke the full blown check performed by `composectl`.


mike-sul:
@detsch This simple approach should work because:

1. If `composectl pull` is successful then it is guaranteed that an app is completely pulled and written/stored on a drive (it does checks internally and write through the tmp file technique).

2. It does the full check after aklite's restart.

detsch (approved):
LGTM

PR #376, opened 2024-12-11
Check app installation along with app running

mike-sul:
Check whether an app is installed while verifying whether an app is running or not in `appengine.isRunning()` call. It guarantees that the running app is exactly what is expected to be running.


mike-sul:
@detsch 
This PR depends on the `composectl` change https://github.com/foundriesio/composeapp/pull/40. Once the `composectl` is merged I'll add a commit to this PR with composectl version update (in the dev&test env).

detsch (approved):

PR #381, opened 2024-12-19
Add user-initiated rollback

detsch:


src/liteclient.cc
```diff
@@ -243,18 +243,18 @@ bool LiteClient::finalizeInstall(data::InstallationResult* ir) {
...
-  const auto rollback_hash{sysroot_->getDeploymentHash(OSTree::Sysroot::Deployment::kRollback)};
+Uptane::Target LiteClient::getRollbackTarget(bool ignore_ostree_hash) {
+  const auto rollback_hash{ignore_ostree_hash ? "" : sysroot_->getDeploymentHash(OSTree::Sysroot::Deployment::kRollback)};
 
   Uptane::Target rollback_target{Uptane::Target::Unknown()};
   {
     std::vector<Uptane::Target> installed_versions;
     storage->loadPrimaryInstallationLog(&installed_versions,
-                                        true /* make sure that Target has been successfully installed */);
+                                        true /* make sure that Target has been successfully installed */, false);
```
mike-sul, line 253: Did we allow to use the current target as "rollback" target before?

mike-sul:
LGTM, will test it a bit.

src/liteclient.cc
detsch: No. The current target was included in the list returned by `loadPrimaryInstallationLog`, but it was never the rollback target because it would not match `rollback_hash`.

src/aklite_client_ext.cc
```diff
@@ -276,3 +277,51 @@ bool AkliteClientExt::RebootIfRequired() {
...
+                                        const std::string& target_name) {
+  auto pending_target = Target::toTufTarget(client_->getPendingTarget());
+  auto current = GetCurrent();
+
+  auto storage = INvStorage::newStorage(client_->config.storage, false);
+  auto update_mode =
+      pending_target.IsUnknown() ? InstalledVersionUpdateMode::kBadTarget : InstalledVersionUpdateMode::kNone;
+
+  // Set current target as failing
+  storage->saveInstalledVersion("", Target::fromTufTarget(current), update_mode);
```
mike-sul, line 303: Is there guarantee that the `current` target is the same as the `pending` one if the `pending` is not null in all use-cases? For example, if ostree update is installed and the rollback command in invoked before device reboot?

src/aklite_client_ext.cc
```diff
@@ -276,3 +277,51 @@ bool AkliteClientExt::RebootIfRequired() {
 
   return false;
 }
+
+InstallResult AkliteClientExt::Rollback(const LocalUpdateSource* local_update_source, int version,
+                                        const std::string& target_name) {
+  auto pending_target = Target::toTufTarget(client_->getPendingTarget());
```
mike-sul, line 290: I suppose we can replace it with `GetPendingTarget()` (the base class implements it).

src/aklite_client_ext.cc
```diff
@@ -276,3 +277,51 @@ bool AkliteClientExt::RebootIfRequired() {
...
+  auto update_mode =
+      pending_target.IsUnknown() ? InstalledVersionUpdateMode::kBadTarget : InstalledVersionUpdateMode::kNone;
+
+  // Set current target as failing
+  storage->saveInstalledVersion("", Target::fromTufTarget(current), update_mode);
+
+  // Get rollback target
+  TufTarget rollback_target;
+  if (version == -1 && target_name.empty()) {
+    rollback_target = Target::toTufTarget(client_->getRollbackTarget(true));
```
mike-sul, line 308: I would change signature of `AkliteClient::GetRollbackTarget()` so it can accept the boolean parameter, or introduce another protected method so we use `GetRollbackTarget()` instead of `client_->getRollbackTarget`.

src/aklite_client_ext.cc
```diff
@@ -276,3 +277,51 @@ bool AkliteClientExt::RebootIfRequired() {
...
+    if (rollback_target.IsUnknown() || rollback_target.Name() == current.Name()) {
```
mike-sul, line 309: I am not sure if this `rollback_target.Name() == current.Name()` works if a new ostree version is installed and a device is not installed yet.

src/aklite_client_ext.cc
```diff
@@ -276,3 +277,51 @@ bool AkliteClientExt::RebootIfRequired() {
...
+      LOG_ERROR << "Failed to find Target to rollback to";
+      return {InstallResult::Status::Failed};
+    }
+  } else {
+    auto checkin_result = CheckInCurrent(local_update_source);
```
mike-sul, line 314: Why do we need need to update the local tuf meta during rollback?

src/aklite_client_ext.cc
mike-sul: I would print some logs saying that we are about to mark the given target as "failing/bad" target.

src/liteclient.cc
```diff
@@ -243,18 +243,19 @@ bool LiteClient::finalizeInstall(data::InstallationResult* ir) {
...
+Uptane::Target LiteClient::getRollbackTarget(bool ignore_ostree_hash) {
+  const auto rollback_hash{ignore_ostree_hash ? ""
+                                              : sysroot_->getDeploymentHash(OSTree::Sysroot::Deployment::kRollback)};
 
   Uptane::Target rollback_target{Uptane::Target::Unknown()};
   {
     std::vector<Uptane::Target> installed_versions;
     storage->loadPrimaryInstallationLog(&installed_versions,
-                                        true /* make sure that Target has been successfully installed */);
+                                        true /* make sure that Target has been successfully installed */, false);
```
mike-sul, line 254: The same question about considering the current target as the target to rollback to under certain circumstances.

src/aklite_client_ext.cc
```diff
@@ -278,6 +278,13 @@ bool AkliteClientExt::RebootIfRequired() {
   return false;
 }
 
+bool AkliteClientExt::isRollbackProtectionEnabled() {
+  OSTree::Sysroot::Ptr ostree_sysroot_;
+  auto ostree_sysroot = std::make_shared<OSTree::Sysroot>(client_->config.pacman);
+  auto bootloader_lite = bootloader::BootloaderLite(client_->config.bootloader, *client_->storage, ostree_sysroot);
+  return bootloader_lite.isRollbackProtectionEnabled();
```
mike-sul, line 285: I would add a method to `LiteClient` similar to `LiteClient::isBootFwUpdateInProgress()`.

src/aklite_client_ext.cc
```diff
@@ -287,6 +294,11 @@ InstallResult AkliteClientExt::Rollback(const LocalUpdateSource* local_update_so
   auto update_mode =
       pending_target.IsUnknown() ? InstalledVersionUpdateMode::kBadTarget : InstalledVersionUpdateMode::kNone;
 
+  if (pending_target.IsUnknown() && isRollbackProtectionEnabled()) {
```
mike-sul, line 297: Is it suffice just to check whether bootloader rollback protection is enabled?  What if the boot fw version of the rollback target is the same is as the current one?

I think it's problematic a bit to check it if the rollback target is not the previously installed target because we cannot fetch info about boot fw version of this target.

Maybe we don't need this check at all since the logic in the rootfs manager has the check if there is boot fw downgrade/rollback. So it will yield an error if it detects for the rollback target?

src/aklite_client_ext.cc
detsch: > Is there guarantee that the `current` target is the same as the `pending` one if the `pending` is not null in all use-cases? For example, if ostree update is installed and the rollback command in invoked before device reboot?

The `Set current target as failing` comment there is wrong. There are 2 different things that we do there, depending if there is or not a pending target:
- Setting the current target as bad (if the previous installation was finalized)
- Creating a new log entry related to the re-installation of the current target

Both happen to be done by calling `storage->saveInstalledVersion`, so when simplifying the code I left only one call, and forgot do adjust the comment accordingly.

I've now adjusted the code to make this distinction more clear, and also added log messages. Next force push will include the changes,

---

On a related note, the whole e2e test sequence for rollback runs in 3 scenarios:
- After a finalizing an installation;
- After rebooting, but before finalizing; and
- Before rebooting

This is done by using pytest parameters:
```
@pytest.mark.parametrize('do_reboot', [True, False])
@pytest.mark.parametrize('do_finalize', [True, False])
```

src/aklite_client_ext.cc
detsch: `CheckInCurrent` does not update the local tuf metadata. This call is only done to get the list of available targets, so we can select the one that matches the given version / name.

src/aklite_client_ext.cc
```diff
@@ -276,3 +277,51 @@ bool AkliteClientExt::RebootIfRequired() {
...
+    if (rollback_target.IsUnknown()) {
+      if (version != -1) {
+        LOG_ERROR << "Unable find target with version " << version;
+      } else {
+        LOG_ERROR << "Unable find target with name " << target_name;
+      }
+      return {InstallResult::Status::Failed};
+    }
+
+    if (rollback_target.Name() == current.Name()) {
```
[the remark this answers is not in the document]
detsch: Perhaps we should allow calling `aklite rollback <current_version>` only when there is a pending installation?

src/aklite_client_ext.cc
detsch: Currently approach is overly-protective, but having a more complete verification would add too much complexity. Assuming that a bootloader update is not very common, it makes sense to disable this check altogether.

src/liteclient.cc
detsch: A change here is probably needed, to allow current depending if the latest install has finished or not. I'll adjust it tomorrow.

src/aklite_client_ext.cc
detsch: Added. As a general rule, I was trying not to add complexity to the external API.

src/aklite_client_ext.cc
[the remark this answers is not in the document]
mike-sul: iirc, there is a pending installation once ostree is installed, even before reboot.

src/aklite_client_ext.cc
mike-sul: I incline towards not doing this check at all at the beginning of the user-triggered process, anyway the installation will fail during this process if the bootloader rollback protection is turned ON because of the check in the rootfs manager.

src/aklite_client_ext.cc
mike-sul: Yes, I was mistaken.

src/aklite_client_ext.cc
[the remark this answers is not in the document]
detsch: This behavior (a target showing as current even before reboot) is specific to the test environment (`config.booted != BootedType::kBooted`). I'm taking a look to see if there is an easy fix. I may have to adjust the rollback-related code based on the correct behavior.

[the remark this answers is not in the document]
mike-sul: Yes, if there is an ostree update, then after installation (new ostree version is deployed) and before reboot, the "get current" returns the version a device is currently booted on provided that a device is booted on ostree.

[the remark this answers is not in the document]
mike-sul: >The install step will already finalize the installation, so GetCurrent will return 4, and it makes no sense to try to rollback to 4.

What if an install of an app-only update fails, then a device is at kind of "undefined" state, so a user needs to run either the rollback command or just install one of the previous version which is almost the same as the user-driven rollback.

The daemon client will auto-rollback to the previous version automatically in this case. However, the CLI client as well as the API should support it and allow rolling back to the current version in this case which is effectively the "sync" current operation.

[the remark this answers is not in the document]
detsch: I've done a quick adjustment in `OstreeManager::getCurrent` when `config.booted != BootedType::kBooted`, to make it consistent with a booted device.
I'm in the process of adjusting the rollback code and tests accordingly.

src/aklite_client_ext.cc
```diff
@@ -282,3 +283,45 @@ bool AkliteClientExt::RebootIfRequired() {
 
   return false;
 }
+
+InstallResult AkliteClientExt::Rollback(const LocalUpdateSource* local_update_source) {
+  auto current = GetCurrent();
```
mike-sul, line 288: Just a nitpick, these `auto` better be `const auto`.

src/aklite_client_ext.cc
```diff
@@ -282,3 +283,45 @@ bool AkliteClientExt::RebootIfRequired() {
...
+  auto pending_target = GetPendingTarget();
+  auto installation_in_progress = !pending_target.IsUnknown();
+  auto bad_target = installation_in_progress ? pending_target : current;
```
mike-sul, line 291: Another nitpick, this can be `const auto&`.

src/aklite_client_ext.cc
```diff
@@ -282,3 +283,45 @@ bool AkliteClientExt::RebootIfRequired() {
...
+
+  LOG_DEBUG << "User initiated rollback. Current Target is " << current.Name();
+  if (installation_in_progress) {
+    LOG_DEBUG << "There is an installation in progress of Target " << pending_target.Name();
```
mike-sul, line 295: Maybe `Target installation is in progress: <target name>` or `There is Target installation in progress: <target name>`?

src/aklite_client_ext.cc
```diff
@@ -282,3 +283,45 @@ bool AkliteClientExt::RebootIfRequired() {
...
+  }
+  auto storage = INvStorage::newStorage(client_->config.storage, false);
+  LOG_INFO << "Marking target " << bad_target.Name() << " as failed";
+  storage->saveInstalledVersion("", Target::fromTufTarget(bad_target), InstalledVersionUpdateMode::kBadTarget);
+
+  // Get rollback target
+  auto rollback_target = GetRollbackTarget(true, installation_in_progress);
```
mike-sul, line 302: If this call takes place after reboot, then we shouldn't allow choosing the current target as the target to rollback to since the "bad" target is already current?
Or I missed something?

src/aklite_client_ext.cc
```diff
@@ -282,3 +283,45 @@ bool AkliteClientExt::RebootIfRequired() {
...
+  auto pending_target = GetPendingTarget();
+  auto installation_in_progress = !pending_target.IsUnknown();
+  auto bad_target = installation_in_progress ? pending_target : current;
+
+  LOG_DEBUG << "User initiated rollback. Current Target is " << current.Name();
+  if (installation_in_progress) {
+    LOG_DEBUG << "There is an installation in progress of Target " << pending_target.Name();
+  }
+  auto storage = INvStorage::newStorage(client_->config.storage, false);
+  LOG_INFO << "Marking target " << bad_target.Name() << " as failed";
```
mike-sul, line 298: Maybe `... as a failing target` to make it consistent with the log at line 315?

src/aklite_client_ext.cc
```diff
@@ -282,3 +283,45 @@ bool AkliteClientExt::RebootIfRequired() {
...
+
+  std::string reason = "User initiated rollback. Marked " + bad_target.Name() +
+                       " as a failing target, and rolling back to " + rollback_target.Name();
+  LOG_INFO << reason;
+  // First, try to rollback without download
+  auto ret = PullAndInstall(rollback_target, reason, "", InstallMode::All, local_update_source, false, true,
+                            !installation_in_progress);
+  if (ret.status == InstallResult::Status::DownloadFailed) {
+    LOG_INFO
+        << "Available data for rollback not fully available. Trying to download data before trying to install again";
```
mike-sul, line 322: Maybe `Not all rollback Target data are available, downloading them before the next install attempt...`?

mike-sul (approved):
LGTM, I have just a few nitpicks.

mike-sul (approved):

mike-sul (approved):

PR #382, opened 2025-01-14
Fix rollback and apps sync update when target is not listed in TUF anymore

detsch:
Still doing a few tests

### DIFF
--- a/src/api.cc
+++ b/src/api.cc
@@ -1038,7 +1038,8 @@ std::unique_ptr<InstallContext> AkliteClient::CheckAppsInSync() const {
 
 std::unique_ptr<InstallContext> AkliteClient::Installer(const TufTarget& t, std::string reason,
                                                         std::string correlation_id, InstallMode install_mode,
-                                                        const LocalUpdateSource* local_update_source) const {
+                                                        const LocalUpdateSource* local_update_source,
+                                                        bool require_target_in_tuf) const {
   if (read_only_) {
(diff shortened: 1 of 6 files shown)


src/api.cc
```diff
@@ -1061,7 +1062,7 @@ std::unique_ptr<InstallContext> AkliteClient::Installer(const TufTarget& t, std:
   }
   if (target == nullptr) {
     const auto uptane_target{Target::fromTufTarget(t)};
-    if (Target::isInitial(uptane_target) && client_->wasTargetInstalled(uptane_target)) {
+    if ((!require_target_in_tuf || Target::isInitial(uptane_target)) && client_->wasTargetInstalled(uptane_target)) {
```
mike-sul, line 1065: I would update the comment below with info about this use-case too.

detsch: @mike-sul  comment updated. Can you take a look to see it if is OK?
Is there any other situation, other than a rollback, that the "Initial target" could be used? If not, perhaps we can remove the `Target::isInitial(uptane_target)` test, and leave only `require_target_in_tuf`

mike-sul: >Is there any other situation, other than a rollback, that the "Initial target" could be used? If not, perhaps we can remove the Target::isInitial(uptane_target) test, and leave only require_target_in_tuf

Yes, there are such situations, e.g. customer can play with sql.db, or remove the `installed_versions` file, so then we can end up doing a regular update or sync for "initial target" which might not be among targets listed in the local tuf db.

PR #384, opened 2025-01-22
[OPS] Add downloading of new apps to finalize cycle

### LINKED ISSUE #383: New App In Compose_apps Toml List Causes Target Failure and Rollback
### Description:
When a new application is added to the ```compose_apps``` list, the target gets marked as failed during the finalize step after a reboot. The issue arises because the composeManager attempts to start all apps in the updated ```compose_apps``` list. This includes the new app, which has not yet been pulled during the update process, leading to a failure.

The failure occurs due to the following sequence of events:

    Update Process:
        When an update is detected, aktualizr-lite only pulls apps that are currently known in the compose_apps list.
        The new app is not part of this initial pull, as it was not known at that stage.

    Reboot and Finalize Step:
        After a reboot, the composeManager attempts to start all apps listed in the updated compose_apps list, including the new app.
        Since the new app has not been pulled and is therefore not present, the startup fails.

    Result:
        The target is locally marked as failed.
        The device performs a rollback to the previous state.

### Steps to Reproduce:

    Add a new app to the compose_apps list.
    Trigger an update and allow the system to reboot.
    Observe that the new app is not pulled, leading to failure in the finalize step.

### Expected Behavior:

    New apps in the compose_apps list should be successfully pulled and started without causing target failure or rollback.

### Actual Behavior:

    The new app is not pulled during the update process.
    After reboot, the app fails to start, marking the target as failed and causing a rollback.

### Impact:

    Prevents successful addition and activation of new apps in the system.

NoelsJaan:
*Retry Mechanism:*

- Implement functionality to retry installing apps that fail to run during the first attempt. By retrying the installation of failed apps, the system can handle transient issues and ensure successful addition of new apps.

Possibly closes #383


mike-sul:
Hi @NoelsJaan,

Thank you for posting this PR.  

Based on my current understanding of your use case, this change should help address the issue you’ve encountered.  

Before we can accept this change, we need to:  

1. Ensure it does not break any other use cases or their corner cases (this is an action item for us).  
2. Review the low-level technical details of this change. I’ll post comments on this soon.  

I have one question: my understanding is that you are currently using LmP version 94 and need to address this issue in v94. If that’s correct, please rebase this PR to the `v94` branch accordingly.

src/composeappmanager.cc
```diff
@@ -475,23 +475,36 @@ data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target
     }
     // "finalize" (run) Apps that were pulled and created before reboot
     for (const auto& app_pair : getApps(target)) {
+      //Here the installation of the apps should be checked, and if they are not installed, they should be
```
mike-sul, line 478: Most likely, you're correct, we should check if all target apps are pulled at this stage of the OTA update process and if not then pull them. Also, we may need to make it configurable, since this logic won't work for all use-cases we need to support (e.g. if a device needs to operate at a fully offline mode after reboot caused by an update).

src/composeappmanager.cc
```diff
@@ -475,23 +475,36 @@ data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target
...
+        LOG_ERROR << "App: " << app_pair.first << ", failed to start: " << run_res.err;
+        // If this fails it is possible that some apps are missing since they are added in this new target.
+        LOG_INFO << "Downloading App because it would not start";
+        Download(Target::toTufTarget(target));  // download the missing apps
+                                                // Retry the run
+        LOG_INFO << "Retrying to start App: " << app_pair.first;
+        const AppEngine::Result rerun_res = app_engine_->run({app_pair.first, app_pair.second});
+        if (!rerun_res) {
+        // again we print the run_res error message
+        LOG_ERROR << "App: " << app_pair.first << ", failed to start at rerun: " << rerun_res.err;
```
mike-sul, line 491: I would just extend the `err_desc` with the `rerun_res.err` below and remove this line. The following error log will output all information about the error, both the start and the repull error.

src/composeappmanager.cc
```diff
@@ -475,23 +475,36 @@ data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target
...
+
+          LOG_ERROR << err_desc;
+          // Do we need to set some flag for the uboot and trigger a system reboot in order to boot on a previous
+          // ostree version, hence a proper/full rollback happens???
+          ir.description += ", however " + err_desc;
+          ir.description += "\n# Apps running:\n" + getRunningAppsInfoForReport();
+          // this is a hack to distinguish between ostree install (rollback) and App start failures.
+          // data::ResultCode::Numeric::kInstallFailed - boot on a new ostree version failed (rollback at boot)
+          // data::ResultCode::Numeric::kCustomError - boot on a new version was successful but new App failed to start
+          return data::InstallationResult(run_res.imagePullFailure() ? data::ResultCode::Numeric::kDownloadFailed
```
mike-sul, line 504: Hm, this is problematic. Unfortunately, the code that invokes this method does not handle the case when `rerun_res.noSpace() == true`, it simply does not expect that the update "finalization" pulls any data. In this case we should NOT do rollback, while the current solution will do it if there is not enough free storage to accommodate new app...

mike-sul:
Currently, I see two issues with the given approach:

1. If there is not enough free storage to fit a new app, then it will lead to rollback while it should not.

2. It will lead to huge delay during update finalization and rollback if a device operates at a fully offline mode after reboot and app failure starts occur (some issue with app which is fully pulled).

mike-sul:
I think we should reconsider approach slightly in order to fulfill all edge cases (e.g. no space) and all other use-cases. At first glance, we need to do the following:

1. We need to detect whether a required app is missing or not. All other errors should be treated as it now.

2. If an app is missing then ignore it at the finalization stage, and let the finalize method to start all available apps. Then, the following "sync" update will pull and start all missing apps (or removed one); the "sync" update already handle all possible edge-cases.

NoelsJaan:
Thank you for your feedback Mike!

I agree that the reconsideration of the approach will be a more solid solution. I will have a look at the possibility to detect the error and to differentiate it from the others. That will be a great first step.

mike-sul:
@NoelsJaan My understanding is that you are currently using the LmP version 94 and need to address this issue in v94?
If that’s correct, please rebase this PR to the v94 branch accordingly.

src/composeappmanager.cc
```diff
@@ -475,23 +475,50 @@ data::InstallationResult ComposeAppManager::finalizeInstall(const Uptane::Target
...
-        // data::ResultCode::Numeric::kCustomError - boot on a new version was successful but new App failed to start
-        return data::InstallationResult(run_res.imagePullFailure() ? data::ResultCode::Numeric::kDownloadFailed
-                                                                   : data::ResultCode::Numeric::kCustomError,
-                                        ir.description);
+        //Make sure not "No space or imagePullFailure"
+        if (!run_res.imagePullFailure() && !run_res.noSpace()) {
+          LOG_INFO << "(Re-)downloading app before retrying to start it again; app: " << app_pair.first;
+          const DownloadResult download_res = Download(Target::toTufTarget(target)); // download the missing apps
+          if(!download_res) {
+            LOG_ERROR << "Failed to (re-)download the missing apps; app: " << app_pair.first;
```
mike-sul, line 496: Does it make sense to continue the `for` loop if an error to re-download app occurs?

NoelsJaan:
@mike-sul yes, this should fix it. Thank you.

mike-sul:
@NoelsJaan FYI, The proposed fix has been merged into the `v94` branch https://github.com/foundriesio/aktualizr-lite/commit/a26ef85a009405c3ee7acb4810a2d83c3dd3cf7d.
Also, it has been pushed to the upcoming `v95` version https://github.com/foundriesio/aktualizr-lite/commit/4a54e8dea654babaad25989fc48f548c37bc7ff0.

If you think that the solution in this PR is still needed then I will create a new PR where both approaches are combined, i.e. optionally (defined by a new configuration variable) an app can be pulled during finalization.

PR #389, opened 2025-01-31
api: Prevent downgrade operations to be done automatically

detsch:
When the latest available target has a version lower than the target that is currently running, we may or not accept it as a valid update to be performed.

Before this commit, aktualizr-lite would automatically select such target and proceed with the update operation. After this commit, the target is not longer selected by default as a valid update target.

It is possible to revert the behavior to the original one by defining ACCEPT_LOWER_VERSION at build time.

---
@mike-sul in order to keep original aklite-offline behavior, I had to keep the `force_downgrade` parameter, and add a new one in `cli::` methods. Using this new parameter we can easily add a command line option for CLI commands. But in order to avoid having users rely too much on automatic downgrades, it seems better to not have those command line options for now.


src/CMakeLists.txt
```diff
@@ -72,6 +72,10 @@ if(BUILD_P11)
   add_definitions(-DBUILD_P11)
 endif(BUILD_P11)
 
+if(ACCEPT_LOWER_VERSION)
+  add_definitions(-DACCEPT_LOWER_VERSION)
```
mike-sul, line 76: I wonder if we could define `ACCEPT_LOWER_VERSION` as `true` or `false` here so we don't need the additional macro definition `AKLITE_ACCEPT_LOWER_VER_DEFAULT`?

mike-sul: We have to use 1 or 0 as values then, I suppose.

src/aklite_client_ext.cc
```diff
@@ -55,6 +56,18 @@ GetTargetToInstallResult AkliteClientExt::GetTargetToInstall(const CheckInResult
...
 
   const auto current = GetCurrent();
+  // It may occur that the TUF targets list only has versions lower than the current one.
+  // The `accept_lower_ver` parameter controls what to do in such situation: Should a version lower than the current
+  //  one be accepted as a valid selected target for installation or not
+  if (!accept_lower_ver && version == -1 && target_name.empty() && candidate_target.Version() < current.Version()) {
+    if (!invoke_post_cb_at_checkin_) {
+      client_->notifyTufUpdateFinished(err);
+    }
+    LOG_INFO << "Preventing selection of a downgrade target. Current: " << current.Version()
```
mike-sul, line 66: Maybe `Rejecting latest target in TUF metadata to prevent downgrade; current: <>, candidate: <>`?

mike-sul (approved):

src/CMakeLists.txt
detsch: IMO, we would need a `#ifndef` block anyway, to avoid making the macro definition obligatory when building. And, although it would work for the current implementation, we may want to implement the downgrade control in a different way, so using this more abstract approach seems to be better.

mike-sul (approved):

PR #390, opened 2025-02-03
offline: Improve usage help text

detsch:
Adds a description to each subcommand, and use a format similar to fioctl help.

---
@vanmaegima suggestions about the newly added description text for each subcommand are welcome (look for `Cmd("` in the changes bellow)

### DIFF
--- a/apps/aklite-offline/cmds.h
+++ b/apps/aklite-offline/cmds.h
@@ -18,21 +18,28 @@ class Cmd {
  public:
   using Ptr = std::shared_ptr<Cmd>;
   const std::string& name() const { return _name; }
+  const std::string& description() const { return _description; }
   const po::options_description& options() const { return _options; }
 
   virtual int operator()(const po::variables_map& vm) const = 0;
 
  protected:
-  Cmd(std::string name, const po::options_description& options) : _name{std::move(name)}, _options{options} {}
+  Cmd(std::string name, std::string description, const po::options_description& options)
+      : _name{std::move(name)}, _description{std::move(description)}, _options{options} {}
 
  private:
   const std::string _name;
+  const std::string _description;
   const po::options_description& _options;
 };
 
 class CheckCmd : public Cmd {
  public:
-  CheckCmd() : Cmd("check", _options) {
+  CheckCmd()
+      : Cmd("check",
+            "Update the device TUF metadata by fetching and validating the offline bundle's metadata. The list of "
+            "available targets is printed",
+            _options) {
     _options.add_options()("help,h", "print usage")("log-level", po::value<int>()->default_value(2),
                                                     "set log level 0-5 (trace, debug, info, warning, error, fatal)")(
         "config,c", po::value<std::vector<boost::filesystem::path>>()->composing(), "Configuration file or directory")(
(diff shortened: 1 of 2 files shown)


detsch:
If we feel this short description is not enough, we can add a long description that is only shown with `aklite-offline <cmd> --help`, and not when listing the subcommands in `aklite-offline --help`

detsch:
Once we agree on the format of the help for aklite-offline, I'll do a similar improvement to aktualizr-lite CLI help.

apps/aklite-offline/cmds.h
```diff
@@ -88,7 +99,9 @@ class InstallCmd : public Cmd {
 
 class RunCmd : public Cmd {
  public:
-  RunCmd() : Cmd("run", _options) {
+  RunCmd()
+      : Cmd("run", "Finalizes the installation of a target that required a reboot, starting the updated apps",
```
vanmaegima, line 103: Is 'run' not required for an apps only update anymore?

vanmaegima:
@detsch I'm okay with the short description.

apps/aklite-offline/cmds.h
detsch: You're right, we did not change that behavior. I've updated the command description.

apps/aklite-offline/cmds.h
```diff
@@ -18,21 +18,28 @@ class Cmd {
...
   const std::string _name;
+  const std::string _description;
   const po::options_description& _options;
 };
 
 class CheckCmd : public Cmd {
  public:
-  CheckCmd() : Cmd("check", _options) {
+  CheckCmd()
+      : Cmd("check",
```
mike-sul, line 39: The `check` command also checks/validates the TUF metadata of the bundle as well as its signature and validity.

apps/aklite-offline/cmds.h
```diff
@@ -57,7 +64,11 @@ class CheckCmd : public Cmd {
 
 class InstallCmd : public Cmd {
  public:
-  InstallCmd() : Cmd("install", _options) {
+  InstallCmd()
+      : Cmd("install",
+            "Installs the selected target. If no target name is specified, the highest version in the bundle will "
```
mike-sul, line 69: I think it should be without `s` endings, it not like `it installs` it is rather instruction what to do `do install`. The same for the rest commands. The `fioctl` defines verbs/actions without ending "(e)s" as well.

apps/aklite-offline/cmds.h
```diff
@@ -112,7 +123,7 @@ class RunCmd : public Cmd {
 
 class CurrentCmd : public Cmd {
  public:
-  CurrentCmd() : Cmd("current", _options) {
+  CurrentCmd() : Cmd("current", "Shows information of the currently running target", _options) {
```
mike-sul, line 126: Maybe `about` instead of `of`?

mike-sul (approved):
LGTM, just a few minor comments.

apps/aklite-offline/cmds.h
detsch: I've now removed the "s"s from the desctiption.

apps/aklite-offline/cmds.h
detsch: Updated

apps/aklite-offline/cmds.h
detsch: I'm now mentioning the validation in the description. Tried to keep the description short, so I suppose "validation" implies that the signature and validity are verified.

apps/aklite-offline/cmds.h
```diff
@@ -18,21 +18,28 @@ class Cmd {
...
+  const std::string _description;
   const po::options_description& _options;
 };
 
 class CheckCmd : public Cmd {
  public:
-  CheckCmd() : Cmd("check", _options) {
+  CheckCmd()
+      : Cmd("check",
+            "Update the device TUF metadata reading and validating the metadata of the offline bundle. The list of "
```
mike-sul, line 40: Maybe `Update the device TUF metadata by fetching and validating the offline bundle's metadata...`?

mike-sul (approved):

apps/aklite-offline/cmds.h
detsch: Updated to the suggested text

PR #394, opened 2025-02-05
main: Improve help message

detsch:
Add a description to every subcommand, and improve the help output format.

----

Doing a similar improvement to that I've done in aklite-offline (PR #390).

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -14,6 +14,7 @@
 #include <boost/format.hpp>
 #include <boost/interprocess/sync/file_lock.hpp>
 #include <boost/program_options.hpp>
+#include <vector>
 
 #include "aktualizr-lite/aklite_client_ext.h"
 #include "aktualizr-lite/api.h"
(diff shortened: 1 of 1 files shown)


vanmaegima (approved):

src/main.cc
```diff
@@ -230,20 +231,35 @@ static int cli_rollback(LiteClient& client, const bpo::variables_map& params) {
...
+  const std::string name;
+  const std::string description;
+  int (*func)(LiteClient&, const bpo::variables_map&);
 };
+
+// clang-format off
+static const std::array<Cmd, 10> commands = {{
+    {"daemon", "Start the update agent daemon", daemon_main},
+    {"update", "Update TUF metadata, download and install the selected target", cli_update},
+    {"pull", "Download or copy the target data to the device, to allow a install operation to be performed", cli_pull},
```
mike-sul, line 244: I would remove `or copy`, if a offline bundle data are located on a thumb drive it is still download.

mike-sul: Maybe just `Download the selected or specified target data`?

mike-sul:
@kprosise Please, look into the command help message since it is kind of part of a user documentation.

src/main.cc
detsch: Updated.

mike-sul (approved):

kprosise (approved):
LGTM!

PR #409, opened 2025-03-28
Fix apps-only rollback triggered during finalization

detsch:
Once approved, I'll proceed to backport the changes to `v94`.

### DIFF
--- a/src/liteclient.cc
+++ b/src/liteclient.cc
@@ -243,25 +243,16 @@ bool LiteClient::finalizeInstall(data::InstallationResult* ir) {
   return ret.result_code.num_code == data::ResultCode::Numeric::kOk;
 }
 
-Uptane::Target LiteClient::getRollbackTarget(bool ignore_ostree_hash, bool allow_current) {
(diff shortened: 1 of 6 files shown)


src/liteclient.cc
```diff
@@ -243,25 +243,16 @@ bool LiteClient::finalizeInstall(data::InstallationResult* ir) {
...
-        rollback_target = *it;
-        break;
-      }
-    }
+Uptane::Target LiteClient::getRollbackTarget(bool allow_current) {
+  std::vector<Uptane::Target> installed_versions;
+  storage->loadPrimaryInstallationLog(&installed_versions,
+                                      true /* make sure that Target has been successfully installed */, allow_current);
+  if (installed_versions.empty()) {
+    LOG_WARNING << "No fully installed target found to rollback to";
```
mike-sul, line 251: I think we need to tackle this case next or at least create a jira ticket.

detsch: I've created a ticket for that.

PR #419, opened 2025-06-09
e2e-tests: use the latest boost

quaresmajose:
This will provide faster adoption of APIs that will be deprecated


quaresmajose:
Take the instructions from there https://www.boost.org/doc/user-guide/getting-started.html

detsch:
Another issue is that the `foundries/aklite-dev` container already installs some boost packages (version 1.74.0), and this generates a version mismatch between compilation and linking phases of current aklite build scripts. We would have to update `docker/Dockerfile` as well, or uninstall the packages inside `docker-e2e-test/Dockerfile`. This are the installed boost packages:
```
  libboost-dev \
  libboost-log-dev \
  libboost-program-options-dev \
  libboost-system-dev \
  libboost-test-dev \
  libboost-thread-dev \
```

PR #420, opened 2025-06-11
cli: Use read-only storage for `list` and `status` commands

detsch:
Also avoid enforce locking when running these commands. The changes prevent unnecessary warning messages to be show, and execution to be blocked by the lock file.


mike-sul:
Good idea!

PR #421, opened 2025-06-17
Support for Boost >= 1.88.0

detsch:
Keep using the deprecated v1 process API, but adjust the namespace to allow the binaries to be build with Boost 1.88.0.

It is possible to reduce the reliance on the deprecated API, I did code changes in that direction, but I believe we can merge this PR first, and then gradually get other improvements integrated as we find appropriate.

### DIFF
--- a/src/exec.h
+++ b/src/exec.h
@@ -4,7 +4,33 @@
 #include <boost/asio/io_context.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/format.hpp>
+
+#include <boost/version.hpp>
+
+#if BOOST_VERSION >= 108800
+#include <boost/process/v1/args.hpp>
+#include <boost/process/v1/async.hpp>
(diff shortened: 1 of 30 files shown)


quaresmajose (approved):
LGTM
I have tried that on the lmp/master branch and it builds weel. Not test at runtime, just building.

src/exec.h
```diff
@@ -4,7 +4,33 @@
 #include <boost/asio/io_context.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/format.hpp>
+
+#include <boost/version.hpp>
+
+#if BOOST_VERSION >= 108800
```
mike-sul, line 10: I wonder why they didn't do such "#if/#else" in the boost itself for the `boost::process` namespace.

mike-sul:
@detsch Thanks for doing it.

PR #422, opened 2025-06-17
Avoid deprecated boost process API

detsch:
Not sure we want to add those changes for now, as PR #421 already address the more urgent situation of building against Boost 1.88.0.

As stated in https://github.com/foundriesio/aktualizr-lite/commit/0983295d30b4e526ae725c9495f4f45be994d377 message, there are a few simplifications that were done.

### DIFF
--- a/src/exec.h
+++ b/src/exec.h
@@ -5,32 +5,6 @@
 #include <boost/filesystem.hpp>
 #include <boost/format.hpp>
 
-#include <boost/version.hpp>
-
-#if BOOST_VERSION >= 108800
-#include <boost/process/v1/args.hpp>
-#include <boost/process/v1/async.hpp>
-#include <boost/process/v1/async_system.hpp>
-#include <boost/process/v1/child.hpp>
-#include <boost/process/v1/cmd.hpp>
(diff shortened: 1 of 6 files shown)


src/exec.h
```diff
@@ -43,40 +17,58 @@ struct ExecError : std::runtime_error {
...
+    throw std::runtime_error("exec: popen() failed!");
+  }
+  std::string err_msg;
   try {
-    LOG_DEBUG << "Running: `" << cmd << "`";
-    bp::child child_process(cmd, bp::std_err > err_output, bp::on_exit = child_process_exit_code, io_context,
-                            std::forward<Args>(args)...);
-
-    io_context.run();
+    std::string result;
```
mike-sul, line 34: Maybe we can declare `result` before the `try`, in this case the `err_msg` can be set in the `if (exit_code != EXIT_SUCCESS)` scope.

detsch: I've removed the `err_msg` variable, as it was redundant with `result`, and adjusted the scope of the `result` variable.

mike-sul (approved):

mike-sul:
@detsch Thanks for this!

PR #427, opened 2025-06-30
e2e-tests: Pin composeapp version to v95.1

detsch:
@mike-sul The recent merge in composeapp broke the build in our test env, due to the update in  golang required version. We can update the golang version in the Dockerfile as well, but I think we can pin the composeapp first, and then decide how to manage the golang version / composeapp version properly later.


detsch:
> I suppose we just update specific version of Dockerfile with correct/matching pair of the golang container and composeapp?

Yes, I think that is the best approach. We still need to make sure to keep running e2e tests using composeapp from `main`  branch, to detect regressions in composeapp, but that is something we can do outside the github workflow.

PR #429, opened 2025-07-02
api: Fill install result description

detsch:


src/api.cc
```diff
@@ -674,7 +674,7 @@ class LiteInstall : public InstallContext {
     // setAppsNotChecked is required to force a re-load of apps list in case of a new Download operation
     client_->setAppsNotChecked();
     if (client_->VerifyTarget(*target_) != TargetStatus::kGood) {
-      return InstallResult{InstallResult::Status::DownloadFailed, ""};
+      return InstallResult{InstallResult::Status::DownloadFailed, "Target verification failed, it may not have been pulled"};
```
doanac, line 677: I've been away from this code long enough to not understand this well. This message almost makes me think the download failed, but I *think* we are saying what we are verifying *on disk* doesn't match what we expected, right?

doanac: thanks - i guess this message is about the best you can do for this.

PR #430, opened 2025-07-08
appengine: Fix detection of non-running apps

detsch:
When a container is stopped manually (docker stop), its status is reported as "exited" in `composectl ps` JSON status output. `checkAppStatus` was not considering this state as a "broken" state, leading `isRunning` to return true even for an app that was not actually running.


detsch:
Closing as the original behavior is expected. `isRunning` detects if the app was correctly started, and is expected to return true even if some container of the application is killed.

PR #431, opened 2025-07-24
Stream child output to stdout

mike-sul:
Refactor and extend the `exec` function to allow:
1. Setting the timeout for the child process running.
2. Printing the child process output (stdout and stderr) to the master's stdout. It is needed for displaying app blob download progress when aklite invokes `composectl pull`. Depends on https://github.com/foundriesio/composeapp/pull/75.

### DIFF
--- a/src/composeapp/appengine.cc
+++ b/src/composeapp/appengine.cc
@@ -21,11 +21,11 @@ AppEngine::Result AppEngine::fetch(const App& app) {
     if (local_source_path_.empty()) {
       exec(boost::format{"%s --store %s pull -p %s --storage-usage-watermark %d"} % composectl_cmd_ % storeRoot() %
                app.uri % storage_watermark_,
-           "failed to pull compose app", "", nullptr, "", true);
+           "failed to pull compose app", "", nullptr, "4h", true);
     } else {
       exec(boost::format{"%s --store %s pull -p %s -l %s --storage-usage-watermark %d"} % composectl_cmd_ %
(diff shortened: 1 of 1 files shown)


mike-sul:
I've tested it, and it is ready for review.

src/composeapp/appengine.cc
```diff
@@ -21,11 +21,11 @@ AppEngine::Result AppEngine::fetch(const App& app) {
     if (local_source_path_.empty()) {
       exec(boost::format{"%s --store %s pull -p %s --storage-usage-watermark %d"} % composectl_cmd_ % storeRoot() %
                app.uri % storage_watermark_,
-           "failed to pull compose app");
+           "failed to pull compose app", "", nullptr, "", true);
```
detsch, line 24: Does the `timeout` tool break `print_output = true`?
Having some timeout set may serve as a safety net in case of something bad happens inside `composectl` which might lead to it hanging indefinitely. So I would keep it if possible.

mike-sul: No, it does not break the `print_output = true`.

The `composectl pull` has its own timeouts, so this external timeout can interfere with the the command internal timeouts.
Also, the pull can take hours under poor bandwidth. We can set `timeout` to a few hours to make it safe for the pull command, not sure it is worth it though.

detsch: I think a few hours is better than nothing

mike-sul: Set 4 hour timeouts for both `pull` and `install` commands, see this commit d79a3f9a.

PR #433, opened 2025-09-09
composeappmanager: Fix app health deducing

mike-sul:
- App service health must be equal to "healthy" to consider it healthy, not just not equal "unhealthy" because the health value set by docker can be equal to "starting" or "none" which indicates that a service is not healthy.

- If the health is not specified then deduce service health based on its state and set it in the service status json. It helps to unify status sent to Device Gateway regardless of the app engine used (restorable or composectl) and their versions.


doanac:
We should probably mention this in our release notes when it makes it into one.

detsch (approved):

PR #436, opened 2025-10-30
cli: Add --report-hw-info option for check and update command

detsch:
When active, the flag will disable calling `lshw` and reporting the result to the device gateway.
Reporting the HW info is usually not required, and only executed once in the daemon mode. This change allows a CLI user to avoid calling lshw excessively.


mike-sul:
> Maybe we should disable reporting the lshw result by default in CLI check?

Yes, I would disable it by default.

mike-sul (approved):

detsch:
@mike-sul I've changed the option name, and disabled reporting of hw info by default. Please take a look again.

PR #454, opened 2026-04-11
Unify and optimize apps status check

mike-sul:
This PR refactors apps update handling so apps state/status is checked once per update flow instead of from multiple scattered call sites.

It centralizes apps status evaluation in a common helper used by API, daemon, sync, and local install paths, removes redundant checks from `ComposeAppManager::Download()` and install paths, and aligns tests with the new API-driven flow.

As a result, update logic becomes simpler, more consistent, and avoids duplicate app checks while preserving forced-update and removed-app cleanup behavior.

It also clears the path for a follow-up cleanup: removing the app-selection state from `ComposeAppManager` entirely (`cur_apps_to_fetch_and_update_` and `cur_apps_to_fetch_`) and passing the resolved lists of apps to fetch&update as part of the target instance. Hence, `ComposeAppManager` becomes stateless then.

### DIFF
--- a/include/aktualizr-lite/aklite_client_ext.h
+++ b/include/aktualizr-lite/aklite_client_ext.h
@@ -99,9 +99,6 @@ class AkliteClientExt : public AkliteClient {
   bool RebootIfRequired();
 
   bool IsAppRunning(const std::string &name, const std::string &uri) const;
-
- private:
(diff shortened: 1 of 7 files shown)


detsch (approved):
LGTM

PR #455, opened 2026-04-20
e2e-tests: Sync test script with recent changes in fioup tests

detsch:
Fioup still uses the same test script as aklite, in order to reuse
logic. Get the changes done there in order to keep script in sync,
and facilitate future development.


mike-sul (approved):
LGTM

PR #456, opened 2026-04-20
Add test for not enough storage space left on the device

detsch:


docker-e2e-test/e2e-test.py
```diff
@@ -1339,3 +1339,51 @@ def test_fioup_daemon():
...
+    # get available bytes in /var/sota
+    statvfs = os.statvfs("/var/sota")
+    available_bytes = statvfs.f_bavail * statvfs.f_frsize
+    logger.info(f"Available space in /var/sota: {available_bytes} bytes")
+    if available_bytes > 100000000:
+        assert False, "Too much free space left, test environment should be configured to have less than 10MB free space for this test to be effective"
+
+    if available_bytes > 50000:
+        # fill /var/sota with data to reduce available space to 50000 bytes, to trigger no space left error
+        with open("/var/sota/fill_space", "wb") as f:
```
mike-sul, line 1377: Why do we need to "fill space" here if we create a file-based loop device and have a full control over its size?

docker-e2e-test/e2e-test.py
```diff
@@ -1339,3 +1339,51 @@ def test_fioup_daemon():
...
+    write_settings(apps, prune)
+
+    invoke_aklite(['check'])
+
+    # get available bytes in /var/sota
+    statvfs = os.statvfs("/var/sota")
+    available_bytes = statvfs.f_bavail * statvfs.f_frsize
+    logger.info(f"Available space in /var/sota: {available_bytes} bytes")
+    if available_bytes > 100000000:
+        assert False, "Too much free space left, test environment should be configured to have less than 10MB free space for this test to be effective"
```
detsch:
```suggestion
        assert False, "Too much free space left, test environment should be configured to have less than 100MB free space for this test to be effective"
```

mike-sul (approved):

PR #457, opened 2026-04-20
Add bad network, process killing and forced update tests

detsch:


docker-e2e-test/e2e-test.py
```diff
@@ -1387,3 +1400,29 @@ def test_no_space(offline_: bool, single_step_: bool):
...
+    restore_system_state()
+    apps = None # All apps, for now
+    write_settings(apps, prune)
+    try:
+        cmd = [ tc_path, "qdisc", "add", "dev", "eth0", "root", "netem", "delay", "500ms" ]
+        subprocess.call(cmd)
+        for i in range(5):
+            logger.info(f" Calling update and interrupting after {i} seconds")
+            cp = invoke_aklite(['update', str(all_primary_tag_targets[Target.UpdateOstreeWithApps].actual_version)], i)
+            assert cp.returncode == -9, cp.stdout.decode("utf-8")
```
mike-sul: Maybe it makes sense to define the return code "-9" as some kind of const/enum as it is done for `ReturnCodes.InstallNeedsReboot` along with comment explaining the error code meaning.

docker-e2e-test/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM golang:1.22.12-bookworm AS composeapp
 # Build composeapp
 WORKDIR /build
 RUN git clone https://github.com/foundriesio/composeapp.git && cd composeapp \
-    && git checkout lmp-97.0.0 \
+    && git checkout main \
```
mike-sul, line 5: I'll make a new composeapp release and we'll update the reference here then.

docker-e2e-test/e2e-test.py
```diff
@@ -1459,3 +1460,74 @@ def test_bad_network(single_step_: bool):
...
+    # Identify a blob hash for the given application
+    cp = subprocess.run([composectl_path, 'inspect', app_uri, '--format', 'json'], capture_output=True)
+    assert cp.returncode == ReturnCodes.Ok, cp.stdout.decode("utf-8")
+    out_json = json.loads(cp.stdout.decode("utf-8"))
+    blob_digest = out_json['bundle']['services'][0]['image']['manifests'][0]['config']['digest']
+
+    # Test forced `pull` command
+    logger.info("Testing corruption of pulled blob, to make sure it's re-downloaded with `pull` command")
+    pulled_blob_path = f"/var/sota/reset-apps/blobs/sha256/{blob_digest.split(':')[1]}"
+    corrupt_file(pulled_blob_path, blob_digest.split(':')[1])
```
mike-sul: The `blob_digest.split(':')[1]` is used in many places. Defining a new variable `bd = blob_digest.split(':')[1]` or `blob_digest = out_json['bundle']['services'][0]['image']['manifests'][0]['config']['digest'].split(':',1)` should make the code slightly better IMO.

mike-sul (approved):
Nice tests, especially those that emulate poor network bandwidth.

docker-e2e-test/e2e-test.py
detsch: Yes, that makes total sense. Fixed.

docker-e2e-test/e2e-test.py
detsch: It is not aklite specific, so I think it is not the case to add it to `ReturnCodes`. But I've added a local variable to make it more clear its meaning.

PR #458, opened 2026-04-23
Add manual workflow for offline e2e tests

detsch:


docker-e2e-test/e2e-test.py
```diff
@@ -139,6 +141,10 @@ def secondary_tag_is_set():
     logger.info(f"  Secondary Tag: {secondary_tag}")
 logger.info(f"  Base target version: {base_version[primary_tag]}")
 logger.info(f"  User Token: {user_token[:2] + '*' * (len(user_token) - 2)}")
+if e2e_test_ostree_tgz:
+    logger.info(f"  OSTree tar.gz: {e2e_test_ostree_tgz[:2]}...")
+else:
+    logger.info(f"  OSTree tar.gz: not set")
```
mike-sul, line 147: Does it make sense to continue running the test if `e2e_test_ostree_tgz` is not set?

docker-e2e-test/e2e-test.py
```diff
@@ -1013,6 +1014,13 @@ def restore_system_state():
     version = all_primary_tag_targets[Target.First].actual_version
     cleanup_tuf_metadata()
     cleanup_installed_data()
+
+    if offline:
+        # offline bundles miss root metadata versions 1 and 2. Force a online check to get it
+        offline = False
+        cp = invoke_aklite(['check'])
```
mike-sul: Maybe it makes sense to emulate the flow that occurs on a real device, i.e.:

1. prefetch the first two versions of root meta to `/usr/lib/sota/tuf/ci`. The test can do it by running a regular http request: `curl -s -H "OSF-TOKEN: $USER_TOKEN" https://api.foundries.io/ota/repo/$FACTORY/api/v1/user_repo/1.root.json`
2. then, aklite, imports the prefetch root meta on the first run.

If a custom location for the prefetched root meta is required then it can be set in `sota.toml:storage:uptane_metadata_path = <path>`.

docker-e2e-test/e2e-test.py
detsch: Online tests, yes. And, for offline tests, the offline bundles may be already created, so the variable is not strictly necessary.

But it makes sense to adjust the `create_offline_bundles` function with an assert failure if the variable is not set and the `offline-bundles` directory is not present. I'll update it.

detsch: I've updated `create_offline_bundles` accordingly

docker-e2e-test/e2e-test.py
detsch: Done. I'm now downloading root v1 and v2 using curl.

mike-sul (approved):

PR #459, opened 2026-05-13
Run targets generation script as a Github Action

detsch:
This PR contains the required improvements to the e2e-tests scripts and adds a manual action that runs the targets generation script, followed by the execution of all the enabled e2e tests, which currently takes a total of about 2 and a half hours.

PR #462, opened 2026-05-19
e2e-tests: Fix execution of rollback and auto_downgrade tests 

detsch:
@mike-sul Fixes for the tests that where disabled in #459

Single commit on top of #461

### DIFF
--- a/docker-e2e-test/e2e-test.py
+++ b/docker-e2e-test/e2e-test.py
@@ -323,6 +323,7 @@ def verify_events(target_version: int, expected_events: Optional[Set[Tuple[str, 
     assert set(event_list) == set(expected_events)
 
 def sys_reboot():
+    logger.info(f"  Rebooting")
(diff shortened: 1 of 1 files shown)


docker-e2e-test/e2e-test.py
```diff
@@ -1176,7 +1177,7 @@ def run_test_switch_tag():
 
 def run_test_auto_downgrade():
     assert secondary_tag_is_set()
-    auto_downgrade_enabled = True
+    auto_downgrade_enabled = False
```
mike-sul: Setting `auto_downgrade_enabled` to `False` in the test called `run_test_auto_downgrade` sounds a bit counterintuitive.

Also, the "if/else" at the line 1188-1191 seams redundant since the initial value of `auto_downgrade_enabled` never changes.

detsch: I've adjusted the code and the name of the test.

mike-sul (approved):

PR #464, opened 2026-05-22
ci: Use ccache in CI and manual workflows

detsch:
Ccache is now active for both aklite-dev and aklite-dev-e2e images, so we can use it in CI and manual workflows to speed up builds.

Each image has a different cache key. The key also contains the git commit hash, to allow updates to the cache when the code changes. If the specific hash is not found, the cache will be restored using the latest cache.

Two new jobs where added to refresh the cache for both images, and are set as dependencies for the other jobs. This avoids multiple jobs having to perform the same compilation in parallel.

When using ccache effectively, the current CI workflow time is reduced from about 22.5 minutes to about 19.5 minutes. Additional reduction is expected once we split the jobs further, which makes more sense now that each job does not need to perform the same compilation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,9 @@ jobs:
       env:
         CXX: clang++
         CC: clang
-        CCACHE_DIR: /tmp/ccache
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


mike-sul (approved):
It is really nice. What the test running time improvement?

detsch:
> It is really nice. What the test running time improvement?

This change alone, from about 22.5 minutes to about 19.5 minutes. But with the change of #465, we got down to 12 minutes.

PR #466, opened 2026-05-25
tuf: resolve https repo source TLS paths against import base

EmbeddedAndroid:
The local readFileIfExists helper in AkHttpsRepoSource called based_path.get("") which discards the import base path, so relative tls_cacert_path / tls_clientcert_path / tls_pkey_path values from the [import] block resolve to bare filenames in the process CWD. The files are not found, an empty string is written to libcurl's temp cert file, and TLS fails with curl error 58 "Problem with the local SSL certificate" and OpenSSL "PEM routines::no start line".

Pass config.import.base_path through to the helper so paths resolve the same way every other consumer of [import] handles them.

No change for users who already supply absolute paths.

---
Re-filed from #463 with the source branch moved into this repository so CI runs (per the reviewer comment on #463).

### DIFF
--- a/src/tuf/akhttpsreposource.cc
+++ b/src/tuf/akhttpsreposource.cc
@@ -26,11 +26,11 @@ AkHttpsRepoSource::AkHttpsRepoSource(const std::string& name_in, boost::property
   init(name_in, pt, config);
 }
 
-static std::string readFileIfExists(const utils::BasedPath& based_path) {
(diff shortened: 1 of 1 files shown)


detsch (approved):

detsch:
Merge was blocked due to missing GPG signature. Added mine on Tyler's commit. Will merge once tests pass.

PR #467, opened 2026-06-22
composeappmanager: add reserved_storage option

detsch:
pacman.storage_watermark stays a plain percentage. A new pacman.reserved_storage option accepts a byte size (e.g. "2GiB", "500MiB") and reserves that absolute amount of free space for non-Apps usage. When both are set, reserved_storage wins and a warning is logged.

The new option is forwarded to composectl as --reserved-storage, while storage_watermark continues to be forwarded as --storage-usage-watermark. getUsageInfo and the skopeo-engine storage space check use the parsed byte value when reserved_storage is set, and the percentage otherwise.

Assisted-by: Claude Code:claude-opus-4-8

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,6 +41,12 @@ create_containers_before_reboot = "0"
 # By default, if the configuration param is not specified, it is set to "80".
 storage_watermark = "60" (set to "80" if not specified)
 
+# Absolute amount of free storage to reserve for non-Apps usage, expressed as a byte size with
+# either a binary (e.g. "2GiB", "500MiB") or decimal (e.g. "2GB", "500MB") suffix. When set, Apps
+# may use all available storage except the reserved amount, and this option takes precedence over
+# storage_watermark (a warning is logged if both are set). Unset by default.
+# reserved_storage = "2GiB"
+
 [logger]
 # Set log level 0-5 (trace, debug, info, warning, error, fatal)
 loglevel = 2
--- a/src/composeapp/appengine.cc
+++ b/src/composeapp/appengine.cc
@@ -31,12 +31,19 @@ AppEngine::Result AppEngine::fetch(const App& app) {
           was_proxy_set = true;
         }
       }
-      exec(boost::format{"%s --store %s pull -p %s --storage-usage-watermark %d"} % composectl_cmd_ % storeRoot() %
-               app.uri % storage_watermark_,
+      // reserved_storage, when set, forwards an absolute reserved free space; otherwise the percentage
+      // watermark is forwarded. composectl gives --reserved-storage precedence over the watermark.
+      const std::string storage_arg{
+          reserved_storage_.empty() ? (boost::format{"--storage-usage-watermark %d"} % storage_watermark_).str()
+                                    : (boost::format{"--reserved-storage %s"} % reserved_storage_).str()};
+      exec(boost::format{"%s --store %s pull -p %s %s"} % composectl_cmd_ % storeRoot() % app.uri % storage_arg,
            "failed to pull compose app", "", nullptr, "4h", true);
     } else {
-      exec(boost::format{"%s --store %s pull -p %s -l %s --storage-usage-watermark %d"} % composectl_cmd_ %
-               storeRoot() % app.uri % local_source_path_ % storage_watermark_,
+      const std::string storage_arg{
+          reserved_storage_.empty() ? (boost::format{"--storage-usage-watermark %d"} % storage_watermark_).str()
+                                    : (boost::format{"--reserved-storage %s"} % reserved_storage_).str()};
+      exec(boost::format{"%s --store %s pull -p %s -l %s %s"} % composectl_cmd_ % storeRoot() % app.uri %
+               local_source_path_ % storage_arg,
            "failed to pull compose app", "", nullptr, "4h", true);
     }
     res = true;
--- a/src/composeappmanager.cc
+++ b/src/composeappmanager.cc
@@ -1,5 +1,9 @@
 #include "composeappmanager.h"
 
+#include <cctype>
+#include <cmath>
+#include <limits>
+#include <regex>
 #include <set>
 
 #include <boost/algorithm/string.hpp>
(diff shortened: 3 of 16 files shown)


src/composeapp/appengine.cc
```diff
@@ -31,12 +31,19 @@ AppEngine::Result AppEngine::fetch(const App& app) {
...
         }
       }
-      exec(boost::format{"%s --store %s pull -p %s --storage-usage-watermark %d"} % composectl_cmd_ % storeRoot() %
-               app.uri % storage_watermark_,
+      // reserved_storage, when set, forwards an absolute reserved free space; otherwise the percentage
+      // watermark is forwarded. composectl gives --reserved-storage precedence over the watermark.
+      const std::string storage_arg{
+          reserved_storage_.empty() ? (boost::format{"--storage-usage-watermark %d"} % storage_watermark_).str()
+                                    : (boost::format{"--reserved-storage %s"} % reserved_storage_).str()};
+      exec(boost::format{"%s --store %s pull -p %s %s"} % composectl_cmd_ % storeRoot() % app.uri % storage_arg,
```
mike-sul, line 39: We could probably pass both params without analyzing them down to `composectl` since it can handle such situation.

docs/configuration.md
```diff
@@ -41,6 +41,12 @@ create_containers_before_reboot = "0"
 # By default, if the configuration param is not specified, it is set to "80".
 storage_watermark = "60" (set to "80" if not specified)
 
+# Absolute amount of free storage to reserve for non-Apps usage, expressed as a byte size with
+# either a binary (e.g. "2GiB", "500MiB") or decimal (e.g. "2GB", "500MB") suffix. When set, Apps
+# may use all available storage except the reserved amount, and this option takes precedence over
+# storage_watermark (a warning is logged if both are set). Unset by default.
+# reserved_storage = "2GiB"
```
mike-sul, line 48: I think this line should be uncommented.

src/docker/restorableappengine.cc
```diff
@@ -31,20 +31,27 @@ class InsufficientSpaceError : public std::runtime_error {
...
+    }
   }
 
-  return [watermark](const boost::filesystem::path& path) {
-    storage::Volume::UsageInfo usage_info{storage::Volume::getUsageInfo(
-        path.string(), (100 > watermark) ? static_cast<unsigned int>(100 - watermark) : 0, "pacman:storage_watermark")};
+  return [watermark, watermark_in_bytes](const boost::filesystem::path& path) {
+    storage::Volume::UsageInfo usage_info{
+        watermark_in_bytes
+            ? storage::Volume::getUsageInfo(path.string(), watermark, "pacman:storage_watermark", true)
```
mike-sul: I think it should be `pacman:reserved_storage` if `watermark_in_bytes` is true.

src/composeappmanager.cc
```diff
@@ -805,16 +870,20 @@ ComposeAppManager::AppsContainer ComposeAppManager::getAppsToFetch(const Uptane:
...
 std::string ComposeAppManager::getAppsFsUsageInfo() const {
   std::stringstream ss;
-  auto usage_info{storage::Volume::getUsageInfo(cfg_.images_data_root.string(), (100 - cfg_.storage_watermark),
-                                                "pacman:storage_watermark")};
+  const bool in_bytes{cfg_.reserved_storage_bytes > 0};
+  const uint64_t reserved{in_bytes ? cfg_.reserved_storage_bytes
+                                   : ((100 > cfg_.storage_watermark) ? static_cast<uint64_t>(100 - cfg_.storage_watermark)
+                                                                     : 0)};
+  auto usage_info{
+      storage::Volume::getUsageInfo(cfg_.images_data_root.string(), reserved, "pacman:storage_watermark", in_bytes)};
```
mike-sul: I think the reason "storage_watermark" or "reserved_storage" should be set depending on `in_bytes`

src/composeappmanager.cc
```diff
@@ -13,6 +15,46 @@
...
 #endif  // USE_COMPOSEAPP_ENGINE
 
+namespace {
+// Parses a human-readable size into bytes, accepting both binary (e.g. "2GiB", "500MiB") and
+// decimal (e.g. "2GB", "500MB") unit suffixes, matching the formats accepted by composectl's
+// --reserved-storage flag so the same config value yields the same reserved amount across the
+// stack. The presence of an "i" in the suffix selects the binary base (1024); otherwise the
+// decimal base (1000) is used. Returns false when `value` lacks a recognized size unit (e.g. a
+// bare integer, which is interpreted as a percentage instead).
+bool parseSizeInBytes(const std::string& value, uint64_t& bytes) {
```
mike-sul: I would add unit tests for this function.

src/composeappmanager.cc
```diff
@@ -13,6 +15,46 @@
...
+  static const std::regex re{R"(^\s*([0-9]+(?:\.[0-9]+)?)\s*([kKmMgGtTpP])([iI])?[bB]?\s*$)"};
+  std::smatch match;
+  if (!std::regex_match(value, match, re)) {
+    return false;
+  }
+  const double num{std::stod(match[1].str())};
```
mike-sul: This is my AI talking:


The most critical issue here is the unchecked conversion from floating point to `uint64_t`. If the parsed value exceeds `std::numeric_limits<uint64_t>::max()`, the final `static_cast<uint64_t>` has undefined behavior. We should validate that the scaled value is finite and within the `uint64_t` range before assigning it to `bytes`. Also, `std::stod()` should be wrapped because very large numeric inputs can throw.

docs/configuration.md
detsch: We should have only one of `reserved_storage` or `storage_watermark` set in the config. That's I think we should keep only one uncommented in the example.

src/composeapp/appengine.cc
detsch: That would lead to a warning. Also, it would be an error to pass an empty string as `--reserved-storage`

src/composeappmanager.cc
detsch: Fixed.

src/docker/restorableappengine.cc
detsch: Fixed.

src/composeappmanager.cc
detsch: Added.

src/composeappmanager.cc
detsch: Fixed.

docs/configuration.md
mike-sul: It is rather doc example. A user should not copy-paste it. Someone may add `# reserved_storage = "2GiB"` into their config based on this doc. This is just my IMO, not critical.

src/composeapp/appengine.cc
mike-sul: Ok, I thought the `reserved_storage` is set to 0 in composeappmanager if not specified. If so, then composectl won't generate an error not warning if `reserved_stoarage` is specified. If both are specified then the warning is OK, anyway we generate such warning at the aklite start, so additional warning during pull can be useful as well since the initial warning at aklite startup becomes invisible later when `aklite update | pull` is called.

src/composeappmanager.cc
```diff
@@ -13,6 +17,54 @@
...
+  static const std::regex re{R"(^\s*([0-9]+(?:\.[0-9]+)?)\s*([kKmMgGtTpP])([iI])?[bB]?\s*$)"};
+  std::smatch match;
+  if (!std::regex_match(value, match, re)) {
+    return false;
+  }
+  double num{0.0};
+  try {
+    num = std::stod(match[1].str());
+  } catch (const std::exception&) {
+    // Numeric literal that the regex accepted but is too large for double, or otherwise unparseable.
```
mike-sul, line 30: Just a nitpick: I would print warning level log here.

mike-sul (approved):
Quite useful change. Thanks! I have just minor non-critical comments.

src/composeappmanager.cc
detsch: done

src/composeapp/appengine.cc
detsch: It adds some complexity, as we also need to track if usage-watermark was explicitly set.  Otherwise a warning would be generated every time the user specifies reserved-storage. IMO it is not worth it.

mike-sul (approved):

PR #468, opened 2026-07-30
Use new fiopull tool

detsch:
Add optional use of the pure-Go `fiopull` helper (ostree_pull_tool=fiopull)
for pulling ostree commits, a combined ostree+apps pre-download space check,
and update-size estimation that queries fiopull's `update-size` (static-delta
superblock or the commit's ostree.sizes metadata) so an update that will not
fit is rejected before the download starts. When neither source is available
fiopull exits 4 and the size is treated as unavailable (fail-open), leaving the
in-pull space checks as the backstop.


detsch:
@mike-sul e2e test fail on the pre-build targets because there are new tests that require a new target to be built. The complete flow (targets generation + tests) work fine.

detsch:
> @mike-sul e2e test fail on the pre-build targets because there are new tests that require a new target to be built. The complete flow (targets generation + tests) work fine.

I've now parameterized the tests, so that we don't break execution when the targets list changes

mike-sul (approved):

